PR #1, opened 2017-02-03
Spec0.2

zbraniecki:
This is by no means ready to land, but I'd like to get an early feedback and I know PR's are easier for that.

I'm aligning this code closely to what we do in Rust, in particular:

 - iter.js has a generic peekable stream for parsers
 - stream.js has a FTL specific extension of that stream
 - parser.js uses the `FTLParserStream`

The code right now is very minimal - handles only the simplest `key = Value`, but I think it's a good time to work out any design decisions that @stasm may have.

### DIFF
--- a/src/ftl/ast/ast.js
+++ b/src/ftl/ast/ast.js
@@ -37,7 +37,7 @@ class Message extends Entry {
 }
 
 class Pattern extends Node {
-  constructor(source, elements, quoted = false) {
+  constructor(elements, quoted = false) {
     super();
     this.type = 'Pattern';
     this.elements = elements
--- a/src/ftl/ast/ast.js
+++ b/src/ftl/ast/ast.js
@@ -85,10 +85,10 @@ class ExternalArgument extends Expression {
 }
 
 class SelectExpression extends Expression {
-  constructor(exp, variants) {
+  constructor(expression, variants) {
     super();
     this.type = 'SelectExpression';
-    this.exp = exp;
+    this.expression = expression;
     this.variants = variants;
   }
 }
--- a/src/ftl/ast/ast.js
+++ b/src/ftl/ast/ast.js
@@ -52,27 +52,19 @@ class Expression extends Node {
   }
 }
 
-class PatternExpression extends Expression {
-  constructor(pattern) {
-    super();
-    this.type = 'PatternExpression';
-    this.pattern = pattern;
-  }
-}
-
 class StringExpression extends Expression {
-  constructor(string) {
+  constructor(value) {
     super();
     this.type = 'StringExpression';
-    this.string = string;
+    this.value = value;
   }
 }
 
 class NumberExpression extends Expression {
-  constructor(number) {
+  constructor(value) {
     super();
     this.type = 'NumberExpression';
-    this.number = number;
+    this.value = value;
   }
 }
 
@@ -93,7 +85,7 @@ class ExternalArgument extends Expression {
 }
 
 class SelectExpression extends Expression {
-  constructor(exp, variants = null) {
+  constructor(exp, variants) {
     super();
     this.type = 'SelectExpression';
     this.exp = exp;
@@ -113,7 +105,7 @@ class AttributeExpression extends Expression {
 class VariantExpression extends Expression {
   constructor(id, key) {
     super();
-    this.type = 'VariantkExpression';
+    this.type = 'VariantExpression';
     this.id = id;
     this.key = key;
   }
@@ -125,14 +117,6 @@ class CallExpression extends Expression {
     this.type = 'CallExpression';
     this.callee = callee;
     this.args = args;
-  }
-}
-
-class ExpressionExpression extends Expression {
-  constructor(callee, exp) {
-    super();
-    this.type = 'ExpressionExpression';
-    this.exp = exp;
   }
 }
 
@@ -196,14 +180,6 @@ class Section extends Node {
   }
 }
 
-class Number extends Node {
-  constructor(value) {
-    super();
-    this.type = 'Number';
-    this.value = value;
-  }
-}
-
 class Function extends Identifier {
   constructor(name) {
     super(name);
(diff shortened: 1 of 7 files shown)


src/ftl/ast/ast.js
```diff
@@ -18,87 +18,108 @@ class Resource extends NodeList {
...
+    this.type = 'StringExpression';
+    this.string = string;
+  }
+}
+
+class NumberExpression extends Expression {
+  constructor(number) {
+    super();
+    this.type = 'NumberExpression';
+    this.number = number;
```
stasm: In the ASDL, the Number type has a `value` field.

Also, we probably don't need the `Number` class which is defined towards the end of the file.

src/ftl/ast/ast.js
```diff
@@ -18,87 +18,108 @@ class Resource extends NodeList {
...
+    this.type = 'PatternExpression';
+    this.pattern = pattern;
+  }
+}
+
+class StringExpression extends Expression {
+  constructor(string) {
+    super();
+    this.type = 'StringExpression';
+    this.string = string;
```
stasm: The String type isn't defined in the ASDL as it's one of ASDL's built-in types. Should we be more explicit about it and define it? I'd also vote for `StringExpression` to have a `value` field instead of `string`.

src/ftl/ast/ast.js
```diff
@@ -18,87 +18,108 @@ class Resource extends NodeList {
...
-    this.key = key;
-    this.value = value;
-    this.default = def;
+    this.type = 'Expression';
   }
 }
 
-class Entity extends Entry {
-  constructor(id, value = null, traits = [], comment = null) {
+class PatternExpression extends Expression {
```
stasm: In a weakly-typed implementation like JS we probably don't need both `Pattern` and `PatternExpression`. One of them should be enough, right?

src/ftl/ast/ast.js
```diff
@@ -18,87 +18,108 @@ class Resource extends NodeList {
...
-    this.expressions = expressions;
+    this.type = 'ExternalArgument';
+    this.id = id;
   }
 }
 
-class SelectExpression extends Node {
-  constructor(expression, variants = null) {
+class SelectExpression extends Expression {
+  constructor(exp, variants = null) {
```
stasm: Can `variants` be `null`?

src/ftl/ast/ast.js
```diff
@@ -18,87 +18,108 @@ class Resource extends NodeList {
...
+    this.id = id;
+    this.name = name;
   }
 }
 
-class CallExpression extends Node {
+class VariantExpression extends Expression {
+  constructor(id, key) {
+    super();
+    this.type = 'VariantkExpression';
```
stasm: Typo: `Variantk`

src/ftl/ast/ast.js
```diff
@@ -107,34 +128,46 @@ class CallExpression extends Node {
   }
 }
 
-class ExternalArgument extends Node {
-  constructor(name) {
+class ExpressionExpression extends Expression {
```
stasm: Again, do we need this in a weakly-typed implantation?

tools/parse.js
```diff
@@ -31,7 +31,7 @@ program
 
 function parse(fileformat, str) {
   if (program.output === 'ast') {
-    return FTLASTParser.parseResource(str);
+    return FTLASTParser.parse(str);
   } else if (program.transform) {
     const parsed = FTLASTParser.parseResource(str);
```
stasm: Change to `parse` here as well?

src/ftl/ast/parser.js
```diff
@@ -2,808 +2,104 @@
...
+  return new AST.Pattern(elements, false);
 }
 
 export default {
-  parseResource: function(string, { withSource = true } = {}) {
-    const parser = new Parser(withSource);
-    return parser.getResource(string);
-  },
-};
+  parse
```
stasm: You can also `export function parse(source) {` at the top of the file and remove this.

zbraniecki:
>  Is the ParserStream a bit like a lexer which operates on the single character level?

No. It's an extended PeekableIterator. See:
 - https://doc.rust-lang.org/core/iter/struct.Peekable.html
 - https://docs.rs/itertools/0.5.9/itertools/structs/struct.MultiPeek.html

for reference. I advanced based on the MultiPeek adding features like "skip_to_peek" etc.

> What is the role of buf in ParserStream? Does it allow you to make multiple-character-long look-aheads?

yes. It allows me to peek ahead as much as I need without affecting the iterator itself. It's based on MultiPeek.

> Could ParserStream yield tokens rather than individual characters? You'd then peek to see what the next token is.

I did run an experiment trying to get a lexer/parser streams [0] and eventually concluded that this is not beneficial for code clarity at this point or end-result because it requires the parser manage the state of the lexer which ends up being quite complex.

Having a powerful yet small and well tested ParserStream based on MultiPeek gives me a good balance between code complexity and performance.

[0] https://github.com/zbraniecki/fluent-rs/tree/lexer

I think I'll get to write a streamable parser (and it may use lexer) eventually which would be sweet, but I don't think it's worth blocking on it right now.

src/ftl/ast/ast.js
zbraniecki: good point, we probably dont! :)

zbraniecki:
r?stas

src/ftl/ast/ast.js
```diff
@@ -18,87 +18,100 @@ class Resource extends NodeList {
...
+    this.id = id;
+  }
+}
+
+class SelectExpression extends Expression {
+  constructor(exp, variants) {
     super();
     this.type = 'SelectExpression';
-    this.expression = expression;
+    this.exp = exp;
```
stasm: Why not `expression`?

I think it makes sense to use full names in the AST parser.  I'll update the [ASDL](https://github.com/projectfluent/syntax/blob/master/fluent.asdl) too.

test/lib/parser/ftl/iter_test.js
```diff
@@ -0,0 +1,182 @@
+'use strict';
```
stasm: Please put this file in `test/ftl`.

PR #4, opened 2017-02-13
Use better error display

zbraniecki:
This is a port of error display code from rust. I think it gives us a better quality output already, but I'm pretty sure we can clean it up more.

Let me know what you think @stasm .

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "eslint": "^3.9.1",
     "mocha": "^3.0.2",
     "prettyjson": "^1.1.3",
+    "cli-color": "^1.1.0",
     "rollup": "^0.41.4"
   },
   "engine": {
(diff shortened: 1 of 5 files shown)


zbraniecki:
sgtm!

stasm:
Having read https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html and http://www.jonathanturner.org/2016/08/helping-out-with-rust-errors.html I changed my thinking a bit. I described it in https://github.com/projectfluent/fluent/wiki/Error-reporting.

eemeli:
@zbraniecki @stasm Is this still relevant to keep around, or is this PR as outdated as it appears to be?

zbraniecki:
I'm ok closing it. I'd love to see some equivalent of annotate-snippets-rs in JS and put it to use, but I don't think this is actionable right now and we'll likely only start looking into it with tooling for MF2.0

stasm:
Closing sounds good to me, thanks.

PR #5, opened 2017-02-14
Language Negotiation

zbraniecki:
I'd like to start discussing our intended language negotiation strategy (I'd also like to share it between Fluent and Gecko).

I looked at rfc4647 and some of the ICU implementations, but I don't think that what they suggest (matching vs filtering) applies to us perfectly.

Thus, I started working on my own algorithm and the JS implementation is imho a good starting point to discuss it.

The patch here is not finalized but it uses two data-driven algorithms:
 - getParent - which in simplest form cuts off last part ("en-US" -> "en") but eventually will have some exceptions to prevent things like  "zh-Hant" -> "zh", and hopefully will land in ECMA402 (see https://github.com/tc39/ecma402/issues/87)
 - getBestSubtags does not have a good equivalent in ICU so will likely stay curated by us. It allows us to easily go up from more generic to more detailed subtags

The algorithm currently has four parts. It tries, in order:
 - exact match at weight `1.0`
 - likely subtags (tests "en-US" for "en") at weight `0.9`
 - parents (tests "en" for "en-US") at weight `0.8`
 - siblings (tests "en-GB" for "en-US") at weight `0.6`

It stops on the best match for the candidate.

I'd like to get feedback and any suggestions at this point from @Pike and @stasm.

### DIFF
--- a/fluent-langneg/CHANGELOG.md
+++ b/fluent-langneg/CHANGELOG.md
@@ -6,6 +6,6 @@
 
 ## fluent-langneg 0.0.1
 
-  - (05d2487c) fluent-langneg 0.0.1
+  - Introduce Language Negotiation module for Fluent
 
     The initial release.
--- a/fluent-langneg/CHANGELOG.md
+++ b/fluent-langneg/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+
+## Unreleased
+
+  - …
+
+## fluent-langneg 0.0.1
+
+  - (05d2487c) fluent-langneg 0.0.1
+
+    The initial release.
--- a/fluent-langneg/README.md
+++ b/fluent-langneg/README.md
@@ -1,0 +1,104 @@
+# fluent-langneg
+
+`fluent-langneg` is an API for negotiating languages. It's part of
+Project Fluent, a localization framework designed to unleash 
+the expressive power of the natural language.
+
+## Installation
+
+`fluent-langneg` can be used both on the client-side and the server-side.
+You can install it from the npm registry or use it as a standalone script.
+
+    npm install fluent-langneg
+
+
+## How to use
+
+```javascript
+import negotiateLanguages from 'fluent-langneg';
+
+const supportedLocales = negotiateLanguages(
+  requestedLocales, availableLocales, defaultLocale
+);
+```
+
+The API reference is available at
+http://projectfluent.io/fluent.js/fluent-langneg.
+
+## Strategies
+
+The API supports three negotiation strategies:
+
+* filtering (defualt)
+
+In this strategy the algorithm will look for the best matching available
+locale for each requested locale.
+
+Example:
+
+requested: ['de-DE', 'fr-FR']
+available: ['it', 'de', 'en-US', 'fr-CA', 'de-DE', 'fr', 'de-AU']
+
+supported: ['de-DE', 'fr']
+
+* matching
+
+In this strategy the algorithm will try to match as many available locales
+as possible for each of the requested locale.
+
+Example:
+
+requested: ['de-DE', 'fr-FR']
+available: ['it', 'de', 'en-US', 'fr-CA', 'de-DE', 'fr', 'de-AU']
+
+supported: ['de-DE', 'de', 'fr', 'fr-CA']
+
+* lookup
+
+In this strategy the algorithm will try to find the single best locale
+for the requested locale list among the available locales.
+
+Example:
+
+requested: ['de-DE', 'fr-FR']
+available: ['it', 'de', 'en-US', 'fr-CA', 'de-DE', 'fr', 'de-AU']
+
+supported: ['de-DE']
+
+API use:
+
+```javascript
+let supported = negotiateLanguages(requested, available, {
+  strategy: 'matching',
+});
+```
+
+## Likely subtags
+
+Fluent Language Negotiation module carries its own minimal list of likely
+subtags data, which is useful in finding most likely available locales
+in case the requested locale is too generic.
+
+An example of that scenario is when the user requests `en` locale, and
+the application supportes `en-GB` and `en-US`.
+
+Unicode CLDR maintains a complete list of likely subtags that the
+user can load into `fluent-langneg` to replace the minimal version.
+
+```javascript
+let data = require('./data/likelySubtags.json');
+
+let supported = negotiateLanguages(requested, available, {
+  likelySubtags: data.supplemental.likelySubtags
+});
+```
+
+## Learn more
+
+Find out more about Project Fluent at [projectfluent.io][], including
+documentation of the Fluent file format ([FTL][]), links to other packages and
+implementations, and information about how to get involved.
+
+
+[projectfluent.io]: http://projectfluent.io
+[FTL]: http://projectfluent.io/fluent/guide/
--- a/fluent-langneg/README.md
+++ b/fluent-langneg/README.md
@@ -29,7 +29,7 @@ http://projectfluent.io/fluent.js/fluent-langneg.
 
 The API supports three negotiation strategies:
 
-* filtering (defualt)
+### filtering (default)
 
 In this strategy the algorithm will look for the best matching available
 locale for each requested locale.
@@ -41,7 +41,7 @@ available: ['it', 'de', 'en-US', 'fr-CA', 'de-DE', 'fr', 'de-AU']
 
 supported: ['de-DE', 'fr']
 
-* matching
+### matching
 
 In this strategy the algorithm will try to match as many available locales
 as possible for each of the requested locale.
@@ -53,7 +53,7 @@ available: ['it', 'de', 'en-US', 'fr-CA', 'de-DE', 'fr', 'de-AU']
 
 supported: ['de-DE', 'de', 'fr', 'fr-CA']
 
-* lookup
+### lookup
 
 In this strategy the algorithm will try to find the single best locale
 for the requested locale list among the available locales.
@@ -65,7 +65,7 @@ available: ['it', 'de', 'en-US', 'fr-CA', 'de-DE', 'fr', 'de-AU']
 
 supported: ['de-DE']
 
-API use:
+### API use:
 
 ```javascript
 let supported = negotiateLanguages(requested, available, {
@@ -86,7 +86,7 @@ Unicode CLDR maintains a complete list of likely subtags that the
 user can load into `fluent-langneg` to replace the minimal version.
 
 ```javascript
-let data = require('./data/likelySubtags.json');
+let data = require('cldr-core/supplemental/likelySubtags.json');
 
 let supported = negotiateLanguages(requested, available, {
   likelySubtags: data.supplemental.likelySubtags
--- a/fluent-langneg/src/index.js
+++ b/fluent-langneg/src/index.js
@@ -1,0 +1,110 @@
+/*
+ * @module fluent-langneg
+ * @overview
+ *
+ * `fluent-langneg` provides language negotiation API that fits into
+ * Project Fluent localization composition and fallbacking strategy.
+ *
+ */
+
+import filterMatches from './matches';
+
+function GetOption(options, property, type, values, fallback) {
+  let value = options[property];
+
+  if (value !== undefined) {
+    if (type === 'boolean') {
+      value = new Boolean(value);
+    } else if (type === 'string') {
+      value = value.toString();
+    }
+
+    if (values !== undefined && values.indexOf(value) === -1) {
+      throw new Error('Invalid option value');
+    }
+
+    return value;
+  }
+
+  return fallback;
+}
+
+/**
+ * Negotiates the languages between the list of requested locales against
+ * a list of available locales.
+ *
+ * It accepts three arguments:
+ *
+ *   requestedLocales:
+ *     an Array of strings with BCP47 locale IDs sorted
+ *     according to user preferences.
+ *
+ *   availableLocales:
+ *     an Array of strings with BCP47 locale IDs of locale for which
+ *     resources are available. Unsorted.
+ *
+ *   options:
+ *     An object with the following, optional keys:
+ *
+ *       strategy: 'filtering' (default) | 'matching' | 'lookup'
+ *
+ *       defaultLocale:
+ *         a string with BCP47 locale ID to be used
+ *         as a last resort locale.
+ *
+ *       likelySubtags:
+ *         a key-value map of locale keys to their most expanded variants.
+ *         For example:
+ *           'en' -> 'en-Latn-US',
+ *           'ru' -> 'ru-Cyrl-RU',
+ *
+ *
+ * It returns an Array of strings with BCP47 locale IDs sorted according to the
+ * user preferences.
+ *
+ * The exact list will be selected differently depending on the strategy:
+ *
+ *   'filtering':
+ *     In the filtering strategy, the algorithm will attempt to find the
+ *     best possible match for each element of the requestedLocales list.
+ *
+ *   'matching':
+ *     In the matching strategy, the algorithm will attempt to match
+ *     as many keys in the available locales in order of the requested locales.
+ *
+ *   'lookup':
+ *     In the lookup strategy, the algorithm will attempt to find a single
+ *     best available locale based on the requested locales list.
+ */
+export default function negotiateLanguages(
+  requestedLocales,
+  availableLocales,
+  options = {}
+) {
+
+  const defaultLocale = GetOption(options, 'defaultLocale', 'string');
+  const likelySubtags = GetOption(
+    options, 'likelySubtags', 'object', undefined);
+  const strategy = GetOption(options, 'strategy', 'string',
+    ['filtering', 'matching', 'lookup'], 'filtering');
+
+  if (strategy === 'lookup' && defaultLocale === undefined) {
+    throw new Error('defaultLocale cannot be undefined for strategy `lookup`');
+  }
+
+  const supportedLocales = filterMatches(
+    requestedLocales, availableLocales, strategy, likelySubtags
+  );
+
+  if (strategy === 'lookup') {
+    if (supportedLocales.length === 0) {
+      supportedLocales.push(defaultLocale);
+    }
+    return supportedLocales;
+  }
+
+  if (defaultLocale && !supportedLocales.includes(defaultLocale)) {
+    supportedLocales.push(defaultLocale);
+  }
+  return supportedLocales;
+}
--- a/fluent-langneg/src/locale.js
+++ b/fluent-langneg/src/locale.js
@@ -1,0 +1,66 @@
+/* eslint no-magic-numbers: 0 */
+
+const languageCodeRe = '([a-z]{2,3}|\\*)';
+const scriptCodeRe = '(?:-([a-z]{4}|\\*))';
+const regionCodeRe = '(?:-([a-z]{2}|\\*))';
+const variantCodeRe = '(?:-([a-z]+|\\*))';
+
+/**
+ * Regular expression splitting locale id into four pieces:
+ *
+ * Example: `en-Latn-US-mac`
+ *
+ * language: en
+ * script:   Latn
+ * region:   US
+ * variant:  mac
+ *
+ * It can also accept a range `*` character on any position.
+ */
+const localeRe = new RegExp(
+`^${languageCodeRe}${scriptCodeRe}?${regionCodeRe}?${variantCodeRe}?$`, 'i');
+
+export const localeParts = ['language', 'script', 'region', 'variant'];
+
+export default class Locale {
+  /**
+   * Parses a locale id using the localeRe into an array with four elements.
+   *
+   * If the second argument `range` is set to true, it places range `*` char
+   * in place of any missing piece.
+   *
+   * It also allows skipping the script section of the id, so `en-US` is
+   * properly parsed as `en-*-US-*`.
+   */
+  constructor(locale, range = false) {
+    const result = localeRe.exec(locale);
+    if (!result) {
+      return;
+    }
+
+    const missing = range ? '*' : undefined;
+
+    const language = result[1] || missing;
+    const script = result[2] || missing;
+    const region = result[3] || missing;
+    const variant = result[4] || missing;
+
+    this.language = language;
+    this.script = script;
+    this.region = region;
+    this.variant = variant;
+  }
+
+  isEqual(locale) {
+    return localeParts.every(part => this[part] === locale[part]);
+  }
+
+  matches(locale) {
+    return localeParts.every(part => {
+      return this[part] === '*' || locale[part] === '*' ||
+        (this[part] === undefined && locale[part] === undefined) ||
+        (this[part] !== undefined && locale[part] !== undefined &&
+        this[part].toLowerCase() === locale[part].toLowerCase());
+    });
+  }
+}
--- a/fluent-langneg/src/matches.js
+++ b/fluent-langneg/src/matches.js
@@ -1,0 +1,210 @@
+/* eslint no-magic-numbers: 0 */
+/* eslint complexity: ["error", { "max": 22 }] */
+
+import Locale from './locale';
+import { getLikelySubtagsMin } from './subtags';
+
+/**
+ * Attempts to build a most likely maximum version of the locale id based
+ * on it's minimal version.
+ *
+ * If possible it uses the lookup table above. If this one is missing,
+ * it only attempts to place the language code in the region position.
+ */
+function getLikelySubtagsLocale(loc, likelySubtags) {
+  if (likelySubtags) {
+    for (const key of Object.keys(likelySubtags)) {
+      if (key.toLowerCase() === loc.toLowerCase()) {
+        return new Locale(likelySubtags[key]);
+      }
+    }
+    return null;
+  }
+  return getLikelySubtagsMin(loc);
+}
+
+/**
+ * Replaces the region position with a range to allow for matches
+ * with the same language/script but from different region.
+ */
+function regionRangeFor(locale) {
+  locale.region = '*';
+  return locale;
+}
+
+function variantRangeFor(locale) {
+  locale.variant = '*';
+  return locale;
+}
+
+function getOrParseLocaleRange(localeStr, cache) {
+  if (!cache.has(localeStr)) {
+    cache.set(localeStr, new Locale(localeStr, true));
+  }
+  return cache.get(localeStr);
+}
+/**
+ * Negotiates the languages between the list of requested locales against
+ * a list of available locales.
+ *
+ * The algorithm is based on the BCP4647 3.3.2 Extended Filtering algorithm,
+ * with several modifications:
+ *
+ *  1) available locales are treated as ranges
+ *
+ *    This change allows us to match a more specific request against
+ *    more generic available locale.
+ *
+ *    For example, if the available locale list provides locale `en`,
+ *    and the requested locale is `en-US`, we treat the available locale as
+ *    a locale that matches all possible english requests.
+ *
+ *    This means that we expect available locale ID to be as precize as
+ *    the matches they want to cover.
+ *
+ *    For example, if there is only `sr` available, it's ok to list
+ *    it in available locales. But once the available locales has both,
+ *    Cyrl and Latn variants, the locale IDs should be `sr-Cyrl` and `sr-Latn`
+ *    to avoid any `sr-*` request to match against whole `sr` range.
+ *
+ *    What it does ([requested] * [available] = [supported]):
+ *
+ *    ['en-US'] * ['en'] = ['en']
+ *
+ *  2) likely subtags from LDML 4.3 Likely Subtags has been added
+ *
+ *    The most obvious likely subtag that can be computed is a duplication
+ *    of the language field onto region field (`fr` => `fr-FR`).
+ *
+ *    On top of that, likely subtags may use a list of mappings, that
+ *    allow the algorithm to handle non-obvious matches.
+ *    For example, making sure that we match `en` to `en-US` or `sr` to
+ *    `sr-Cyrl`, while `sr-RU` to `sr-Latn-RU`.
+ *
+ *    This list can be taken directly from CLDR Supplemental Data.
+ *
+ *    What it does ([requested] * [available] = [supported]):
+ *
+ *    ['fr'] * ['fr-FR'] = ['fr-FR']
+ *    ['en'] * ['en-US'] = ['en-US']
+ *    ['sr'] * ['sr-Latn', 'sr-Cyrl'] = ['sr-Cyrl']
+ *
+ *  3) variant/region range check has been added
+ *
+ *    Lastly, the last form of check is against the requested locale ID
+ *    but with the variant/region field replaced with a `*` range.
+ *
+ *    The rationale here laid out in LDML 4.4 Language Matching:
+ *      "(...) normally the fall-off between the user's languages is
+ *      substantially greated than regional variants."
+ *
+ *    In other words, if we can't match for the given region, maybe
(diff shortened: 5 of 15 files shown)


zbraniecki:
RFC4647 - https://tools.ietf.org/html/rfc4647

Pike:
Looking at https://tools.ietf.org/html/rfc4647#section-3.3.2, that's mostly what we want, aside from the problem that, AFAICT, `langneg(["de-DE"], ["de"]) === []`, and we want that to be `["de"]`, right?

I think the right way to address that problem is to treat the available languages as language ranges. Which is what we intend to do: `fr` is used for French spoken across the globe. Not French spoken nowhere. I think that getBestSubtags is the wrong concept.

For `langneg(["en"], ["en-US", "en-GB"])` that does leave us with some ambiguity, and I think that's accurate. On our UX side, we should discourage the situation. We can build in a deterministic ordering of sibling language tags, which would do the thing you built getBestSubtags for, in this use case?

For siblings, conceptually, we ammend the requested language range with a more generic range, if applicable? `["de-DE"]` becomes `["de-DE", "de-*"]` ? I think that's sound if we have the "don't drop script ranges for locales where that matters". Are there more constraints to that?

We have two choices here, `["de-DE", "fr-CA"]` can either become `["de-DE", "de-*", "fr-CA", "fr-*"]` or `["de-DE", "fr-CA", "de-*", "fr-*"]`. I think it's the first we want?

The numerical weights are hard to get right, I think. In particular in the code right now, the weights mix as soon as you have 10 or so requested languages, as you mix order within the requested language array and weight. I'd much rather just use different output lists, and concat and unique them afterwards.

Last but not least, the idea that this algorithm always returns an non-empty result is OK, but not something that can be implemented generically. It's a project setup question on which localization to fall back to in the end. Take the BN, maybe? Site developed in Polish, with a German localization? You'd need to configure your l10n setup to fall back to Polish in this case.

I think the correct way to implement this is to post-fix the requested languages with the site default language, and only then to call into the language negotiation code. By doing that, you guarantee a non-empty result, but that's not part of the algorithm, but part of the setup.

zbraniecki:
Thanks Pike!

I started implementing it and the only problem I encountered so far is the one related to:

```
langneg(["en"], ["en-US", "en-GB"])
```

Since the second list is not ordered, there's no way for me to predict which one will it pick.

ICU introduces `addLikelySubtags` which I was intended to use - http://www.icu-project.org/apiref/icu4c/uloc_8h.html#a0cb2dcd65f745e7a966a729395499770

Do you think it's ok to use that?

I may also have to expand ranges with scripts (so, "en" is not becoming "en-*", but "en-Latn-*" in order to handle sr, zh etc.).

zbraniecki:
I *think* I'm implementing this:  http://www.unicode.org/reports/tr35/#LanguageMatching
so I'll have weights there.

Is that ok, @Pike ?

Pike:
The whole weights stuff in tr35 is a bit bewildering, their math is close to using complex numbers?

We'd run the match algorithm as a filter, though, right?

Also, in the gecko version, we can use addLikelySubtags, but we'd still need languageInfo.xml data? And for a vanilla js impl, we'd need that and likelySubtags.xml, too?

zbraniecki:
> We'd run the match algorithm as a filter, though, right?

yes

> but we'd still need languageInfo.xml data? And for a vanilla js impl, we'd need that and likelySubtags.xml, too?

For JS, I'd like to limit the amount of data we carry. I'm thinking about trying to filter the data based on all-locales.

Pike:
Am 17/02/2017 um 19:47 schrieb Zibi Braniecki:
>
>     but we'd still need languageInfo.xml data? And for a vanilla js
>     impl, we'd need that and likelySubtags.xml, too?
>
> For JS, I'd like to limit the amount of data we carry. I'm thinking 
> about trying to filter the data based on all-locales.
>
In the context of fluent, reducing the dataset in general to the needs 
of mozilla doesn't seem to be the right approach. I'm not sure if 
there's a way to drill down on the dataset for a general-use low-level 
library.

zbraniecki:
@Pike - I updated the patch to mostly follow the RFC4647, with a few bits from LDML likelySubtags and languageMatching.

I think the code is actually pretty good for how small it is (less than 100 lines of code). It doesn't do everything that I'd like it to do (including ability for each available locale to provide its own fallback chain), but I think it's a good start.

Let me know what you think.

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+ */
+function addLikelySubtags(locale) {
+  if (likelySubtags.hasOwnProperty(locale)) {
+    return likelySubtags[locale];
+  }
+  const loc = parseLocale(locale, true);
+  if (loc[2] === '*') {
+    loc[2] = loc[0].toUpperCase();
+    return loc.join('-');
+  }
```
Pike: We should make this fully data-driven. Assuming that a language tag in upper case was a region code is just wrong. Let's not introduce `ca-CA` or `lij-LIJ` as likely subtags?

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+  return null;
+}
+
+/**
+ * Replaces the region position with a range to allow for matches
+ * with the same language/script but from different region.
+ */
+function getRegionRange(locale) {
```
Pike: I'd name this something else, maybe `regionRangeFor`? getRegionRange sounds like a getter of information, not a transformation of the passed in data.

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * language: en
+ * script:   Latn
+ * region:   US
+ * variant:  mac
+ *
+ * It can also accept a range `*` character on any position.
+ */
+const localeRe = new RegExp(
+`^${languageCodeRe}${scriptCodeRe}?${regionCodeRe}?${variantCodeRe}?$`);
```
Pike: The regular expression isn't implementing the spec, I'd recommend to just parse a script tag as a {4}, and make it optional. Like
```
/^([a-z]{2,3})(?:[-([a-z]{4}|\*))?/i
```
I'm not sure if we should allow for trailing junk?

Also, we should probably special-case `-mac` as something that violates well-formed locale tags, as long as we still have it.

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+    territory = result[3];
+  }
+
+  let variant = range ? '*' : undefined;
+  if (result[4]) {
+    variant = result[4];
+  }
+
+  return [language, script, territory, variant];
+}
```
Pike: ... this code would become easier, if the regular expression was more standards-conformant, I think.

Also, we should probably normalize casing here, so that we can compare strings later on. 'cause language tags are compared case-independently.

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+ *
+ *    What it does ([requested] * [available] = [supported]):
+ *
+ *    ['en-AU'] * ['en-US'] = ['en-US']
+ *    ['sr-RU'] * ['sr-Latn-RO'] = ['sr-Latn-RO'] // sr-RU -> sr-Latn-RU
+ */
+function filterMatches(requestedLocales, availableLocales) {
+  const supportedLocales = [];
+
+  outer:
```
Pike: This is probably up to @stasm, but GOTOs?

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+
+    for (const availableLocale of availableLocales) {
+      if (requestedLocale === availableLocale) {
+        supportedLocales.push(availableLocale);
+        continue outer;
+      }
+    }
+
+    for (const availableLocale of availableLocales) {
+      if (compareLocales(requestedLocale, availableLocale)) {
```
Pike: This parses the language tags over and over in a loop, I'd refactor this code to not do that.

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+ *    ['sr-RU'] * ['sr-Latn-RO'] = ['sr-Latn-RO'] // sr-RU -> sr-Latn-RU
+ */
+function filterMatches(requestedLocales, availableLocales) {
+  const supportedLocales = [];
+
+  outer:
+  for (const requestedLocale of requestedLocales) {
+
+    for (const availableLocale of availableLocales) {
+      if (requestedLocale === availableLocale) {
```
Pike: ... needs to be independent of casing.

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+      }
+    }
+  }
+
+  return supportedLocales;
+}
+
+export function negotiateLanguages(requestedLocales,
+                                   availableLocales,
+                                   defaultLocale) {
```
Pike: defaultLocale isn't covered in tests, otherwise the bug below would have been caught.

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+  const supportedLocales = filterMatches(requestedLocales, availableLocales);
+  if (supportedLocales.includes(defaultLocale)) {
+    supportedLocales.push(defaultLocale);
+  }
```
Pike: ... if undefined is in list, add it to the list.

You want that the other way around.

fluent/src/langneg.js
```diff
@@ -0,0 +1,246 @@
...
+        supportedLocales.push(availableLocale);
+        continue outer;
+      }
+    }
+  }
+
+  return supportedLocales;
+}
+
+export function negotiateLanguages(requestedLocales,
```
Pike: This is the only public method of this module, and the only without documentation :-/

@stasm, what's the intent for the generated documentation? It seems to be done, but I haven't found it having a prominent role yet, so maybe it doesn't matter. Your call.

zbraniecki:
Thank you Like! That's precisely the level of detail I want to iron things out on right now.

Let's get to an algorithm we both like and then I'll add detailed tests for each function and step.

I'll work on applying your feedback today.

zbraniecki:
*Pike (autocorrect snafu)

fluent/src/langneg.js
zbraniecki: So, here's the thing. LikelySubtags is a 50kb uncompressed JSON table [0].
If we have it (Gecko does), or in case where the user will want to pay the price (give me fluent with likelySubtags), then we don't need a dummy code like that.

But when the user doesn't want to pay the price, and 50kb is a pretty hefty price to pay on the client side, lack of this dummy code will make us not be able to catch any kind of 'ab-AB' extrapolation from 'ab'. 
Since it's unlikely that anyone will provide 'ca-CA' or 'lij-LIJ' in available locales, there's no consequence of testing against it. But in the simplest form, they'll at least match all 'fr' to 'fr-FR', and 'it' to 'it-IT' and 'de' to 'de-DE' etc.

My idea was to allow for three "modes":
* Full CLDR likelySubtags (50kb price for excellent matching)
* Limited CLDR likelySubtags (say, 10-20 most common ones) + dummy
* Just dummy

I can see us curating this limited CLDR likelySubtags just to catch the most common ones like 'en' into 'en-US', and 'fr' into 'fr-FR' etc.

I believe that if we remove the dummy all together, we'll make the algorithm not work for most common cases and basically require the likelySubtags.

What do you think?

[0] https://github.com/unicode-cldr/cldr-core/blob/master/supplemental/likelySubtags.json

fluent/src/langneg.js
zbraniecki: I'm following BCP47 in language-script-region-variant.

fluent/src/langneg.js
zbraniecki: Yeah, I'll normalize case, either by going all lower-case or running it via Canonicalize.
what do you mean by easier?

fluent/src/langneg.js
zbraniecki: we can improve the exact code later if we agree on how it's supposed to work and what kind of matches produce.

fluent/src/langneg.js
zbraniecki: agree. I'll improve the exact code once we agree on what the code is supposed to do :)

fluent/src/langneg.js
zbraniecki: I'll update the regexp, but I think it makes sense to keep the 4 pieces. I may accept and cut off any extension keys (`-t-`, `-x-` and `-u-`).

fluent/src/langneg.js
stasm: Thanks for bringing this up. For now, each package in the `fluent.js` repository has its docstrings extracted using the `documentation.js` package. The extracted markdown files (one per package) land in the top-loevel `docs/` directory.

It might be better to move the docs into each package's directory as `$(PACKAGE)/docs/api.md`, in case we want to add more hand-written docs later on.

fluent/src/langneg.js
stasm: I don't have a strong preference about labelled for-loops.

fluent/src/langneg.js
stasm: > It might be better to move the docs into each package's directory as $(PACKAGE)/docs/api.md, in case we want to add more hand-written docs later on.

Implemented in https://github.com/projectfluent/fluent.js/commit/6c46100af45fb29c19e611c04c60f66758d2970f.

fluent/src/langneg.js
Pike: I find the line about "a language tag is a script tag with bad casing" pretty revolting, tbh. Let's not set precedent for that, people look at mozilla for how to handle standards, and this is just not like it.

I also start to realize that I actually disagree with the way you use of likelySubtags. In that context, I think most of the unfiltered likelySubtags data goes away. I'll take that top-level.

fluent/test/langneg_test.js
```diff
@@ -0,0 +1,142 @@
...
+      negotiateLanguages(['en-Latn-GB', 'en-Latn-IN'], ['en-IN', 'en-GB']),
+      ['en-GB', 'en-IN']);
+
+    assert.deepEqual(
+      negotiateLanguages(['sr-RU'], ['sr-Cyrl', 'sr-Latn']),
+      ['sr-Latn']);
+
+    assert.deepEqual(
+      negotiateLanguages(['sr'], ['sr-Cyrl', 'sr-Latn']),
+      ['sr-Cyrl']);
```
Pike: I spent half a day trying to find out why cldr says these things about Serbian, and to the best of my findings, this is just made up. Notably, sr-Latn-RS is given a writingPopulation of 5%,  because

    <reference type="R1017">For languages not customarily written, the writing population is artificially set to 5% in the absence of better information.</reference>

stasm:
> On the topic of modes, how is that supposed to work in practice? As in, I npm install fluent, and then?

I'd like the language negotiation module to be a separate package.  You would then do:

    npm install fluent-langneg

And:

```javascript
import negotiateLanguages from 'fluent-langneg';
negotiateLanguages(requested, available);
```

Any additional data can be published in a separate package or as a submodule:

```javascript
import negotiateLanguages from 'fluent-langneg';
import withLikelySubtags from 'fluent-langneg/subtags';
withLikelySubtags(negotiateLanguages)(requested, available);
```

For this to work, `negotiateLanguages` should accept `likelySubtags` as a parameter. Perhaps a better approach overall would be to mimic other `Intl` objects:

 ```javascript
const negotiator = LanguageNegotiation({
  available: […],
  default: […],
  likelySubtags: […]
});
```

And then you `import likelySubtags from 'fluent-langneg/subtags';`

File: fluent-langneg/test/langneg_test.js [hunk 12]
```diff
@@ -0,0 +1,144 @@
+
```
stasm: Why are there two blank lines here and in `test/setup.js`? Do you know if the `use strict` pragma is still required in node? Can we skip it?

zbraniecki: I think we can skip it.

fluent-langneg/README.md
```diff
@@ -0,0 +1,34 @@
...
+## How to use
+
+```javascript
+import negotiateLanguages from 'fluent-langneg';
+
+const supported = negotiateLanguages(requested, available, default);
+```
+
+The API reference is available at
+http://projectfluent.io/fluent.js/fluent-syntax.
```
stasm: http://projectfluent.io/fluent.js/fluent-langneg :)

fluent-langneg/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+    npm install fluent-langneg
+
+
+## How to use
+
+```javascript
+import negotiateLanguages from 'fluent-langneg';
+
+const supported = negotiateLanguages(requested, available, default);
```
stasm: `default` is a reserved keyword in JS, so I'd suggest changing this to:

    const supportedLanguages = negotiateLanguages(
        requestedLanguages, availableLanguages, defaultLanguage
    );

stasm: Also, are these _languages_ or _locales_?

fluent-langneg/src/index.js
```diff
@@ -0,0 +1,266 @@
...
+
+    return value;
+  }
+
+  return fallback;
+}
+
+export default function negotiateLanguages(requestedLocales,
+                                   availableLocales,
+                                   options = {}) {
```
stasm: Weird indentation here. You can just indent one level like so:

```
export default function negotiateLanguages(
  requestedLocales,
  availableLocales,
  options = {}
) {
```

fluent-langneg/src/index.js
```diff
@@ -0,0 +1,266 @@
...
+
+  const defaultLocale = GetOption(options, 'defaultLocale', 'string');
```
stasm: Isn't `defaultLocale` common enough that it should get its own positional argument?

fluent-langneg/src/index.js
```diff
@@ -0,0 +1,266 @@
...
+  const likelySubtags = GetOption(
+    options, 'likelySubtags', 'object', undefined, {});
+
+  const supportedLocales =
+    filterMatches(requestedLocales, availableLocales, likelySubtags);
```
stasm: nit: can you change this to:

      requestedLocales, availableLocales, likelySubtags
    );

fluent-langneg/data/likelySubtags-min.json
```diff
@@ -0,0 +1,18 @@
+{
```
stasm: This is so small that I think we should include this in the main package.  Can rollup bundle JSON files? If not, I'd just put this in `src/subtags.js`.

fluent-langneg/src/index.js
```diff
@@ -0,0 +1,266 @@
+/*
+ * @module fluent-langneg
```
stasm: Please add the `@overview` tag here.

fluent-langneg/src/index.js
zbraniecki: I don't think it is. Since the algorithm works perfectly fine without it, I'd prefer to keep only the necessary arguments as positional - requestedLocales and availableLocales.

fluent-langneg/README.md
zbraniecki: those are locales. We will negotiate between the language part of those locales so we will negotiate languages, but we pass locales and the returned ids are locales as well. I know it's complex :(

fluent-langneg/src/index.js
stasm: Perhaps I'm missing something but I don't see it in tests. What happens when there are no common languages between `requested` and `available` and no default has been specified?

zbraniecki: we return an empty list as per language matching spec in rfc4647. Added to tests.

fluent-langneg/data/likelySubtags-min.json
zbraniecki: I don't know how small it will be yet. I'll need to generate the data for the `min` variant, and so far it's just a stub. I agree we'll want to bundle it, but I'd like to make it a conditional bundling i.e. when you're building fluent-langneg I'd like you to be able to choose if you want with `min` or full likelySubtags.

zbraniecki:
@Pike , @stasm - I think this is ready for another round of reviews.

There are two three known things I'd like to add once we agree on the logic:

1) more tests
2) ability to build fluent-langneg with either 'min' or full likelySubtags
3) a script to generate the 'likelySubtagsMin' based on our own threshold.

But first, I'd like to iron out the first version of the algorithm to be stable.

fluent-langneg/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+    throw new Error('defaultLocale cannot be undefined for strategy `lookup`');
+  }
+
+  const supportedLocales = filterMatches(
+    requestedLocales, availableLocales, strategy, likelySubtags
+  );
+
+  if (strategy === 'lookup') {
+    if (supportedLocales.length === 0) {
+      supportedLocales.push(defaultLocale);
```
stasm: What if the `defaultLocale` is undefined? Are you okay pushing it to `supportedLocales` too?

fluent-langneg/src/locale.js
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Parses a locale id using the localeRe into an array with four elements.
+ *
+ * If the second argument `range` is set to true, it places range `*` char
+ * in place of any missing piece.
+ *
+ * It also allows skipping the script section of the id, so `en-US` is properly
+ * parsed as `en-*-US-*`.
+ */
+export function parseLocale(locale, range = false) {
```
stasm: `export default`?

fluent-langneg/src/matches.js
```diff
@@ -0,0 +1,183 @@
...
+ *    we can match for the same language/script but other region, and
+ *    it will in most cases be preferred over falling back on the next
+ *    language.
+ *
+ *    What it does ([requested] * [available] = [supported]):
+ *
+ *    ['en-AU'] * ['en-US'] = ['en-US']
+ *    ['sr-RU'] * ['sr-Latn-RO'] = ['sr-Latn-RO'] // sr-RU -> sr-Latn-RU
+ */
+export function filterMatches(
```
stasm: `export default`?

fluent-langneg/src/matches.js
```diff
@@ -0,0 +1,183 @@
...
+  requestedLocales, availableLocales, strategy, likelySubtags
+) {
+  const supportedLocales = new Set();
+
+  const availableLocalesCache = {};
```
stasm: Use `Map`? That way you won't have to check with `hasOwnProperty` in `getOrParseLocale`.

fluent-langneg/src/index.js
zbraniecki: If it's undefined it'll throw earlier.

zbraniecki:
> My recommendation right now would be to ship 0.0.1 with 10-20 most likely subtags hardcoded in the JS source

The challenge here is how we select the 10-20? Is 'it' more important than 'in'?

> and with a dummy algorithmic way of guessing more. It's okay if it makes mistakes at this stage.

@Pike would you be ok with this?

Alternatively, I could add a curated by us list of locals that are ok to expand:

```javascript
const localesThatHasDefaultThatMatchRegion = ['it', 'fr', 'ru', 'cs', 'pl', ...];
```

instead?

zbraniecki:
> I'd also put the 50KB JSON in the data/ folder so that anyone can use it if they wish. They'll need to figure out how to load it themselves, but at least it's there.

Would you prefer that over external dependency or just a link to cldr-json npm package?

zbraniecki:
> I'd also prefer if the JSON file name was all lowercase, as this is the file naming scheme we've been using so far.

My hope was to keep a direct link to CLDR and they use camel case.

> I'd suggest using default exports more, especially with well-scoped modules. From https://esdiscuss.org/topic/moduleimport#content-0: ES6 favors the single/default export style.

The reason I didn't do that is so that I can expose more internal methods for tests. If each file can only expose one function, I can't expose others for testing.

> Let's move this to a future milestone, together with finding out how to bundle JSON files using our build system.

SGTM.

@Pike?

fluent-langneg/src/index.js
stasm: Ah, I didn't notice that. Thanks.

stasm:
> Would you prefer that over external dependency or just a link to cldr-json npm package?

Ah, if that's an option, then let's go for it: let's link the CLDR repo in the README.

> The reason I didn't do that is so that I can expose more internal methods for tests. If each file can only expose one function, I can't expose others for testing.

You sure can :)

    import foo, { bar } from './module'

imports the default exports as `foo` as well as the named `bar` export

zbraniecki:
I want to ❤️️ es6 right now.

fluent-langneg/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+
+## Unreleased
+
+  - …
+
+## fluent-langneg 0.0.1
+
+  - (05d2487c) fluent-langneg 0.0.1
```
stasm: This is my bad: I shouldn't have included the SHAs in other changelogs.  I've since stopped doing it.  Changelogs should describe changes between named versions and they don't have to correspond to commits. Let's remove the SHA of the commit here and squash everything into a single commit.

fluent-langneg/README.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+The API reference is available at
+http://projectfluent.io/fluent.js/fluent-langneg.
+
+## Strategies
+
+The API supports three negotiation strategies:
+
+* filtering (defualt)
```
stasm: Typo in `default`.

stasm: Also, please either use a `###` header or intend the text below by 4 space to make it part of the bullet point.

fluent-langneg/README.md
```diff
@@ -0,0 +1,104 @@
...
+in case the requested locale is too generic.
+
+An example of that scenario is when the user requests `en` locale, and
+the application supportes `en-GB` and `en-US`.
+
+Unicode CLDR maintains a complete list of likely subtags that the
+user can load into `fluent-langneg` to replace the minimal version.
+
+```javascript
+let data = require('./data/likelySubtags.json');
```
stasm: We're not shipping this file with the rest of the package (which is good) so maybe link to the CLDR repo and put the following here:

    const data = require('cldr-core/supplemental/likelySubtags.json');

fluent-langneg/README.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+The API reference is available at
+http://projectfluent.io/fluent.js/fluent-langneg.
+
+## Strategies
+
+The API supports three negotiation strategies:
+
+### filtering (default)
```
Pike: Wouldn't the general use-case within fluent be to use `matching` instead of `filtering`? In that case, I'd make that default.

fluent-langneg/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+) {
+
+  const defaultLocale = GetOption(options, 'defaultLocale', 'string');
+  const likelySubtags = GetOption(
+    options, 'likelySubtags', 'object', undefined);
+  const strategy = GetOption(options, 'strategy', 'string',
+    ['filtering', 'matching', 'lookup'], 'filtering');
+
+  if (strategy === 'lookup' && defaultLocale === undefined) {
+    throw new Error('defaultLocale cannot be undefined for strategy `lookup`');
```
Pike: This condition should be in the docs.

fluent-langneg/src/matches.js
```diff
@@ -0,0 +1,210 @@
...
+ */
+function regionRangeFor(locale) {
+  locale.region = '*';
+  return locale;
+}
+
+function variantRangeFor(locale) {
+  locale.variant = '*';
+  return locale;
+}
```
Pike: `regionRangeFor` and `variantRangeFor` sound like they'd create new Locale objects, IMHO. Not sure if it's better to actually do a copy, or make this a setter. I could even go as far as to not have this be a function. Or make it a setter on the object?

fluent-langneg/src/matches.js
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+    }
+
+    const requestedLocale = new Locale(reqLocStrLC);
+
+    // Attempt to match against the available range
+    // This turns `en` into `en-*-*-*` and `en-US` into `en-*-US-*`
+    // Example: ['en-US'] * ['en'] = ['en']
+    for (const availableLocale of availableLocales) {
```
Pike: It took me quite a while to untangle this code, mostly because the patterns in the code aren't factored in.

Is there a way to make this a second inner loop, between this and the inner
```
if (matches) {
  add_and_bail_if_not_matching
}
```
? Might make the code shorter, and easier to digest, I think

fluent-langneg/src/subtags.js
```diff
@@ -0,0 +1,62 @@
...
+  'pt': 'pt-latn-br',
+  'sr': 'sr-cyrl-rs',
+  'sr-ru': 'sr-latn-ru',
+  'sv': 'sv-latn-se',
+  'ta': 'ta-taml-in',
+  'uk': 'uk-cyrl-ua',
+  'zh': 'zh-hans-cn',
+  'zh-gb': 'zh-hant-gb',
+  'zh-us': 'zh-hant-us',
+};
```
Pike: It'd be good to add to the comment on how you got to this particular sublist.

fluent-langneg/test/langneg_filtering_test.js
```diff
@@ -0,0 +1,153 @@
+import assert from 'assert';
+import negotiateLanguages from '../src/index';
+
+describe('Basic Language Negotiation without likelySubtags', () => {
+  const nl = negotiateLanguages;
+
+  it('exact match', () => {
```
Pike: Should these it() statements read as full sentences starting with "it ..."? mocha docs say yes, not sure if there's a special style among the fluent repos.

fluent-langneg/README.md
zbraniecki: That's up to @stasm I guess.

fluent-langneg/src/matches.js
zbraniecki: How would I bail out of the `outer` if I factored it out to a separate function?

fluent-langneg/src/subtags.js
zbraniecki: I'm ashamed to admit that I went through likelySubtags and selected locales that I consider to be more significant and likely to have someone specify the minimized version (like `ab` in requested instead of `ab-CD`) where there are more than one region/script.

Not sure how to document it and I believe we should eventually do this in a more formal and algorithmic way.

fluent-langneg/src/matches.js
zbraniecki: @stasm - what do you think? Is it worth creating a new object over modifying the existing one?

fluent-langneg/src/matches.js
Pike: I was thinking of just loops, not functions.
You could also do functions, and instead of continue :outer, do early returns?

fluent-langneg/src/subtags.js
Pike: I'd go for a comment saying that then. You could even file an issue for the follow-up and link to that from the comment?

zbraniecki: Will do!

fluent-langneg/src/matches.js
zbraniecki: I'm working on the C++ port right now for Gecko.  Once I'm done with it, I'll update the JS implementation to match it. It does have this change as well.

fluent-langneg/README.md
zbraniecki: Implemented it in C++ as the default, I'll update js version once I'm done with C++.

fluent-langneg/test/langneg_filtering_test.js
stasm: I filed https://github.com/projectfluent/fluent.js/pull/14.

PR #7, opened 2017-02-16
Add spans to Entries and Annotations to Resource

stasm:
This change allows consumers of the fluent-syntax API to display nice error
messages.

`Entry` instances now have a `span: { from, to, lineno }` field. `from` and
`to` are positions relative to the start of the file. `lineno` is the 0-based
line number in which the entry starts.  Two examples follow.

```json
{
    "type": "Message",
	"id": {
	    "type": "Identifier",
	    "name": "hello-world"
	},
	"value": {
	    "type": "Pattern",
	    "elements": [
	    {
		"type": "StringExpression",
		"value": "Hello, world!"
	    }
	    ],
	    "quoted": false
	},
	"attributes": null,
	"comment": null,
	"span": {
	    "from": 38,
	    "to": 65,
	    "lineno": 2
	}
},
```

```json
{
    "type": "Junk",
	"content": "hello-world Hello, world!\n\n",
	"span": {
	    "from": 44,
	    "to": 71,
	    "lineno": 4
	}
},
```

Parsing errors are now collected into the `annotations` field of `Resource`
instances.  In the future, we can add warnings and notes there, too.  Each
annotation holds a reference to an `Entry` instance which created the
annotation.  This allows the consumers of the API to build their own formatting
rules for annotations.

```json
{
    "name": "SyntaxError",
    "message": "Missing field",
    "lineNumber": 4,
    "columnNumber": 13,
    "entry": {
        "type": "Junk",
        "content": "hello-world Hello, world!\n\n",
        "span": {
            "from": 44,
            "to": 71,
            "lineno": 4
        }
    }
},
```

As per [Error-reporting][], in the future we may considering adding `labels` to
annotations to provide explanation of the error and suggested ways of fixing
it.

[Error-reporting]: https://github.com/projectfluent/fluent/wiki/Error-reporting

Lastly, the `tools/parse` script now prints the following output for errors.
I didn't use any colors at this point because I didn't want to make the patch
too complicated.

    ! SyntaxError on line 3:
      | @xxx
      … ^----- Expected entry

    ! SyntaxError on line 5:
      | hello-world Hello, world!
      …             ^----- Missing field

    ! SyntaxError on line 7:
      | shared-photos = { $user_name { $photo_count ->
      …                              ^----- Expected token }
      |     [0] hasn't added any photos yet
      |     [one] added a new photo
      |    *[other] added { $photo_count } new photos
      | }.

    ! SyntaxError on line 17:
      | liked-comment = { $user_name } liked your comment on { $user_gender ->
      |     [male] his
      |     [female] her
      |     [other] their
      | } post.
      …                  ^----- Missing default variant

The last example shows a case where the `lineNumber` and `columnNumber` of the
error are wrong.  I'll fix this later in the parser.

### DIFF
--- a/fluent-syntax/src/ast.js
+++ b/fluent-syntax/src/ast.js
@@ -3,11 +3,12 @@ class Node {
 }
 
 export class Resource extends Node {
-  constructor(body = [], comment = null) {
-    super(body, comment);
+  constructor(body = [], comment = null, source = '') {
+    super();
     this.type = 'Resource';
     this.body = body;
     this.comment = comment;
+    this.source = source;
   }
 }
 
@@ -15,6 +16,15 @@ export class Entry extends Node {
   constructor() {
     super();
     this.type = 'Entry';
+    this.annotations = [];
+  }
+
+  addSpan(from, to) {
+    this.span = { from, to };
+  }
+
+  addAnnotation(annot) {
+    this.annotations.push(annot);
   }
 }
 
@@ -179,10 +189,10 @@ export class Function extends Identifier {
   }
 }
 
-export class JunkEntry extends Entry {
+export class Junk extends Entry {
   constructor(content) {
     super();
-    this.type = 'JunkEntry';
+    this.type = 'Junk';
     this.content = content;
   }
 }
--- a/fluent-syntax/src/errors.js
+++ b/fluent-syntax/src/errors.js
@@ -1,26 +1,18 @@
-export function error(ps, message) {
-  const err = new SyntaxError(message);
-  err.lineNumber = ps.getLineNumber();
-  err.columnNumber = ps.getColumnNumber();
-  return err;
+class Annotation {
+  constructor(message) {
+    this.message = message;
+  }
 }
 
-export function getErrorSlice(source, start, end) {
-  const len = source.length;
-
-  let startPos;
-  let sliceLen = end - start;
-
-  if (len < sliceLen) {
-    startPos = 0;
-    sliceLen = len;
-  } else if (start + sliceLen >= len) {
-    startPos = len - sliceLen - 1;
-  } else {
-    startPos = start;
+export class ParseError extends Annotation {
+  constructor(message) {
+    super(message);
+    this.name = 'ParseError';
   }
+}
 
-  return startPos > 0 ?
-    source.substr(startPos - 1, sliceLen) :
-    source.substr(0, sliceLen);
+export function error(ps, message) {
+  const err = new ParseError(message);
+  err.pos = ps.getIndex();
+  return err;
 }
--- a/fluent-syntax/src/parser.js
+++ b/fluent-syntax/src/parser.js
@@ -2,10 +2,9 @@
 
 import * as AST from './ast';
 import { FTLParserStream } from './ftlstream';
-import { error, getErrorSlice } from './errors';
+import { ParseError, error } from './errors';
 
 export function parse(source) {
-  const errors = [];
   let comment = null;
 
   const ps = new FTLParserStream(source);
@@ -22,17 +21,29 @@ export function parse(source) {
       if (entryStartPos === 0 && entry.type === 'Comment') {
         comment = entry;
       } else {
+        entry.addSpan(entryStartPos, ps.getIndex());
         entries.push(entry);
       }
-    } catch (e) {
-      entries.push(getJunkEntry(ps, source, entryStartPos));
-      errors.push(e);
+    } catch (err) {
+      if (!(err instanceof ParseError)) {
+        throw err;
+      }
+
+      ps.skipToNextEntryStart();
+      const nextEntryStart = ps.getIndex();
+
+      // Create a Junk instance
+      const slice = source.substring(entryStartPos, nextEntryStart);
+      const junk = new AST.Junk(slice);
+      junk.addSpan(entryStartPos, nextEntryStart);
+      junk.addAnnotation(err);
+      entries.push(junk);
     }
 
     ps.skipWSLines();
   }
 
-  return [new AST.Resource(entries, comment), errors];
+  return new AST.Resource(entries, comment, source);
 }
 
 function getEntry(ps) {
@@ -540,12 +551,4 @@ function getLiteral(ps) {
   const name = getIdentifier(ps);
   return new AST.MessageReference(name);
 
-}
-
-function getJunkEntry(ps, source, entryStart) {
-  ps.skipToNextEntryStart();
-
-  const slice = getErrorSlice(source, entryStart, ps.getIndex());
-
-  return new AST.JunkEntry(slice);
 }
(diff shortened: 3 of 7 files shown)


fluent-syntax/src/ast.js
```diff
@@ -179,10 +185,10 @@ export class Function extends Identifier {
   }
 }
 
-export class JunkEntry extends Entry {
+export class Junk extends Entry {
   constructor(content) {
     super();
-    this.type = 'JunkEntry';
+    this.type = 'Junk';
     this.content = content;
```
stasm: With each `Entry` now storing its `span`, is it OK to additionally store `content` in `Junk`?

fluent-syntax/src/errors.js
```diff
@@ -1,26 +1,7 @@
 export function error(ps, message) {
   const err = new SyntaxError(message);
```
stasm: Are `SyntaxErrors` OK for now, or would it be better to start with our own `Annotation` class here?

fluent-syntax/src/stream.js
```diff
@@ -24,6 +26,7 @@ export class ParserStream {
     }
 
     this.index++;
+    this.checkLineBreak();
```
zbraniecki: I think it's very risky. We're not guaranteed to always advance the stream with next, and we may want to have methods that allow us to skip chars in the future (like "skipToX").

I would much prefer to calculate the line at the time we encounter an error.

Since the error scenario is already an exception, I believe the performance doesn't matter, and just going through the source and counting number of '\n' in it should be enough).

zbraniecki: So, I guess I'd prefer to have an error method "getLineForPos(source, pos)" over this.

stasm: > I would much prefer to calculate the line at the time we encounter an error.

In the PR, I do that for every entry, valid or not. This will allow to easily print warnings for valid entries the same way errors are displayed right now. Do you think it's not worth it?

zbraniecki: I think it breaks the separation of concerns. In my design, Streams operate on... streams of characters. They don't keep the state of anything except the pointer in the Iterator and the buffer.

With your change, the stream becomes much more "aware". It now tracks state of the file. I feel like I'd prefer that not to happen in the stream for reasons mentioned above.

An example of where it would not work well is Rust where we can use CharStream to skip to a given character without using next. That would make the stream code per-implementation while so far I w as able to keep it parallel across implementations.

So, it's a bit hard, but I *think* that yes, I'd suggest taking line no for each entry separately, or, keep the pos in the AST and calculate line/col only when displaying.

stasm: Thanks, that's very helpful. I agree about the separations of concerns. Also, it seems like it would be easier to provide a helper for now and add `lineno` to the AST later than the other way around.

fluent-syntax/src/parser.js
```diff
@@ -15,24 +28,36 @@ export function parse(source) {
...
+
+      const spanTo = ps.getIndex();
+      const slice = source.substring(spanFrom, spanTo);
+      const junk = new AST.Junk(slice);
+      junk.addSpan(spanFrom, spanTo);
+      entries.push(junk);
+
+      errors.push(Object.assign(err, {
+        lineNumber: lineOffset(source, err.pos) + 1,
+        columnNumber: columnOffset(source, err.pos) + 1,
```
stasm: `lineNumber` and `columnNumber` are modelled after [[1]] but I just realized that in `Errors`, `lineNumber` is 1-based and `columnNumber` is 0-based :/

Python is more consistent, having a 1-based `lineno` and a 0-based `col_offset` [[2]].

@Pike, what do you think would be the most useful for tools? Having worked a bit with ACE, for instance, I'd prefer to go for offsets both for lines and columns.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Mozilla
[2]: https://docs.python.org/3/library/ast.html#ast.AST.lineno

fluent-syntax/src/parser.js
```diff
@@ -2,7 +2,20 @@
...
+  return source.substring(0, pos).split('\n').length - 1;
+}
+
+export function columnOffset(source, pos) {
+  const lastLineBreak = source.lastIndexOf('\n', pos);
+  return lastLineBreak === -1
+    ? pos
+    // Substracting two offsets gives length; substract 1 to get the offset.
+    : pos - lastLineBreak - 1;
+}
```
Pike: https://github.com/Pike/compare-locales/blob/master/compare_locales/parser.py#L187 is how I code this problem in compare-locales.

Effectively, I cache the position of line breaks in the source, and then use bisect to find which of those positions matter for my current pos. Based on the position of the found line break in the list I know the line number, based on the value, I can compute the column.

I recognize that this is a lot easier to implement in python, as there's no finditer in js regexp :-(

fluent-syntax/src/parser.js
Pike: In compare-locales, I use 1-based indices for both, because I want to give user-digestible positions.

I think toolchains are inconsistent enough that we don't need to care about matching what they expect.

fluent-syntax/src/parser.js
stasm: Interesting. I think we could implement it the compare-locales way, too. I think I'm going to keep `source.split` approach for now and consult with @zbraniecki about his preference, too.

fluent-syntax/src/parser.js
stasm: Good to know, thanks for weighing in. I'll go for offsets everywhere, because that's the easiest for my brain. Off-by-one errors are the worst :)

stasm:
> From a general tooling perspective, I'm not sure if errors being a global property is the best way to deal with this. In pontoon, you'd have the message, and you want to know if there's an error in it. I also failed to find the back-referencing thing from error to entry?

Each annotation (error) has an `entry` field referencing the entry it occurred in. For Pontoon, one easy solution would be to filter annotations given the current entry:

```javascript
function annotationsForEntry(annotations, entry) {
    return annotations.filter(
        annot => annot.entry === entry
    );
}
```

We can also consider adding a special-purpose `parseMessage` function which takes a string and returns a single `Message` instance and its annotations.

stasm:
That said, it would be just as easy to collect all annotations attached to individual entries.

stasm:
I updated https://github.com/projectfluent/fluent/wiki/Error-reporting with the description of the data model that landed in this PR.

@Pike mentioned on IRC that in compare-locales, each entry holds a reference to a parse content which in turn stores the original string of the resource. This might be something interesting to implement in `fluent-syntax` as well.  It would allow us to remove the `source` field from `Resource` and perhaps move the `lineOffset` and `columnOffset` helpers to getters on `Entry` or `Annotation`. @zbraniecki, if you think that's a good direction of the parser, let me know and I'll file a new issue.

Thanks all!

PR #8, opened 2017-02-20
Move Intl.PluralRules into its own package

stasm:
I'd like to move the `Intl` polyfills out of the core `fluent` package.

`NumberFormat` is [well-supported](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat#Browser_compatibility) and I think it's time to remove it.

`PluralRules` is being worked on: the proposal reached Stage 3 and there's an [implementation in Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1270146). By moving the polyfill out of the core package we will reduce the size of `fluent.js` (or `l20n.js`) shipping in Gecko. It also allows some users to use a different implementation or reduce the number of supported locales if needed.

In this PR, I moved the current polyfill into a new  package: `fluent-intl-polyfill`. However, I'd like to point out that our current polyfill doesn't support the entire proposal (`type: "ordinal"` and passing number formatting options aren't supported).  My recommendation right now would be to use [intl-pluralrules](https://www.npmjs.com/package/intl-pluralrules).

We can do this in two ways: either recommend `intl-pluralrules` in `fluent`'s README and _not publish_ `fluent-intl-polyfill`, or make `intl-pluralrules` its dependency.  The latter approach would be good in case we plan to start polyfilling other proposals. Perhaps `ListFormat` or `RelativeTimeFormat` once they reach a more mature Stage?

@zbraniecki what's your take on this? If you agree to use `intl-pluralrules` I'll be happy to change this PR.


heyakyra:
Is this the reason why in my js project, I'm unable to use the ordinal example at the end of the [selectors](https://projectfluent.org/fluent/guide/selectors.html) section of the guide?

PR #9, opened 2017-02-23
fluent-react

stasm:
`fluent-react` provides experimental integration between Fluent and React.  See https://github.com/projectfluent/fluent.js/wiki/React-bindings-for-Fluent and the examples/ directory and for an in-depth description.

### DIFF
--- a/fluent-react/README.md
+++ b/fluent-react/README.md
@@ -1,0 +1,126 @@
+# fluent-react
+
+`fluent-react` provides React bindings for Project Fluent, a localization
+framework designed to unleash the expressive power of the natural language.
+
+
+## Installation
+
+`fluent-react` can be used both on the client-side and the server-side.  You
+can install it from the npm registry or use it as a standalone script (as the
+`FluentReact` global).
+
+    npm install fluent-react
+
+
+## How to use
+
+`fluent-react` takes advantage of React's Components system and the virtual
+DOM.  Translations are stored using the Fluent syntax ([FTL][]) and exposed to
+components via the _provider_ pattern.
+
+An example FTL file might look like the following:
+
+```
+hello-world = Hello, world!
+welcome = Welcome, { $username }.
+```
+
+
+Wrap your localizable elements in `LocalizedElement`:
+
+```javascript
+import { LocalizedElement } from 'fluent-react';
+
+<LocalizedElement id="hello-world">
+    <p>Hello, world!</p>
+</LocalizedElement>
+```
+
+The `id` prop should be the unique identifier of the translation.  Any
+attributes found in the translation will be applied to the wrapped element.
+You can also pass arguments to the translation as `$`-prefixed props on
+`LocalizedElement`:
+
+```javascript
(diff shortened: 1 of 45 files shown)


fluent-react/README.md
```diff
@@ -0,0 +1,129 @@
...
+</LocalizedElement>
+```
+
+Notice how in the above example the text content of the `<p>` element is
+a string expression, not a text node.  It is provided as help for the developer
+and may be used in the future to extract the source copy. For now, however,
+it's not used for anything.  By putting the text in a simple string expression
+we're saving React's cycles: `<p>Welcome, { $username }</p>` would be rendered
+by React before being passed to `LocalizedElement` where it would be replaced
+by the translation anyways.
```
zbraniecki: I'm wondering if we have to do that.

It seems to me like that's a minor glitch between Fluent and React and React could solve it by allowing us to take precedence and resolve the localized content before rendering. I'd love to avoid having to do that just because React cannot allow us to execute code in the right order of operations.

fluent-react/README.md
```diff
@@ -0,0 +1,129 @@
...
+localized elements in the app without passing it explicitly.
+
+```javascript
+import 'fluent-intl-polyfill';
+import negotiateLanguages from 'fluent-langneg';
+import { LocalizationProvide } from 'fluent-react';
+
+ReactDOM.render(
+  <LocalizationProvider
+    locale={negotiateLanguages(navigator.language, ['en-US', 'de'])}
```
zbraniecki: that's `locales`, right?

stasm: yeah, it should be. I documented it in NOTES as something I want to do. I'll probably wait for `fluent-langneg`.

stasm: Update: I changed it to `locales`.

fluent-react/README.md
stasm: React gives us ways to control whether a components needs to update (`shouldComponentUpdate`). And because we integrate right into the virtual DOM, we actually avoid double-rendering. What we don't avoid is the evaluation of leaves of the tree before the parent.  So if you have `<p>Hello, { name.toUpperCase() }</p>`, `toUpperCase` will be called before the element makes it to `<LocalizedElement>`.  `<LocalizedElement>` will then replace the content with the translation and hand it off to React for rendering. So the rendering happens only once.

Putting `<p>Hello, { name }</p>` as the content looks good on paper and it makes it such this content can be used as an ultimate fallback, but it has a number of shortcomings:

  - (mentioned above) `{ name }` is evaluated before the localization logic happens,
  - `{ name }` is not the proper Fluent syntax for this message; it should be `{ $username }`; it's a coincidence that JSX also uses braces for interpolated expressions,
  - if the message needs plurals in the source language, you can't put them here,
  - it may be misleading: it's regular JSX so developers might be tempted to put more logic in it; think `{items.map(item => <MyItem>{item}</MyItem>) }`.

(I updated https://github.com/projectfluent/fluent.js/wiki/React-bindings-for-Fluent with these arguments.)

PR #12, opened 2017-03-02
Adapt runtime parser to handle spec 0.3

zbraniecki:
This is an initial patch. It handles new comments sygil, tags and multiline.


zbraniecki:
I have to say that it's becoming increasingly hard to work with this parser due to lack of testing infra.

I'd like to minimize the work on it until we get parser tests and then do a medium size refactor to get it optimized for the current spec.

PR #13, opened 2017-03-02
Update fluent-syntax parser to spec 0.3

zbraniecki:


fluent-syntax/src/parser.js
```diff
@@ -131,11 +134,15 @@ function getMessage(ps, comment) {
     attrs = getAttributes(ps);
   }
 
-  if (pattern === undefined && attrs === undefined) {
+  if (ps.isPeekNextLineTagStart()) {
+    tags = getTags(ps);
```
stasm: As per the [EBNF](https://github.com/projectfluent/fluent/blob/03f387ccab170b4b98756cc16d1d705f8ce4ad53/spec/fluent.ebnf#L43), a message is not allowed to have both tags and attributes. The allowed options are: value, value with tags, value with attributes and just attributes. I think we should throw an error here if `attrs` is already defined.

PR #16, opened 2017-03-08
fluent-merge

stasm:
`fluent-merge` is a simple command line utility for merging localization files. It uses the first resource as the reference. Messages from other resources which don't exist in the first resource are ignored.

In the future we can change the exact merging logic. For instance, we could follow the logic of `Object.assign` in ECMAScript, or additionally merge any messages referenced by other merged messages.

### DIFF
--- a/fluent-merge/src/index.js
+++ b/fluent-merge/src/index.js
@@ -1,0 +1,66 @@
+import { Resource, parse, serialize } from 'fluent-syntax';
+
+/*
+ * Merge text files into one file.
+ */
+export function mergeFiles(...files) {
+  const resources = files.map(parse);
+  const merged = resources.reduce(merge);
+  return serialize(merged);
+}
+
+/*
+ * Merge AST trees into one tree.
+ */
+export function mergeResources(...resources) {
+  const merged = resources.reduce(merge);
+  return merged;
+}
+
+/*
+ * Merge messages of two resources.
+ *
+ * The first resource is treated as the reference. Sections and Junk entries
+ * are ignored.
+ */
+function merge(res1, res2) {
+  const body = [];
+
+  for (const message of messages(res1)) {
+    const other = findMessage(res2, message.id.name);
+    if (other) {
+      body.push(other);
+    } else {
+      body.push(message);
+    }
+  }
+
+  const merged = new Resource(body, res1.comment);
+  merged.source = serialize(merged);
+
+  return merged;
+}
+
+/*
+ * Find a message in `res` by its `id`.
+ */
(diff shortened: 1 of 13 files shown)


stasm:
We'll need to release a new version of `fluent-syntax` in order for the tests to pass.

fluent-merge/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+  merged.source = serialize(merged);
+
+  return merged;
+}
+
+/*
+ * Find a message in `res` by its `id`.
+ */
+function findMessage(res, id) {
+  for (const other of messages(res)) {
```
zbraniecki: the `other` variable name strikes me as odd as I read through it. Why not just `msg`? I'm looking for a message, in `messages`.

fluent-merge/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+/*
+ * Create an iterator over Resource's messages.
+ */
+function* messages(res) {
+  for (const entry of res.body) {
+    if (entry.type === 'Message') {
+      yield entry;
+    }
+  }
+}
```
zbraniecki: Due to the way you're using it (iterate over `res1`, for each entry in it iterate over `res2`) and the fact that in most cases I'd expect that most entries will be in both, I'd say that `let msgs = res.body.filter(...)` done once should be way cheaper memory wise and faster.
But this will work, so up to you:)

fluent-merge/fluent-langneg.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+}
+
+exports.mergeFiles = mergeFiles;
+exports.mergeResources = mergeResources;
+
+Object.defineProperty(exports, '__esModule', { value: true });
+
+})));
```
zbraniecki: This file is confusing? Is it lang-neg? Seems like a copy of index.js

fluent-merge/src/index.js
stasm: Good point! `other` was probably an artifact of an earlier iteration.

fluent-merge/src/index.js
stasm: Can you explain how you'd use `filter`?

fluent-merge/fluent-langneg.js
stasm: It's very confusing :) I have no idea what it's doing here, I guess I accidentally committed it. Thanks for spotting it!

stasm:
Has it been almost a year already? :) Closing. We'll soon be using async iteration for this purpose.

PR #19, opened 2017-03-14
Improve Language Negotiation logic to match the C++ implementation

zbraniecki:
not finished yet


fluent-langneg/test/locale_test.js
```diff
@@ -51,11 +51,11 @@ suite('Parses simple locales', () => {
       variant: 'mac'
     }));
 
-    assert.ok(isLocaleEqual('lij-Arab-FA-linux', {
+    assert.ok(isLocaleEqual('lij-Arab-FA-lin', {
```
zbraniecki: We only allow the variant code to be 3 chars long. If we want to relax it, I'll be happy to use `-linux`.

Pike: According to https://tools.ietf.org/html/bcp47#section-2.2.5, that's a bug. [a-z] variants must be at least 5 characters long.

We should fix that on our gecko impl, and just whitelist 'mac' in ja-JP-mac as an exception.

zbraniecki: Added to https://bugzilla.mozilla.org/show_bug.cgi?id=1345957
I'll fix it in C++ and then backport here.

PR #22, opened 2017-03-30
Introduce behavior and structure testing

stasm:
#### Behavior Tests

FTL files which may include special markup with the information about the expected error.  The test runner uses a preprocessor to parse and remove the error annotation from the file.  If there is no error annotation, the file is expected to parse without errors.

#### Structure Tests

FTL + JSON AST pairs.  JSON files are generated via `make fixtures` and reviewed by a human.  The test runner compares the output of the parser against the saved AST.


stasm:
I filed https://github.com/projectfluent/fluent/pull/40 for ASDL changes.

fluent-syntax/src/errors.js
```diff
@@ -1 +1,51 @@
-export class ParseError extends Error {}
+export class ParseError extends Error {
+  constructor(code, ...args) {
+    super();
+    this.code = code;
+    this.message = getErrorMessage(code, args);
+  }
+}
+
+function getErrorMessage(code, args) {
```
stasm: The list of errors comes from https://github.com/projectfluent/fluent-rs/blob/master/src/syntax/errors/list.rs. It's a first draft and I'm sure the list will change and expand as we start writing more tests.

fluent-syntax/src/ast.js
```diff
@@ -227,11 +227,14 @@ export class Span extends Node {
...
     super();
     this.type = 'Annotation';
-    this.name = name;
+    this.code = code;
     this.message = message;
-    this.pos = pos;
+  }
+
+  addSpan(start, end) {
+    this.span = new Span(start, end);
```
stasm: I filed https://github.com/projectfluent/fluent/pull/40 for ASDL changes.

PR #27, opened 2017-04-14
fluent-react: LocalizationProvider

stasm:
This PR adds the `LocalizationProvider` component capable of synchronous fallback in case of missing translations.

All `<Localized>` components need access to an instance of the `Localization` class, the central localization store to which they subscribe.  The `<LocalizationProvider>` component implements the _provider_ pattern known from Redux.  Taking advantage of React's [context][] feature, it makes translations available to all localized elements in the app without passing them explicitly.

```javascript
ReactDOM.render(
    <LocalizationProvider messages={…}>
        …
    </LocalizationProvider>,
    document.getElementById('root')
);
```

The `LocalizationProvider` component takes one prop: `messages`.  It should be an iterable of `MessageContext` instances in order of user's preferred languages. This iterable will be used by `Localization` to format translations.  If a translation is missing in one language, `Localization` will fall back to the next locale.

In it's simplest form, `locales` can be an array:

```javascript
function generateMessages(currentLocales) {
    return currentLocales.map(locale => {
        const cx = new MessageContext(locale);
        cx.addMessages(MESSAGES_ALL[locale]);
        return cx;
    });
}

<LocalizationProvider messages={generateMessages(['en-US'])}>
    …
</LocalizationProvider>,
```

In order to avoid having to create all `MessageContext` instances for all locales up front, it's a good idea to make `messages` an iterator.  The `Localization` class will iterate over it (memoizing the items it yields) in case fallback is required.

```javascript
export function* generateMessages(currentLocales) {
    for (const locale of currentLocales) {
        const cx = new MessageContext(locale);
        cx.addMessages(MESSAGES_ALL[locale]);
        yield cx;
    }
}

<LocalizationProvider messages={generateMessages(['en-US'])}>
    …
</LocalizationProvider>,
```

The design of the `LocalizationProvider` requires a little bit of work from the developer.  The `messages` iterable needs to be created manually.  This is intentional: it gives the most control to the developer with regards to the following three areas:

 - translations - the developer decides where translations are stored and how they're fetched,
 - language negotiation - the developer decides which factors are taken into account for the purpose of the language negotiation, as well as how exactly it's being done (we recommend [fluent-langneg][]),
 - custom extensions - the developer can pass options to the `MessageContext` constructor to configure its behavior or to define functions available to translations.

In the future we might end up providing ready-made generators of the `messages` iterable for the most common scenarios.

One limitation of the current design is that in asynchronous scenarios, all translations (including any fallback) must be fetched at once before `<LocalizationProvider>` is rendered.  See the `*-async` examples in the `examples/` directory for more information.

In the future we might be able to allow async fetching of fallback locales. For now it is out of scope of this PR.

[context]: https://facebook.github.io/react/docs/context.html
[fluent-langneg]: https://github.com/projectfluent/fluent.js/tree/master/fluent-langneg


stasm:
I'll add a few more tests tomorrow. The core of the PR is ready. @zbraniecki, I'd like primarily to get your review on the changes to the `MessageContext` constructor (the first commit). Let me know what you think about the `LocalizationProvider` component (the second commit), too! Thanks.

Pike:
I'm hesitant about the use of `locale` vs `language` here. Do we have documented somewhere what we use when?

stasm:
Do you know any alternatives? I found `blanket` which isn't maintained anymore; the author recommends `istanbul`.

stasm:
I struggle with the distinction myself :) The way I understood @zbraniecki's explanation, is that `locale` is strictly a wider term than `language` and can include the information about h12/24 formatting, calendars etc. Thus, `navigator.languages` is actually an array of `locale` names. `negotiateLanguages` from `fluent-langneg` takes `locale` names as input, negotiates the `language` parts of them, and returns `locales` again. I expect that most of the time, the app will define its available _languages_ rather then _locales_, so `negotiateLanguages` will actually return an array of matching `languages` which also happen to be valid `locale` tags. @zbraniecki please correct me here if I'm wrong. Also, would you like to document the proper explanation somewhere on the wiki? Thanks!

PR #33, opened 2017-05-04
fluent-react: Introduce the withLocalization HOC

stasm:
The `withLocalization` decorator is a higher-order component which can be used to gain imperative access to a `Localization`.

Sometimes it's useful to imperatively format a translation rather than rely on the declarative `<Localized />` API.  Good examples of this are the `window.alert` and `window.confirm` functions.

The `withLocalization` HOC connects a component to its enclosing `LocalizationProvider` and injects the `formatString` prop into the wrapped component.  The `formatString` method returns a translation given an id and an optional arguments object.

```javascript
import { withLocalization } from 'fluent-react/compat';

function HelloButton(props) {
  const { formatString } = props;

  return (
    <button onClick={() => alert(formatString('hello'))}>
      👋
    </button>
  );
}

export default withLocalization(HelloButton);
```

### DIFF
--- a/fluent-react/CHANGELOG.md
+++ b/fluent-react/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+  - Added the `withLocalization` HOC.
+
+    It may be used to connect a component with a Localization. It injects the
+    `formatString` prop.
+
   - The compat build is now transpiled using rollup-plugin-babel.
 
     This ensures that the "use strict" pragma is scoped to the UMD wrapper.  It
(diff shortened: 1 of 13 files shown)


stasm:
@6a68 - Is this something that you think would be useful for showing localized messages via `window.alert` and `window.confirm`?

stasm:
@zbraniecki - I'd like to get your feedback on this as well. Am right to assume that this kind of an imperative API will be mostly used where only the value of a message is useful? Or, should we consider formatting attributes as well?

zbraniecki:
I think it's a good start. All current uses that I know of want just a single value. I can imagine that changing so I'd suggest designing the API that can be further extended to cover compound messages later.

PR #35, opened 2017-05-15
FluentDOM

zbraniecki:
Add FluentDOM with web example.


fluent-dom/src/localization.js
```diff
@@ -0,0 +1,120 @@
...
+
+
+  formatValues(...keys) {
+    const keyTuples = keys.map(
+      key => Array.isArray(key) ? key : [key, null]
+    );
+    return this.formatWithFallback(keyTuples, valueFromContext);
+  }
+
+  formatValue(id, args) {
```
stasm: Can you rewrite this to use `async/await`? I'm also considering a rename to `formatString(id, args)`, but I'm fine doing that in a follow-up.

fluent-dom/src/registry.js
```diff
@@ -0,0 +1,24 @@
+import { MessageContext } from '../../fluent/src/index';
```
stasm: Remove this file from this PR, please. It should be part of the planned `fluent-web` package which glues other `fluent-*` packages together for the purpose of using them in HTML documents.

fluent-dom/examples/simple.html
```diff
@@ -0,0 +1,18 @@
+<html>
```
stasm: Remove the example from this PR.

fluent-dom/src/localization.js
```diff
@@ -0,0 +1,120 @@
...
+          formattedAttr
+        ]);
+      }
+    }
+  }
+
+  return formatted;
+}
+
+function keysFromContext(method, ctx, keys, translations) {
```
stasm: I would like to refactor this code in a follow-up, so that we can measure the performance characteristics.  In essence, I'd like the `Localization` class to manage fallback on the context level via a `getMessageContext(id)` method.

fluent-dom/.eslintrc.json
```diff
@@ -0,0 +1,31 @@
+{
```
zbraniecki: To support async/await, but I think I'll enable it in the main .eslintrc instead

zbraniecki:
So, seems like jsdoc fails on async/await :( @stasm - what should I do here?

stasm:
It looks like it's a known issue: https://github.com/jsdoc3/jsdoc/issues/1188. There are very recent comments there. The parser jsdoc uses has already been fixed so now it's a question of when jsdoc updates.

There are three paths for a temporary work-around:

 1. Use the [jsdoc-strip-async-await](https://github.com/jcowgar/jsdoc-strip-async-await) jsdoc plugin.
 2. Or use the more general [jsdoc-babel](https://github.com/ctumolosus/jsdoc-babel) jsdoc plugin.
 3. Or switch the jsdoc dependency to this fork for the time being: https://github.com/jvanveen/jsdoc/commit/a789c902c56e7852ecff1825dc4f5d8fcd36a5ed

zbraniecki:
I tried with the plugin, but I like the fork commit solution better :)

zbraniecki:
yay, green travis! :)

fluent-dom/src/document_localization.js
```diff
@@ -0,0 +1,179 @@
...
+
+  translateRoots() {
+    const roots = Array.from(this.roots);
+    return Promise.all(
+      roots.map(root => this.translateRoot(root))
+    );
+  }
+
+  translateRoot(root) {
+    return this.translateFragment(root);
```
stasm: IIRC, `translateRoot` was also responsible for setting the `lang` and `dir` attributes on the root node and dispatching the `DOMRetranslated` event.

fluent-dom/src/document_localization.js
```diff
@@ -0,0 +1,179 @@
...
+
+  getKeysForElement(element) {
+    return [
+      element.getAttribute('data-l10n-id'),
+      JSON.parse(element.getAttribute('data-l10n-args') || null)
+    ];
+  }
+}
+
+function overlayElement(element, translation) {
```
stasm: Why not use the actual `overlayElement` from l20n.js?

fluent-dom/src/document_localization.js
```diff
@@ -0,0 +1,179 @@
...
+      attribute: true,
+      characterData: false,
+      childList: true,
+      subtree: true,
+      attributeFilter: ['data-l10n-id', 'data-l10n-args']
+    };
+  }
+
+  onLanguageChange() {
+    super.onLanguageChange();
```
stasm: `Localization` doesn't seem to implement `onLanguageChange`. Also, it might be a good idea to make `DocumentLocalization` implement the `EventListener` interface via the `handleEvent` method.

fluent-dom/src/document_localization.js
zbraniecki: I think this should be done by fluent-web.

zbraniecki:
@stas, I applied your feedback as discussed, including merging `Localization` into `DocumentLocalization`.

I have to say that I still feel that this is a wrong move and the resulting 600 lines file is mixing DOM and non-dom related code, but I also believe that this code should represent your vision for how Fluent will localize DOM, so I did my best to merge the code and apply as much of the l20n.js docstrings as possible.

The resulting code is much bigger and more complicated (in my opinion) than the source code I took from my Gecko patches, and since it follows your, not mine, vision, I'd prefer if you took it over if you want more architectural changes to be made.

I have to now go back and see if the code produced by this is at all applicable onto Gecko and I'm seriously worried about the burden on any Gecko engineer when working with this (either to review this, or help us make this code or Gecko faster). :(

fluent-dom/src/document_localization.js
stasm: Can you elaborate why? Setting the attributes is essential for the completeness of the localized DOM. Dispatching the event is harmless in the inert scenario (build-time or server-side rendering); moving it to `fluent-web` will mix concerns.

zbraniecki: interesting. I have the opposite view - that DocumentLocalization does not interact with outside world (doesn't set event listeners etc.) and only has methods to operate on the DOM provided to it.

But another concern is that at the moment DocumentLocalization does not know which languages it is in.

fluent-dom/src/localization.js
```diff
@@ -0,0 +1,287 @@
...
+    const [val] = await this.formatValues([[id, args]]);
+    return val;
+  }
+
+  handleEvent() {
+    this.onLanguageChange();
+  }
+
+  onLanguageChange() {
+    this.ctxs = this.generateContexts(this.id, this.resIds);
```
stasm: Use `generateMessages` here.

fluent-dom/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+On top of that, DocumentLocalization can localize any DOMFragment by
+identifying localizable elements with `l10n-id` and translating them.
+
+```javascript
+
+const l10n = new DocumentLocalization(document, [
+  '/browser/main.ftl',
+  '/toolkit/menu.ftl'
+], generatemessage);
```
stasm: `generateMessages`

fluent-dom/README.md
```diff
@@ -0,0 +1,76 @@
...
+  '/toolkit/menu.ftl'
+], generateMessages);
+
+async function main() {
+  const msg = await l10n.formatValue('welcome', { name: 'Anna' });
+  // → 'Welcome, Anna!'
+}
+```
+
+On top of that, DocumentLocalization can localize any DOMFragment by
```
stasm: Fix the repetition of `On top of that`

fluent-dom/README.md
```diff
@@ -0,0 +1,76 @@
...
+identifying localizable elements with `l10n-id` and translating them.
```
stasm: `data-l10n-id`

fluent-dom/README.md
stasm: Or simply remove the initialization here. This snippet may be treated as the continuation of the previous one.

fluent-dom/src/dom_localization.js
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  onLanguageChange() {
+    super.onLanguageChange();
+    this.translateDocument();
+  }
+
+  /**
+   * Set the `data-l10n-id` and `data-l10n-args` attributes on DOM elements.
+   * L20n makes use of mutation observers to detect changes to `data-l10n-*`
```
stasm: Remove `L20n` here and in other places.

fluent-dom/src/dom_localization.js
```diff
@@ -0,0 +1,306 @@
...
+   * @returns {Promise}
+   */
+  translateDocument() {
+    return this.translateRoots();
+  }
+
+  /**
+   * Translate a DOM element or fragment asynchronously using this
+   * `DocumentLocalization` and any `DOMLocalization` objects which it can
+   * delegate to.
```
stasm: Remove the mention about delegation.

fluent-dom/src/dom_localization.js
```diff
@@ -0,0 +1,306 @@
...
+ * formatting of translations and methods for observing DOM
+ * trees with a `MutationObserver`.
+ */
+export default class DOMLocalization extends Localization {
+  /**
+   * @returns {DocumentLocalization}
+   */
+  constructor(doc, resIds, generateMessages) {
+    super(doc.location.href, resIds, generateMessages);
+    this.document = doc;
```
stasm: Is `this.document` used anywhere?

fluent-dom/src/localization.js
```diff
@@ -0,0 +1,286 @@
...
+ *   hasFatalErrors: boolean,
+ *   translations: Array<string>|Array<{value: string, attrs: Object}>}} prev
+ *
+ * @returns {{
+ *   errors: Array<Error>,
+ *   withoutFatal: Array<boolean>,
+ *   hasFatalErrors: boolean,
+ *   translations: Array<string>|Array<{value: string, attrs: Object}>}}
+ */
+function keysFromContext(method, ctx, keys, translations) {
```
stasm: Please re-add the full logic and the comments from l20n.js in this method.

fluent-dom/src/dom_localization.js
```diff
@@ -0,0 +1,306 @@
...
+    this.query = '[data-l10n-id]';
+
+    // A Set of DOM trees observed by the `MutationObserver`.
+    this.roots = new Set();
```
stasm: Should we initialize `this.roots` with the `document` passed to the constructor?

fluent-dom/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+```javascript
+
+const l10n = new DocumentLocalization(document, [
+  '/browser/main.ftl',
+  '/toolkit/menu.ftl'
+], generatemessage);
+
+l10n.translateDocument();
```
stasm: How about we make this example the first one? It shows off the declarative API which is the one that we recommend over the imperative one.

PR #38, opened 2017-05-18
Add Headers parsing function

### LINKED ISSUE #34: langneg: add Accept-Language header parsing support
Since fluent-langneg runs on both server and client, it would be nice to not have to separately parse the Accept-Language header string into an array of locales:

```js
negotiateLanguages(req.header('Accept-Language'), availableLocales)
```

As things stand, I've either got to add another language negotiation library, like accept-language-parser, or roll my own, probably buggy, regex to do the job ;-)

zbraniecki:
Fixes #34

### DIFF
--- a/fluent-langneg/src/accepted_languages.js
+++ b/fluent-langneg/src/accepted_languages.js
@@ -1,10 +1,8 @@
 
-export default function aceptedLanguages(string = '') {
+export default function acceptedLanguages(string = '') {
   if (typeof string !== 'string') {
     throw new TypeError('Argument must be a string');
   }
   const tokens = string.split(',').map(t => t.trim());
-  return tokens.filter(t => t !== '').map(t => {
-    return t.split(';')[0];
-  });
+  return tokens.filter(t => t !== '').map(t => t.split(';')[0]);
 }
--- a/fluent-langneg/src/accepted_languages.js
+++ b/fluent-langneg/src/accepted_languages.js
@@ -1,0 +1,10 @@
+
+export default function aceptedLanguages(string = '') {
+  if (typeof string !== 'string') {
+    throw new TypeError('Argument must be a string');
+  }
+  const tokens = string.split(',').map(t => t.trim());
+  return tokens.filter(t => t !== '').map(t => {
+    return t.split(';')[0];
+  });
+}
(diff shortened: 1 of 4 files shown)


fluent-langneg/src/headers.js
```diff
@@ -0,0 +1,8 @@
+
+export default function parseAcceptedLanguagesHeaderIntoArray(string = '') {
+  if (typeof string !== 'string') throw new TypeError('Argument must be a string');
+  const tokens = string.split(',').map(t => t.trim());
+  return tokens.filter(t => t !== '').map(t => {
+    return t.split(';')[0];
```
stasm: This doesn't take into effect the q-value nor the specificity of the locale tag. Are we fine with this?

fluent-langneg/src/headers.js
```diff
@@ -0,0 +1,8 @@
+
+export default function parseAcceptedLanguagesHeaderIntoArray(string = '') {
```
stasm: Rename this simply to `acceptedLanguages` please.

fluent-langneg/src/headers.js
```diff
@@ -0,0 +1,8 @@
...
+  if (typeof string !== 'string') throw new TypeError('Argument must be a string');
```
stasm: Please break this into an `if .. ` and a `{}` block.

stasm:
And maybe rename the file to `accepted_languages.js` :)

fluent-langneg/src/headers.js
zbraniecki: yes, this is on purpose. We don't act upon the q-value at all. The order is the only indication for our fallback, there's no increase/decrease in fallback based on the q-value.

zbraniecki:
Applied feedback

fluent-langneg/src/headers.js
stasm: Do you plan to extend this in the future? The RFC says the q-values are in fact important.

fluent-langneg/src/accepted_languages.js
```diff
@@ -0,0 +1,10 @@
+
+export default function aceptedLanguages(string = '') {
```
stasm: Typo: `s/acepted/accepted/`.

fluent-langneg/src/accepted_languages.js
```diff
@@ -0,0 +1,10 @@
...
+  if (typeof string !== 'string') {
+    throw new TypeError('Argument must be a string');
+  }
+  const tokens = string.split(',').map(t => t.trim());
+  return tokens.filter(t => t !== '').map(t => {
+    return t.split(';')[0];
+  });
```
stasm: This is a nit, but there's no need to use `{}` here and an explicit `return`. Can you change this to a one-liner arrow function?

PR #40, opened 2017-05-23
Introduce the MessageSyncSequence class.

stasm:
The `MessageSyncSequence` class manages an ordered sequence of `MessageContexts`.

The current negotiated fallback chain of languages is stored in form of an
iterable of `MessageContext` instances.  This iterable is used to find the
best existing translation for a given identifier.

    const seq = new MessageSyncSequence();
    seq.setIterable([ctx1, ctx2]);

Use the `mapContext` method to find the first `MessageContext` which
contains the translation with the given identifier.  If the sequence
is ordered according to the result of a language negotiation the returned
`MessageContext` contains the best available translation.

A simple function which formats translations based on the identifier might
be implemented as follows:

    formatString(id, args) {
        const ctx = seq.mapContext(id);

        if (ctx === null) {
            return id;
        }

        const msg = ctx.getMessage(id);
        return ctx.format(msg, args);
    }

The iterable passed to `MessageSyncSequence.setIterable` can also be an iterator.
The elements it yields will be cached by the `MessageSequence`.  This allows
multiple calls to `mapContext` without advancing and eventually
depleting the iterator.

    function *generateMessages() {
        // Some lazy logic for yielding MessageContexts.
        yield *[ctx1, ctx2];
    }

    const seq = new MessageSyncSequence();
    seq.setIterable(generateMessages());

### DIFF
--- a/fluent/CHANGELOG.md
+++ b/fluent/CHANGELOG.md
@@ -2,7 +2,50 @@
 
 ## Unreleased
 
-  - …
+  - Introduce the MessageSyncSequence class.
+
+    The MessageSyncSequence class manages an ordered sequence of
+    MessageContexts.
+
+    The current negotiated fallback chain of languages is stored in form of an
+    iterable of MessageContext instances.  This iterable is used to find the
+    best existing translation for a given identifier.
+
+        const seq = new MessageSyncSequence();
+        seq.setIterable([ctx1, ctx2]);
+
+    Use the mapContext method to find the first MessageContext which contains
+    the translation with the given identifier.  If the sequence is ordered
+    according to the result of a language negotiation the returned
+    MessageContext contains the best available translation.
+
+    A simple function which formats translations based on the identifier might
+    be implemented as follows:
+
+        formatString(id, args) {
+            const ctx = seq.mapContext(id);
+
+            if (ctx === null) {
+                return id;
+            }
+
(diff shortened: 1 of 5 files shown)


stasm:
I'm looking for feedback on naming, ccing @zbraniecki and @Pike. Please help :)

The class that I'm introducing in this PR is an abstraction for an ordered and lazy list of `MessageContext` instances. It essentially models the fallback strategy in case of missing translations.  The `Localization` class that we know from the DOM and React bindings can be its subclass which adds anything required by the bindings.

In this PR I'm only introducing the sync version of the class which means that the iterable used to create the sequence must be a sync iterable. In the future I'd like to be able to introduce an async counterpart which will be using async iterables.

For now the working name is `MessageSyncSequence`.  Other ideas include:

    SyncFallbackSequence, ContextSequenceSync, MessageContextSyncIter

stasm:
I talked to @Pike otp, and he pointed out that since `setIterable` and `mapContext` are not supposed to be exposed as public API on subclasses of `MessageSyncSequence` (such as `Localization`), they shouldn't be subclasses :)

Out current thinking is to get rid of the class altogether and instead implement this functionality with two functions: `cacheIterable(iterable)` and `mapContext(iterable, ids)`.

stasm:
For posterity, I'm going to close this PR and submit a new one with the updated code.

PR #41, opened 2017-05-30
Introduce mapContextSync and CachedIterable.

stasm:
(Another take on #40.)

An ordered iterable of MessageContext instances can represent the
current negotiated fallback chain of languages.  This iterable can be
used to find the best existing translation for a given identifier.

The mapContext* methods can be used to find the first MessageContext in
the given iterable which contains the translation with the given
identifier.  If the iterable is ordered according to the result of
a language negotiation the returned MessageContext contains the best
available translation.

A simple function which formats translations based on the identifier
might be implemented as follows:

    formatString(id, args) {
        const ctx = mapContextSync(contexts, id);

        if (ctx === null) {
            return id;
        }

        const msg = ctx.getMessage(id);
        return ctx.format(msg, args);
    }

In order to pass an iterator to mapContext*, wrap it in CachedIterable.
This allows multiple calls to mapContext* without advancing and
eventually depleting the iterator.

    function *generateMessages() {
        // Some lazy logic for yielding MessageContexts.
        yield *[ctx1, ctx2];
    }

    const contexts = new CachedIterable(generateMessages());
    const ctx = mapContextSync(contexts, id);

### DIFF
--- a/fluent/CHANGELOG.md
+++ b/fluent/CHANGELOG.md
@@ -2,7 +2,44 @@
 
 ## Unreleased
 
-  - …
+  - Introduce mapContextSync and CachedIterable.
+
+    An ordered iterable of MessageContext instances can represent the
+    current negotiated fallback chain of languages.  This iterable can be
+    used to find the best existing translation for a given identifier.
+
+    The mapContext* methods can be used to find the first MessageContext in
+    the given iterable which contains the translation with the given
+    identifier.  If the iterable is ordered according to the result of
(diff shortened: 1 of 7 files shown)


stasm:
> Thinking about mapContext*, ctx.format can raise runtime errors, right? And that we need to continue in the fallback chain if that happens? Requesting changes for that.

The errors raised by `format` are not in the scope of this PR. Nor do I think that we should fall back to different languages and context when they happen. We have good runtime error handling and translation salvaging built into `format` for this.

PR #45, opened 2017-06-05
Add Spans to all Nodes

stasm:
Most AST nodes can now have a Span.  Use `new FluentParser({ withSpans: true })` to enable this behavior. The `withAnnotations` config option to `FluentParser` has been removed. The parser always produces Annotations if necessary now.


stasm:
Following what I did in python-fluent, I separated the changes into two commits here as well. It has the benefit of making the history of the repository cleaner and potentially improving future blames.

PR #47, opened 2017-06-25
Fix a couple AST mistakes caught while porting the parser to Rust

zbraniecki:
Minor improvements to the parser and aligning error messages with the actual code. Found during porting to Rust.

### DIFF
--- a/fluent-syntax/src/parser.js
+++ b/fluent-syntax/src/parser.js
@@ -153,7 +153,7 @@ export default class FluentParser {
 
       ps.next();
 
-      if (ps.current() === '/') {
+      if (ps.currentIs('/')) {
         content += '\n';
         ps.next();
         ps.expectChar('/');
@@ -212,7 +212,7 @@ export default class FluentParser {
       tags = this.getTags(ps);
     }
 
-    if (pattern === undefined && attrs === undefined && tags === undefined) {
+    if (pattern === undefined && attrs === undefined) {
       throw new ParseError('E0005', id.name);
     }
 
@@ -566,7 +566,14 @@ export default class FluentParser {
 
       ps.expectChar(')');
 
-      return new AST.CallExpression(literal.id, args);
+      if (!/^[A-Z_-]+$/.test(literal.id.name)) {
+        throw new ParseError('E0008');
+      }
+
+      return new AST.CallExpression(
+        new AST.Function(literal.id.name),
+        args
+      );
     }
 
     return literal;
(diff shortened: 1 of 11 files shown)


fluent-syntax/src/ast.js
```diff
@@ -210,13 +210,6 @@ export class Section extends Entry {
   }
 }
 
-export class Function extends Identifier {
```
stasm: Removing this doesn't look right. `Function` is a valid ASDL production: https://github.com/projectfluent/fluent/blob/a9fb740ecbf9326a484b812800f62a338a43f7da/spec/fluent.asdl#L72

Instead, I think when we create a `CallExpression`, we should ensure that the `callee` field is a `Function`, not just an `Identifier`: https://github.com/projectfluent/fluent.js/blob/3e8066cfa81b31b4d46aec99a74e477792c61f5b/fluent-syntax/src/parser.js#L569

fluent-syntax/src/parser.js
```diff
@@ -212,7 +212,7 @@ export default class FluentParser {
       tags = this.getTags(ps);
     }
 
-    if (pattern === undefined && attrs === undefined && tags === undefined) {
```
stasm: Why are you removing the `tags` check here?

zbraniecki: My understanding was that we don't accept messages with just tags (they have to have attrs or value). Am I wrong?

stasm: No, you're right! That's a good point. The EBNF confirms this: https://github.com/projectfluent/fluent/blob/master/spec/fluent.ebnf#L43. Thanks for correcting me.

zbraniecki:
updated to your feedback!

fluent-syntax/src/parser.js
```diff
@@ -566,7 +566,10 @@ export default class FluentParser {
 
       ps.expectChar(')');
 
-      return new AST.CallExpression(literal.id, args);
+      return new AST.CallExpression(
+        new AST.Function(literal.id.name),
```
stasm: Can you add an assertion here to make sure `literal.id.name` conforms to the EBNF? It should only accept [`[A-Z_?-]+`](https://github.com/projectfluent/fluent/blob/a9fb740ecbf9326a484b812800f62a338a43f7da/spec/fluent.ebnf#L28).

zbraniecki: how would you like the assertion to look like? A regexp?

zbraniecki:
Re-requesting review because of the added check in the parser and test.

I decided not to add `?` because it's not supported anywhere in the parser, I raised a suggestion to remove it from the spec - https://github.com/projectfluent/fluent/issues/49 - and if this suggestion will get rejected, we will have to update all parsers, so we can fix it here then as well.

zbraniecki:
PR: https://github.com/projectfluent/python-fluent/pull/18

PR #54, opened 2017-08-08
Apply feedback from :mossop

zbraniecki:
Nothing major, just adding `cachedIterable` into localization.js (for now) and some docs updates, but since I'm bumping jsdoc version, I'd like to get your r+ :)

### DIFF
--- a/fluent-dom/src/localization.js
+++ b/fluent-dom/src/localization.js
@@ -1,28 +1,44 @@
 /* eslint no-console: ["error", { allow: ["warn", "error"] }] */
 /* global console */
 
+import CachedIterable from '../../fluent/src/cached_iterable';
+
+/**
+ * Specialized version of an Error used to indicate errors that are result
+ * of a problem during the localization process.
+ *
+ * We use them to identify the class of errors the require a fallback
+ * mechanism to be triggered vs errors that should be reported, but
+ * do not prevent the message from being used.
+ *
+ * An example of an L10nError is a missing entry.
+ */
 class L10nError extends Error {
-  constructor(message, id, lang) {
+  constructor(message) {
     super();
     this.name = 'L10nError';
     this.message = message;
-    this.id = id;
-    this.lang = lang;
   }
 }
 
 /**
- * The `Localization` class is responsible for fetching resources and
- * formatting translations.
+ * The `Localization` class is a central high-level API for vanilla
+ * JavaScript use of Fluent.
+ * It combines language negotiation, MessageContext and I/O to
+ * provide a scriptable API to format translations.
  */
 export default class Localization {
   /**
+   * @param {Array<String>} resourceIds      - List of resource IDs
+   * @param {Function}      generateMessages - Function that returns a
+   *                                           generator over MessageContexts
+   *
    * @returns {Localization}
    */
-  constructor(resIds, generateMessages) {
-    this.resIds = resIds;
+  constructor(resourceIds, generateMessages) {
+    this.resourceIds = resourceIds;
     this.generateMessages = generateMessages;
-    this.ctxs = this.generateMessages(this.resIds);
+    this.ctxs = new CachedIterable(this.generateMessages(this.resourceIds));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


fluent-dom/src/localization.js
```diff
@@ -1,28 +1,44 @@
 /* eslint no-console: ["error", { allow: ["warn", "error"] }] */
 /* global console */
 
+import CachedIterable from '../../fluent/src/cached_iterable';
```
stasm: Why can't we leave `CachedIterable` out of this file and keep it as a responsibility of the glue code?

package.json
```diff
@@ -15,7 +15,7 @@
     "eslint": "^3.19.0",
     "fuzzer": "^0.2.1",
     "gh-pages": "^0.12.0",
-    "jsdoc": "git+https://github.com/jvanveen/jsdoc.git#a789c902c56e7852ecff1825dc4f5d8fcd36a5ed",
+    "jsdoc": "^3.5.4",
```
stasm: Awesome to see JSDoc update the parser (https://github.com/jsdoc3/jsdoc/blob/master/CHANGES.md#354-august-2017).

fluent-dom/src/localization.js
```diff
@@ -82,12 +99,10 @@ export default class Localization {
...
    * Retrieve translations corresponding to the passed keys.
    *
-   * A generalized version of `DOMLocalization.formatValue`. Keys can
-   * either be simple string identifiers or `[id, args]` arrays.
-   *
    *     docL10n.formatValues([
    *       ['hello', { who: 'Mary' }],
-   *       ['hello', { who: 'John' }]
+   *       ['hello', { who: 'John' }],
+   *       ['welcome']
```
stasm: This could be written simply as `'welcome'` to illustrate the point about passing simple strings.

fluent-dom/src/localization.js
```diff
@@ -82,12 +99,10 @@ export default class Localization {
   /**
    * Retrieve translations corresponding to the passed keys.
    *
-   * A generalized version of `DOMLocalization.formatValue`. Keys can
-   * either be simple string identifiers or `[id, args]` arrays.
```
stasm: Don't remove the `Keys can..` sentence.

fluent-dom/src/localization.js
```diff
@@ -1,28 +1,44 @@
...
+   * @param {Array<String>} resourceIds      - List of resource IDs
+   * @param {Function}      generateMessages - Function that returns the
+   *                                           generator over MessageContexts
+   *
    * @returns {Localization}
    */
-  constructor(resIds, generateMessages) {
-    this.resIds = resIds;
+  constructor(resourceIds, generateMessages) {
+    this.resourceIds = resourceIds;
```
stasm: Thumbs up to renaming `resIds` to `resourceIds`.

fluent-dom/src/localization.js
```diff
@@ -1,28 +1,44 @@
...
- * formatting translations.
+ * The `Localization` class is a central high-level API for vanilla
+ * JavaScript use of Fluent.
+ * It combines language negotiation, MessageContext and I/O to
+ * provide a scriptable API to format translations.
  */
 export default class Localization {
   /**
+   * @param {Array<String>} resourceIds      - List of resource IDs
+   * @param {Function}      generateMessages - Function that returns the
```
stasm: `a generator`

fluent-dom/src/localization.js
```diff
@@ -1,28 +1,44 @@
...
+ * An example of an L10nError is a missing entry.
+ */
 class L10nError extends Error {
-  constructor(message, id, lang) {
+  constructor(message) {
     super();
     this.name = 'L10nError';
     this.message = message;
-    this.id = id;
-    this.lang = lang;
```
stasm: Thanks for removing these.

fluent-dom/src/localization.js
zbraniecki: Because there's no glue code in Localization.jsm. As Mossop noted, we never allow for non-cached iterable in it.

fluent-dom/src/localization.js
zbraniecki: Except that they cannot be simple strings. We do not support it at the moment :)

fluent-dom/src/localization.js
zbraniecki: That doesn't work.

zbraniecki:
I updated the patch to add the `formatValues` ability to handle string keys.

fluent-dom/src/localization.js
stasm: Ah, right, I was confused by l20n.js. Sorry, my bad!

stasm:
> I updated the patch to add the formatValues ability to handle string keys.
 
Actually, please revert this. I think it's better to not allow it right now. I was confused by the old code in l20n.js, my bad!

PR #55, opened 2017-08-08
Add eager parameter to the constructor of CachedIterable

zbraniecki:
This will allow both fluent-web and fluent-gecko to trigger the initial fetch as soon as the iterable is created.


stasm:
Please amend the commit message before merging this.

PR #58, opened 2017-08-11
Use relative imports in fluent-web and fluent-dom

### LINKED ISSUE #57: Use relative imports in fluent-web and fluent-dom
Both `fluent-web` and `fluent-dom` are at versions 0.0.1. I expect the API to fluctuate before it stabilizes.  Let's use relative imports for the while being so that it's easier to iterate.

Once we reach 0.1, we can bring back package imports and proper dependency management via `package.json`.

stasm:
Fixes #57.

Both fluent-web and fluent-dom are at versions 0.0.1. I expect the API
to fluctuate before it stabilizes. Let's use relative imports for the
while being so that it's easier to iterate.

Once we reach 0.1, we can bring back package imports and proper
dependency management via package.json.

Along with this, this PR cleans up a few things:

  - I removed `fetchSync` intorduced in 60eb916

  - I renamed `resIds` to `resourceIds` in line with the change from
    fb50012

  - I removed `CachedIterable` from fluent-web; as of 5f45963 it is part
    of fluent-dom.

### DIFF
--- a/fluent-dom/src/localization.js
+++ b/fluent-dom/src/localization.js
@@ -1,7 +1,7 @@
 /* eslint no-console: ["error", { allow: ["warn", "error"] }] */
 /* global console */
 
-import CachedIterable from '../../fluent/src/cached_iterable';
+import { CachedIterable } from '../../fluent/src/index';
 
 /**
  * Specialized version of an Error used to indicate errors that are result
--- a/fluent-web/package.json
+++ b/fluent-web/package.json
@@ -32,7 +32,6 @@
   },
   "dependencies": {
     "fluent": "^0.4.0",
-    "fluent-langneg": "^0.0.2",
-    "fluent-dom": "0.0.1"
+    "fluent-langneg": "^0.0.2"
   }
 }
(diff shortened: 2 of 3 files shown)


zbraniecki:
Removing CachedIterable and fluentSync makes fluent-web impossible to operate without FOUC. I added those two to allow for fluent-web to work without FOUC without manual firstPaint management from the user.

stasm:
Unless I'm mistaken, we don't need `CachedIterable` _in_ fluent-web, because you added it to fluent-dom. So removing it here just fixes the redundancy.

Using sync loading is a no-go. We just can't be doing that on the web. Sync XHR is deprecated. We need to find other solutions to fixing FOUC.

zbraniecki:
Interesting. I can swear that when I described it on the weekly call (I'm working on getting fluent-web to work without fouc, sync xhr seems to be required for that) you did agree on my course of action.

Clearly I misunderstood you. Apologies.

Unfortunately I do not believe that there is other way. My intention was to add sync fetch as an option. Yes. It is deprecated, but it also will not be removed from any browser on the next 5-10 years due to widespread use.
My idea was to offer it as an example of how a most plain HTML translation can work, and notify users that they should control the firstpaint manually and switch to async.
In the long term, my hope is that W3C will provide methods to control firstpaint and we will switch to it for fluent-web.

stasm:
In the interest of landing this PR, I removed the `fetchSync` change and filed #59.

PR #63, opened 2017-08-18
Apply :mossop feedback on DOMLocalization

zbraniecki:
Based on https://bugzilla.mozilla.org/show_bug.cgi?id=1347799

### DIFF
--- a/fluent-dom/src/dom_localization.js
+++ b/fluent-dom/src/dom_localization.js
@@ -1,10 +1,10 @@
 import overlayElement from './overlay';
 import Localization from './localization';
 
-const l10nIdAttrName = 'data-l10n-id';
-const l10nArgsAttrName = 'data-l10n-args';
+const L10NID_ATTR_NAME = 'data-l10n-id';
+const L10NARGS_ATTR_NAME = 'data-l10n-args';
 
-const l10nElementQuery = `[${l10nIdAttrName}]`;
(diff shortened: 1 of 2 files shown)


fluent-dom/src/overlay.js
```diff
@@ -143,7 +142,7 @@ function overlay(sourceElement, translationElement) {
   // XXX Attributes previously set here for another language should be
   // cleared if a new language doesn't use them; https://bugzil.la/922577
   if (translationElement.attributes) {
-    for (k = 0, attr; (attr = translationElement.attributes[k]); k++) {
+    for (const attr of Array.from(translationElement.attributes)) {
```
zbraniecki: I had to use the `Array.from` here because otherwise jsdom complained.

fluent-dom/src/dom_localization.js
```diff
@@ -1,6 +1,11 @@
 import overlayElement from './overlay';
 import Localization from './localization';
 
+const l10nIdAttrName = 'data-l10n-id';
+const l10nArgsAttrName = 'data-l10n-args';
```
stasm: Use ALL_CAPS here?

zbraniecki:
@stasm I'd prefer to land this PR and work on the next one based on the next round of reviews in bugzilla. This helps me stay in sync while I'm discussing the next batch of changes.

Would it be ok with you to review this and get the next batch as a separate PR?

PR #64, opened 2017-08-21
Add AST.Placeable

stasm:
Implement projectfluent/fluent#52.


fluent-syntax/src/ftlstream.js
```diff
@@ -84,22 +87,6 @@ export class FTLParserStream extends ParserStream {
     return ((cc >= 48 && cc <= 57) || cc === 45); // 0-9
   }
 
-  isPeekNextLineIndented() {
```
stasm: This wasn't used anywhere.

fluent-syntax/src/ftlstream.js
```diff
@@ -208,7 +195,7 @@ export class FTLParserStream extends ParserStream {
   }
 
   skipToNextEntryStart() {
-    while (this.next()) {
+    while (this.ch) {
```
stasm: When the syntax error happens exactly on the newline character, `this.next()` would overshoot the beginning of the entry on the next line, if present.

fluent-syntax/src/index.js
```diff
@@ -15,14 +15,22 @@ export function serialize(resource, opts) {
...
 export function columnOffset(source, pos) {
-  const lastLineBreak = source.lastIndexOf('\n', pos);
-  return lastLineBreak === -1
-    ? pos
-    // Substracting two offsets gives length; substract 1 to get the offset.
-    : pos - lastLineBreak - 1;
+  // Find the last line break starting backwards from the index just before
+  // pos.  This allows us to correctly handle ths case where the character at
+  // pos  is a line break as well.
+  const fromIndex = pos - 1;
```
stasm: Without this, `columnOffset` was returning `0` when the character at `pos` was a new line.

fluent-syntax/src/parser.js
```diff
@@ -415,80 +415,83 @@ export default class FluentParser {
...
-      }
-      elements.push(text);
-    }
+    return new AST.TextElement(buffer);
+  }
 
-    return new AST.Pattern(elements);
+  getPlaceable(ps) {
+    ps.expectChar('{');
+    const expression = this.getExpression(ps);
```
stasm: `getExpression` should know how to handle nested placeables. I filed #66 for this.

fluent-syntax/src/parser.js
```diff
@@ -415,96 +415,106 @@ export default class FluentParser {
...
 
       ps.expectChar('\n');
       ps.expectChar(' ');
-      ps.skipLineWS();
+      ps.skipInlineWS();
 
       return new AST.SelectExpression(null, variants);
     }
 
+    ps.skipInlineWS();
```
stasm: I filed #67 to track this.

fluent-syntax/src/serializer.js
```diff
@@ -150,6 +152,20 @@ function serializeTextElement(text) {
 }
 
 
+function serializePlaceable(placeable) {
+  const expr = placeable.expression;
+
+  switch (expr.type) {
+    case 'Placeable':
+      return `{${serializePlaceable(expr)}}`;
```
stasm: I can't test this yet because the parser doesn't support nested Placeables.

fluent-syntax/src/parser.js
```diff
@@ -415,96 +415,106 @@ export default class FluentParser {
...
-          break;
-        }
 
+      // The end condition for getPattern's while loop is a newline
+      // which is not followed by a valid pattern continuation.
+      if (ch === '\n' && !ps.isPeekNextLinePattern()) {
+        break;
+      }
+
+      if (ch === '{') {
```
zbraniecki: I'd prefer to not enforce escaping if it's not necessary

fluent-syntax/src/util.js
```diff
@@ -0,0 +1,3 @@
+export function includes(arr, elem) {
+  return arr.indexOf(elem) > -1;
```
zbraniecki: could we get a compat-utils that are included only in the compat build? :) I think it would be nice to invest a bit of our time to design a build system for fluent which does not penalize evergreens for the sins of their ancestors. Although I agree that in this case the polyfill is small :)

fluent-syntax/src/parser.js
```diff
@@ -415,80 +415,83 @@ export default class FluentParser {
...
     const elements = [];
-    let firstLine = true;
+    ps.skipInlineWS();
 
-    if (this.withSpans) {
-      var spanStart = ps.getIndex();
+    // Special-case: trim leading whitespace and newlines.
+    if (ps.isPeekNextLinePattern()) {
+      ps.skipBlankLines();
+      ps.skipInlineWS();
```
zbraniecki: it would be nice if we could just peek to the first character that is non-WS and if it's a pattern start, skip to peek here, instead of resetting and skipping again, but since it's not a runtime parser, I don't think it's a blocker.

fluent-syntax/src/util.js
stasm: We could change `String.prototype` and `Array.prototype` and somehow do this only for the compat build. I don't think it's worth it though. Almost all of our needs are covered by the current pipeline. `includes` is a rare exception.

fluent-syntax/src/parser.js
stasm: Well, my question is then: do you think it's necessary? No syntax error in `key = { foo } }` is a bit weird to me.

fluent-syntax/src/parser.js
stasm: Yeah, I kept this to stay in line with other `isPeek...` methods which always reset the peek. We could try changing them such that they `skipToPeek` before returning `true`, although that would still need a bit of hand-holding in `isPeekNextLineVariantStart`.

fluent-syntax/src/util.js
zbraniecki: Nah, let's do what you did in your patch.

fluent-syntax/src/parser.js
zbraniecki: sorry, I should have state my opinion. I don't believe that it's necessary. I think it only looks weird because of the semantic meaning you give to `}` when seen in this context.
I'd argue that:

```
foo = Bar = Baz
```

looks equally weird. :)

stasm: OK, that's a good point, thanks!

PR #69, opened 2017-08-23
An example of a webcomponent localized using fluent-web

zbraniecki:


zbraniecki:
since it's fluent-web in its early days, I'm gonna merge and leave post-commit review on @stasm :)

PR #70, opened 2017-08-24
Add data-l10n-attrs for whitelisting attributes

zbraniecki:
This is necessary for web components to be localizable.

### DIFF
--- a/fluent-dom/src/overlay.js
+++ b/fluent-dom/src/overlay.js
@@ -71,14 +71,12 @@ export default function overlayElement(element, translation) {
     return;
   }
 
-  const whitelistedAttrs = element.hasAttribute('data-l10n-attrs') ?
-    element.getAttribute('data-l10n-attrs')
-      .split(',').map(attr => attr.trim()) :
-    null;
+  const explicitlyAllowed = element.hasAttribute('data-l10n-attrs')
+    ? element.getAttribute('data-l10n-attrs').split(',').map(i => i.trim())
+    : null;
 
   for (const [name, val] of translation.attrs) {
-    if (isAttrAllowed({ name }, element) ||
-        (whitelistedAttrs && whitelistedAttrs.includes(name))) {
+    if (isAttrAllowed({ name }, element, explicitlyAllowed)) {
       element.setAttribute(name, val);
     }
   }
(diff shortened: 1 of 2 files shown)


fluent-web/examples/webcomponent.html
```diff
@@ -71,6 +71,8 @@
       </div>
     </template>
 
-    <fluent-widget data-l10n-id="proceed-button"></fluent-widget>
+    <fluent-widget
+      data-l10n-id="proceed-button"
+      data-l10n-attrs="label, action-ok, action-cancel"></fluent-widget>
```
stasm: `class` uses space-separated lists of values, but the newer `srcset` and `sizes` both use comma-separated, so this is fine. (I recall being against commas at some point.)

fluent-dom/src/overlay.js
```diff
@@ -71,8 +71,14 @@ export default function overlayElement(element, translation) {
...
 
+  const whitelistedAttrs = element.hasAttribute('data-l10n-attrs') ?
+    element.getAttribute('data-l10n-attrs')
+      .split(',').map(attr => attr.trim()) :
+    null;
+
   for (const [name, val] of translation.attrs) {
-    if (isAttrAllowed({ name }, element)) {
+    if (isAttrAllowed({ name }, element) ||
+        (whitelistedAttrs && whitelistedAttrs.includes(name))) {
```
stasm: Is the fact that this logic is only here and not in `isAttrAllowed` on purpose? Please add a comment explaining why.

fluent-dom/src/overlay.js
```diff
@@ -71,8 +71,14 @@ export default function overlayElement(element, translation) {
     return;
   }
 
+  const whitelistedAttrs = element.hasAttribute('data-l10n-attrs') ?
+    element.getAttribute('data-l10n-attrs')
+      .split(',').map(attr => attr.trim()) :
+    null;
```
stasm: I'd prefer the following indentation:

    declaration = predicate
      ? consequent
      : alternate;

Or, since the alternate isn't very interesting (`null`) you could also use a `var` inside of an `if` statement.

fluent-dom/src/overlay.js
```diff
@@ -71,8 +71,14 @@ export default function overlayElement(element, translation) {
     return;
   }
 
+  const whitelistedAttrs = element.hasAttribute('data-l10n-attrs') ?
```
stasm: Perhaps call this `explicitlyAllowed`? Also, would `data-l10n-attributes` be better?

fluent-dom/src/overlay.js
zbraniecki: Wanted to avoid `getAttribute`, `split` and `map` on each attribute per element. Is that a good reason?

stasm: You can process the list once, but move the `includes` check into `isAttrAllowed`. **If** that's what you want. It sounds like it shouldn't be possible to add those whitelisted attributes to the element's children?

zbraniecki:
Updated to your feedback. Didn't enable explicitly allowed attrs in children yet, but can be added later without hassle.

PR #73, opened 2017-09-06
Bug 922577 - Do not leak attributes between translations

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=922577

### DIFF
--- a/fluent-dom/src/overlay.js
+++ b/fluent-dom/src/overlay.js
@@ -7,7 +7,7 @@ const reOverlay = /<|&#?\w+;/;
  *
  * Source: https://www.w3.org/TR/html5/text-level-semantics.html
  */
-const ALLOWED_ELEMENTS = {
+const LOCALIZABLE_ELEMENTS = {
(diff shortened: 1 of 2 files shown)


zbraniecki:
If I understand the patch correctly, you'd like to forbid localizable attributes to be placed in the source.
It seems like it could surprise a developer - maybe we should log a warning if that happens?

stasm:
> If I understand the patch correctly, you'd like to forbid localizable attributes to be placed in the source.

I think there are a few things we could do to improve the developer experience:
  - warn when a non-localizable element is present in source (it will be removed),
  - warn when a localizable attribute is removed from the source element,
  - warn when a localizable attribute us removed from a source element's child.

Do you think we should do all of them in this PR?

zbraniecki:
Maybe we should even start with throwing?

stasm:
I think throwing is only a viable option for the first use-case from my comment above. In the other two we could break the UI on retranslation if the new localization is missing an attribute.

PR #76, opened 2017-09-27
Parser syntax sparse bl

zbraniecki:
That's more or less how I think it'll have to work if we want to keep the spans and proper tokenizer flow.

I can't say I like it... :(

### DIFF
--- a/fluent-syntax/src/ftlstream.js
+++ b/fluent-syntax/src/ftlstream.js
@@ -33,7 +33,7 @@ export class FTLParserStream extends ParserStream {
 
   peekSkipBlankLines() {
     while (true) {
-      let lineStart = this.getPeekIndex();
+      const lineStart = this.getPeekIndex();
 
       this.peekInlineWS();
 
--- a/fluent-syntax/src/ftlstream.js
+++ b/fluent-syntax/src/ftlstream.js
@@ -31,6 +31,21 @@ export class FTLParserStream extends ParserStream {
     }
   }
 
+  peekSkipBlankLines() {
+    while (true) {
+      let lineStart = this.getPeekIndex();
+
+      this.peekInlineWS();
+
+      if (this.currentPeekIs('\n')) {
+        this.peek();
+      } else {
+        this.resetPeek(lineStart);
+        break;
+      }
+    }
+  }
+
   skipInlineWS() {
     while (this.ch) {
       if (!includes(INLINE_WS, this.ch)) {
@@ -94,6 +109,8 @@ export class FTLParserStream extends ParserStream {
 
     this.peek();
 
+    this.peekSkipBlankLines();
+
     const ptr = this.getPeekIndex();
 
     this.peekInlineWS();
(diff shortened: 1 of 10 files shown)


zbraniecki:
I think this is ready to be reviewed.

I analyzed the change in span ranges and the cause is that we now start the span for message value after we skip the inline whitespace. For example:

```
key = Value
     ^
     |---- start before
```


```
key = Value
      ^
      |---- start after
```

and:

```
key =
     ^
     |---- start before
    Value
```


```
key =
    Value
    ^
    |---- start after
```

zbraniecki:
I'm still testing performance but first indications are that it doesn't impact runtime perf, just the fluent-syntax parsing one, which is not significant for us. I tested on node, need to recompile spidermonkey to test against it.

zbraniecki:
Ok, got some more performance tests.

This patch does actually regress on jsshell:

```
parseFTL: 
  mean:   21812.23 (+24%)
  stdev:  2660.13
  sample: 30
parseFTLEntries: 
  mean:   3006.2 (+8%)
  stdev:  223.84
  sample: 30
format: 
  mean:   2752.9 (+76%)
  stdev:  452.41
  sample: 30
```

I have no idea why the `format` test gets such a penalty, since the AST produced with the patch is identical to the one produced on master. @stasm - any ideas?

zbraniecki:
Here's the same data for nodejs:

```
parseFTL: 
  mean:   8712.83 (+16%)
  stdev:  620.47
  sample: 30
parseFTLEntries: 
  mean:   1983.53 (+13%)
  stdev:  352.69
  sample: 30
format: 
  mean:   744.17 (+7.000000000000001%)
  stdev:  88.91
  sample: 30
```

So, still a regression but not as significant as on jsshell. I have no idea what's up with jsshell... maybe with the patch we cross some boundary due to, I don't know, number of variables? Lines of code?

fluent-syntax/src/ftlstream.js
```diff
@@ -31,6 +31,21 @@ export class FTLParserStream extends ParserStream {
     }
   }
 
+  peekSkipBlankLines() {
```
stasm: This name is confusing. Does it peek or does it skip?

fluent-syntax/src/parser.js
```diff
@@ -242,6 +243,7 @@ export default class FluentParser {
 
     while (true) {
       ps.expectChar('\n');
+      ps.skipBlankLines();
       ps.skipInlineWS();
```
stasm: This looks like a common pattern. Perhaps it's worth having a function called `expectIndent`?

stasm:
I filed [bug 1406880](https://bugzilla.mozilla.org/show_bug.cgi?id=1406880) because it looks like the spans of multiline Patterns are slightly different from the ones visualized in [your comment above](https://github.com/projectfluent/fluent.js/pull/76#issuecomment-333573836). I'm not yet sure which approach is better, let's decide in the bug and fix this if needed.

PR #80, opened 2017-10-17
Catch errors thrown by Intl formatters

stasm:
Intl formatters may throw errors when constructed (if unrecognized
options are passed) or in format() if the formatted value is not
supported.

This is a simple fix which simply catches those errors.  The real fix
should report them, for instance into env.errors, but it's not clear to
me right now how to do this right.


stasm:
Github is having hiccups and it doesn't seem to see the updates on my branch. Perhaps closing and re-opening will help?

stasm:
Github is having a hiccup and I can't update my branch for some reason. I'll open a new PR.

PR #87, opened 2017-10-30
Use babel-preset-env

galdiolo:
Use babel-preset-env instead of babel-preset-latest (deprecated)


stasm:
Thanks for the PR! I think you'll also need to change the Babel config here:

https://github.com/projectfluent/fluent.js/blob/cc32a1ec5608b2409132f2f0c0551106100d8cb5/compat_config.js#L8

I haven't used `babel-present-env` yet. Do we need to specify the list of supported browsers and their versions?

stasm:
I opened #99 to continue this and fix the failing tests.

PR #88, opened 2017-11-03
Construct the translation array with the length of the expected return value

zbraniecki:
Small change, but I think worth having - this makes `formatValues` and `formatMessages` return the array with the length matching the length of passed keys, even if no translation can be found.

### DIFF
--- a/fluent-dom/src/localization.js
+++ b/fluent-dom/src/localization.js
@@ -54,7 +54,7 @@ export default class Localization {
    * @private
    */
   async formatWithFallback(keys, method) {
-    const translations = [];
+    const translations = new Array(keys.length);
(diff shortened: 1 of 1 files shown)


stasm:
In which cases this isn't already the case? Is this a workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1410849 ?

zbraniecki:
If generateMessages returns no contexts, the `for await` loop is empty and we just return `translations`.

stasm:
In this case the `translations` array is sparse and full of `undefined` values. This breaks the contract of always returning the id.

I agree that we should fix this but I don't think this is the right fix.

zbraniecki:
fair enuf.

Issue #92, opened 2017-11-14
tests failing, also not all tests are running


hkasemir:
Following instructions to set up and test the fluent packages results in tests failing

also, this line is preventing some tests from running: https://github.com/projectfluent/fluent.js/blob/master/fluent/test/functions_builtin_test.js#L86

with that `.only` included, one test fails, by removing it, 6 tests fail.

zbraniecki:
The `.only` has been introduced in #81. With it, I see:

```
> make test
  59 passing (99ms)
  49 pending
```

when i remove it I see:

```
> make test
  196 passing (170ms)
  54 pending
```

@hkasemir - can you investigate which tests are failing and why? I suspect something is off with your node environment and its different than the one we have. It may be that you need Intl API, or it may be that we need to handle more variants of Node+Intl combinations, but I'd like to start with identifying tests and their failures.

stasm:
Sorry about the regression introduced in #81. I use `.only` quite often to isolate a single test during development. I found https://www.npmjs.com/package/eslint-plugin-no-only-tests which would prevent such errors in the future.

stasm:
There's also https://www.npmjs.com/package/eslint-plugin-mocha which has more configuration options.

stasm:
I opened #94 to add a linting rule forbidding `.only` tests.

I have the same output as @zbraniecki before and after fixing that rogue `.only` in `functions_builtin_test.js`. I tried on a fresh clone and I was able to set it up by following the instructions in the README. I'm using node 8.9.1.

hkasemir:
using node v6.11.2 and npm 5.4.1 fresh clone of the repo:
<img width="471" alt="image" src="https://user-images.githubusercontent.com/10947914/32868907-21ccdc26-ca29-11e7-92eb-5ba21e53242b.png">

using node v8.9.1,  seems all tests are passing 🤔 could be a good idea to define which version of node is required in the readme under instructions for local dev. Glad that was an easy fix. I did take the liberty of checking https://github.com/projectfluent/fluent.js/issues/93 to make sure it was not fixed by node version, but it is still an issue.

stasm:
I found out why the tests fail. Node 6.x doesn't support `Object.entries` which is used here: https://github.com/projectfluent/fluent.js/blob/da0a580ec8291fc43ab849b5dfed2c71febd7f85/fluent/src/builtins.js#L29

[...]

PR #95, opened 2017-11-16
Drop support for node versions before 8 LTS (8.9.0)

stasm:
As per #92, let's explicitly drop support for node versions prior to 8.9.0 which is the first release of the current LTS line (_Carbon_).

References:

  - https://github.com/nodejs/Release
  - https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md
  - https://medium.com/the-node-js-collection/news-node-js-8-moves-into-long-term-support-and-node-js-9-becomes-the-new-current-release-line-74cf754a10a0

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,10 +73,13 @@ feedback.
 Local Development
 -----------------
 
-Hacking on `fluent.js` is easy!  To quickly get started clone the repo:
+Hacking on `fluent.js` is easy! To quickly get started clone the repo:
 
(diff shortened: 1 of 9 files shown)


stasm:
@hkasemir r?

stasm:
> Only thought is you could also include a .nvmrc file and include nvm use in the instructions to ensure that devs also work with the same version.

I don't have much experience with `.nvmrc` personally but I'll be happy to take a new PR to add it to the repo.

Issue #89, opened 2017-11-07
[fluent-react] Localized nested elements as props is failing 


hkasemir:
I have the following (simplified, I know the link should also be localized, but first things first) code: 

```js
<Localized
  id='door-code-explanation'
  $contactLink={<a href='mailto:support@foo.com'>Contact us</a>}>
  <p>
    This code will be surfaced to people attending the upcoming session.
  </p>
</Localized>
```

```
// ftl
door-code-explanation = 
    This code will be surfaced to people attending the upcoming session.
    { $contactLink } if you have further questions.
```

what I would expect to see:
> This code will be surfaced to people attending the upcoming session. [Contact us](foo.com) if you have further questions.

instead I see
> This code will be surfaced to people attending the upcoming session. contactLink if you have further questions.

There are no errors in the console.
Also, when I change the `$contactLink` prop to a regular string, the result is as expected.

stasm:
Thanks for the report. This should work in `fluent-react` 0.4.x. I'll take a look. I see that there are no tests for this so perhaps I regressed something.

stasm:
It looks like this is Webpack's fault. At least in my setup which uses `create-react-app`.

When Webpack bundles all modules and their dependencies, it ends up duplicating the following imports (also when I `import from 'fluent'` rather than `import from 'fluent/compat'`). `FluentType` should `=== MessageArgument` and yet it doesn't.

https://github.com/projectfluent/fluent.js/blob/cce4d42997e33cb8d0960bd62a56f5db36ca190a/fluent/src/resolver.js#L50

https://github.com/projectfluent/fluent.js/blob/cce4d42997e33cb8d0960bd62a56f5db36ca190a/fluent-react/src/localized.js#L3

I think the solution is to do what React does with its elements: add a `$$typeof: Symbol.for('FluentType')` field to all `FluentTypes` so that it's easy to discover them without relying on `instanceof`.

That said, my mid-term plan is to remove `MessageContext.formatToParts` and remove the feature of passing React elements as arguments to translations and replace it with a more declarative approach, similar to DOM overlays in `fluent-web`.

hkasemir:
OK, I know that for us it is important to be able to include components or DOM elements as jsx. I don't know what you mean by DOM overlays - do you have examples or documentation?

stasm:
I just published `fluent` 0.4.2 which fixed this.

Issue #90, opened 2017-11-08
[fluent-react] DATETIME formatting always uses default


hkasemir:
fluent version 0.4.1
```
// ftl file
today-is = today is { DATETIME($date, month: "long", day: "numeric") }
```

```
// React component
...
  <Localized id='today-is' $date={new Date()}>
    <p>today is {(new Intl.DateTimeFormat("fr", {month: "long", day: "numeric"})).format()}</p>
  </Localized>
  <p>today is {(new Intl.DateTimeFormat("fr", {month: "long", day: "numeric"})).format()}</p>
...
```

result
<img width="163" alt="image" src="https://user-images.githubusercontent.com/10947914/32575286-6c811a58-c488-11e7-83bf-afdaac6dff1b.png">

node version 6.11.2

zbraniecki:
Seems like an issue with fluent-react, since I can't reproduce it in fluent-web.

@stasm ?

stasm:
I can't reproduce this. In https://github.com/projectfluent/fluent.js/commit/cce4d42997e33cb8d0960bd62a56f5db36ca190a I added a `DATETIME($date, month: "long", day: "numeric")` placeable to `fluent-react`'s `hello-world` example. I can see the full name of the month spelled out correctly.

hkasemir:
I can see that this is not reproduced by the examples, however, I was able to create a very basic test repo that repro's all the issues: https://github.com/projectfluent/fluent.js/issues/91, https://github.com/projectfluent/fluent.js/issues/90, and https://github.com/projectfluent/fluent.js/issues/89

https://github.com/hkasemir/test-fluent
it's using node 8.9.1, and requires `npm link` to the local fluent-react

hkasemir:
however I did just discover that if I also `npm link fluent`, issue #90 is solved.

zbraniecki:
I was able to reproduce all three issues using Heidi's minimized testcase. In my case, `npm link fluent` did not fix issue #90.

stasm:
Thanks for putting together the testcase, @hkasemir. I can confirm that `npm link fluent` fixes #90. This is thanks to #81 which landed on master but hasn't made it into a released version yet.

Linking both `fluent` and `fluent-react` also fixes #89 for me. I suspect this is because when they're both linked, Webpack bundles them together and the `instanceof FluentType` actually works.

stasm:
[...]

PR #98, opened 2017-11-21
Use a Symbol for detecting instances of FluentType

stasm:
Some bundlers make `instanceof` unreliable. Use a static `FluentType.isTypeOf` method to detect instances of `FluentType` via a `$$typeof` field and a `FluentType` symbol.

### DIFF
--- a/fluent/src/resolver.js
+++ b/fluent/src/resolver.js
@@ -382,7 +382,8 @@ function ExternalArgument(env, {name}) {
 
   const arg = args[name];
 
-  if (arg instanceof FluentType) {
+  // Return early if the argument already is an instance of FluentType.
+  if (FluentType.isTypeOf(arg)) {
     return arg;
   }
 
--- a/fluent/src/types.js
+++ b/fluent/src/types.js
@@ -40,6 +40,39 @@ export class FluentType {
   valueOf() {
     throw new Error('Subclasses of FluentType must implement valueOf.');
   }
+
+  /**
+   * Internal field used for detecting instances of FluentType.
+   *
+   * @private
+   */
+  get $$typeof() {
+    return Symbol.for('FluentType');
+  }
+
+  /**
+   * Check if a value is an instance of FluentType.
+   *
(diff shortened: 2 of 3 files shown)


stasm:
I still need to write a test or two. @hkasemir, do you want to see if this fixes #89 for you? To apply this locally, `cd` into `fluent-react` and `npm link fluent` to make sure it uses the local source rather than the one in `fluent-react/node_modules/fluent`.

hkasemir:
unfortunately this did not work locally for me. Adding a couple logs to try to see what was going wrong, I found that these logs:
```js
    console.log(obj)
    console.log(obj instanceof Object)
    console.log(obj.$$typeof)
    console.log(Symbol.for('FluentType'))
```
logged the following:
```
ElementArgument {value: {…}, opts: undefined}
compat.js:1246 true
compat.js:1247 undefined
compat.js:1248 Symbol(FluentType)
```

so it seems like `$$typeof` is not running properly for me, as it is returning `undefined` and logs inside the getter function don't log anything. I verified that I am using node 8.9.1 - and linked local to my [test-fluent](https://github.com/hkasemir/test-fluent) repo for checking.

hkasemir:
here's a screenshot of the expanded `ElementArgument`
image

stasm:
I updated the PR with a fix to `$$typeof` being `undefined`. I also amended the existing tests to verify the value of `$$typeof` on custom arguments. This is now ready for review.

Pike:
I'm wondering, should we still do this?

Reading https://github.com/projectfluent/fluent.js/issues/90#issuecomment-347162277, it seems that the issue this tries to fix is coming from bundling/installing (potentialy) incompatible versions?

stasm:
I don't think this is related to different versions, but rather the fact that some code might use the ES2015+ version of `fluent` and other code might use `fluent/compat`, the version transpiled to ES5. Or the fact that sometimes we're using `fluent` from `node_modules` and sometimes it's `fluent` linked locally. I don't feel like learning how Webpack works exactly and where it looks for modules so this fix seems like a reasonable way forward to me.

PR #99, opened 2017-11-21
Use babel-preset-env and define the browser support for compat builds

stasm:
This is a continuation of #87. `babel-preset-env` uses https://github.com/ai/browserslist to parse version ranges definitions. I specified a rather wide range here which is based on what `create-react-app` supports.

### DIFF
--- a/compat_config.js
+++ b/compat_config.js
@@ -5,13 +5,21 @@ export default {
     babel({
       'babelrc': false,
       'presets': [
-        ['latest', {
-          'es2015': {
-            'modules': false
+        ['env', {
+          // Cf. https://github.com/rollup/rollup-plugin-babel#modules
+          'modules': false,
+          'targets': {
+            'browsers': [
+              '>1%',
+              'last 4 versions',
+              'Firefox ESR',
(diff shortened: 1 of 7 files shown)


stasm:
Here's the browserlist query I propose we use for compat builds:

    >1%, last 4 versions, Firefox ESR, not ie < 9

Pike:
I tried to see what that query evaluates to, and failed. http://browserl.ist/ at least doesn't seem to take the usage stats into account.

One thing that concerns me a bit is the question of getting reproducable builds. Seems that the list of browsers we support in a particular version of our code can change from day to day?

stasm:
It [works for me](http://browserl.ist/?q=%3E1%25%2C+last+4+versions%2C+Firefox+ESR%2C+not+ie+%3C+9). It gives me a global coverage of 91.63%.

And yes, with time, we will end up supporting fewer and fewer legacy browsers. It's an exciting prospect. If reproducible builds are a hard requirement for someone, I'd encourage them to use the vanilla Fluent code and transpile themselves.

PR #101, opened 2017-12-09
DOM overlays for fluent-react

### LINKED ISSUE #104: Remove `MessageFormat.formatToParts`
The main motivation behind `formatToParts` was the use-case of passing React elements into translations in `fluent-react`. With a plan to implement overlays in `fluent-react` (#103) this use-case goes away.

`formatToParts` has been buggy and under-spec'ed. Let's remove it for good. It will also make the interface of `MessageContext` very simple: the `format` method always returns a `null` or a `string`.

stasm:
This is the first part of #103. It also fixes #104.


stasm:
This is now ready for reviews. The to-do list at the bottom of #103 has more information about my next steps. For now, I'd like to land this minimal implementation without doing any sanitization of attributes or elements.

There are four commit in this PR. The first one is a small change to how fluent-react's tests work which made it easier to write tests for the second commit. The second commit is where the DOM overlays feature is introduced. The third commit removes `MessageContext.formatToParts` whose only use-case was `fluent-react`. The fourth renames `FluentType.valueOf` to `toString` because now it really always returns a string. It was called `valueOf` because it could return other types, like React elements.

@hkasemir I'd love to get your input on this, especially on the second commit.

fluent-react/test/localized_overlay_test.js
```diff
@@ -0,0 +1,195 @@
...
+
+  test('one element is matched', function() {
+    const mcx = new MessageContext();
+    const l10n = new ReactLocalization([mcx]);
+
+    mcx.addMessages(`
+foo = Click <button>me</button>!
+`)
+
+    const wrapper = shallow(
```
hkasemir: it's cool to see how you set this up for testing!

fluent-react/src/localized.js
```diff
@@ -116,12 +100,36 @@ export default class Localized extends Component {
     }
 
     const msg = mcx.getMessage(id);
-    const args = toArguments(this.props);
-    const { parts, attrs } = l10n.formatCompound(mcx, msg, args);
+    const [args, elems] = toArguments(this.props);
+    const { value, attrs } = l10n.formatCompound(mcx, msg, args);
+
+    if (value === null || !value.includes('<')) {
```
hkasemir: I suppose there are no unexpected side-effects if the string inlcudes the `<` character outside of the context of a markup element. For example `The result is that 0 < 3 ...`.

It looks like the `parseMarkup` that comes next will just leverage the DOM API to split out the different parts, and so this step is more of an optimization in case the translation string does not contain react components. Just thinking out loud and checking my understanding - does that sound about right?

fluent-react/src/localized.js
```diff
@@ -1,53 +1,37 @@
...
- * With `ElementArgument` it becomes possible to pass React elements as
- * arguments to translations.  This may be useful for passing links or buttons,
- * or in general: elements with logic which should be defined in the app.
- */
-class ElementArgument extends MessageArgument {
-  valueOf() {
-    return this.value;
-  }
-}
+const RESERVED_PROPS = [
```
zbraniecki: can you document where this list comes from and what are reserved props?

fluent-react/src/localized.js
```diff
@@ -116,12 +100,36 @@ export default class Localized extends Component {
...
+        return childNode.textContent;
+      }
+
+      return cloneElement(
+        elems[childNode.localName],
+        // XXX Explcitly ignore any attributes defined in the translation.
+        null,
+        // Void elements (like input) must not have any children. If
+        // childNode.textContent is an empty string we need to pass null
+        // instead.
```
zbraniecki: I struggle to understand the comment. 
Why are you testing textContent in the context of void elements, and if it's an empty string you're returning null instead?
May be an indication that other readers will struggle with it as well.

fluent-react/examples/html-markup/src/App.js
```diff
@@ -0,0 +1,29 @@
...
+  function showAlert(id) {
+    const { getString } = props;
+    alert(getString(id));
+  }
+
+  return (
+    <div>
+      <Localized id="sign-in-or-cancel"
+        button={<button onClick={() => showAlert('clicked-sign-in')}></button>}
+        a={<button className="text" onClick={() => showAlert('clicked-cancel')}></button>}
```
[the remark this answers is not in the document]
stasm: Yes, the result will be two buttons.

I went for names that are familiar if you know HTML. Otherwise we'd need something like `<button1>` and `<button2>`, or `<signin>` and `<cancel>`. I'm slightly worried about using non-standard HTML names in the translations because the are more likely to be translated by the localizers.

Let's also keep in mind that it's rare to have two elements of the same type.

fluent-react/examples/html-markup/src/l10n.js
```diff
@@ -0,0 +1,37 @@
...
+
+agree-prompt = Nazywam się <input/> i <button>zgadzam się</button>.
+agree-alert = Fajnie, zgadzamy się.
+  `,
+  'en-US': `
+sign-in-or-cancel = <button>Sign in</button> or <a>cancel</a>.
+clicked-sign-in = You are now signed in.
+clicked-cancel = OK, nevermind.
+
+agree-prompt = My name is <input/> and <button>I agree</button>.
```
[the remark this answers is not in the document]
stasm: Great question. It's out of scope of this PR. I put it on the to-do list in #103. We'll need to build a list of known elements and their localizable attributes and we'll also need a way for developers to define which props are localizable on custom component.

fluent-react/src/localized.js
stasm: Correct. In the rare scenario of a false-positive (like in the example you gave) we'll pay a bit extra to parse HTML which is not there. It will parse as a text node and things should work just fine. I'll add a test.

fluent-react/test/localized_overlay_test.js
```diff
@@ -0,0 +1,195 @@
...
+    const l10n = new ReactLocalization([mcx]);
+
+    mcx.addMessages(`
+foo = <confirm>Sign in</confirm> or <cancel>cancel</cancel>.
+`)
+
+    const wrapper = shallow(
+      <Localized id="foo"
+        confirm={<button className="confirm"></button>}
+        cancel={<button className="cancel"></button>}
```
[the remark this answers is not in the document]
stasm: I like this because it's very semantic, but I'm slightly afraid that localizers also translate the names of elements, like so:

    <confirm>Sign in</confirm> or <cancel>cancel</cancel>.
    <potwierdź>Zaloguj się</potwierdź> albo <anuluj>anuluj</anuluj>.

For this reason I think it's safest to use known HTML names, like `button` or `a`.

PR #102, opened 2017-12-09
Sync gecko changes with mozilla-central

zbraniecki:
Pretty trivial updates to fluent-gecko and a couple simple changes that spill to other dependencies.

I had to update rollup to handle more of the syntax we use currently (due to new acron). Guess it's not a bad thing to stay up to date with rollup :)

### DIFF
--- a/fluent-gecko/makefile
+++ b/fluent-gecko/makefile
@@ -7,25 +7,25 @@ build: MessageContext.jsm Localization.jsm DOMLocalization.jsm l10n.js
 MessageContext.jsm: $(SOURCES)
 	@rollup $(CURDIR)/src/message_context.js \
 	    --config ./xpcom_config.js \
-	    --output ./dist/$@
+	    --output.file ./dist/$@
 	@echo -e " $(OK) $@ built"
 
 Localization.jsm: $(SOURCES)
 	@rollup $(CURDIR)/src/localization.js \
 	    --config ./xpcom_config.js \
-	    --output ./dist/$@
+	    --output.file ./dist/$@
 	@echo -e " $(OK) $@ built"
 
 DOMLocalization.jsm: $(SOURCES)
 	@rollup $(CURDIR)/src/dom_localization.js \
 	    --config ./xpcom_dom_config.js \
-	    --output ./dist/$@
+	    --output.file ./dist/$@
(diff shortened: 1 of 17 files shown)


fluent-gecko/makefile
```diff
@@ -7,25 +7,25 @@ build: MessageContext.jsm Localization.jsm DOMLocalization.jsm l10n.js
...
 	    --config ./xpcom_dom_config.js \
-	    --output ./dist/$@
+	    --output.file ./dist/$@
 	@echo -e " $(OK) $@ built"
 
 l10n.js: $(SOURCES)
 	@rollup $(CURDIR)/src/l10n.js \
-	    --config ./xpcom_config.js \
-	    --output ./dist/$@
+		  --output.format es \
```
stasm: Nit: please fix the indentation here.

fluent-gecko/src/dom_localization.js
```diff
@@ -1,4 +1,24 @@
...
+ * In the future, we may want to allow certain modules to override this
+ * with a different negotitation strategy to allow for the module to
+ * be localized into a different language - for example DevTools.
+ */
+function defaultGenerateMessages(resourceIds) {
+  return L10nRegistry.generateContexts(undefined, resourceIds);
+}
+
+class GeckoDOMLocalization extends DOMLocalization {
+  constructor(resourceIds, generateMessages = defaultGenerateMessages) {
```
stasm: Is there a reason to allow passing custom `generateMessages` to the constructor of `GeckoDOMLocalization`? It would appear that it will always want to use the L10nRegistry.

If it's for testing, then perhaps we should stub `L10nRegistry.generateContexts` rather than increase the API surface?

zbraniecki: Sure, done.

zbraniecki: ugh, actually, after trying it on tests, it's pretty quirky to do. I think I'd prefer to expose the ability to override `generateMessages` on GeckoLocalization/GeckoDOMLocalization.

zbraniecki:
Filed bug https://bugzilla.mozilla.org/show_bug.cgi?id=1424810 for the constructor refactor.

PR #109, opened 2017-12-13
Add 0.5 comments support

zbraniecki:
Implements a subset of https://github.com/projectfluent/fluent/issues/58 required for limited Firefox x-channel compatibility across beta and stable.

### DIFF
--- a/fluent-syntax/src/ast.js
+++ b/fluent-syntax/src/ast.js
@@ -19,11 +19,10 @@ class SyntaxNode extends BaseNode {
 }
 
 export class Resource extends SyntaxNode {
-  constructor(body = [], comment = null) {
+  constructor(body = []) {
     super();
     this.type = 'Resource';
     this.body = body;
-    this.comment = comment;
   }
 }
 
@@ -192,20 +191,31 @@ export class Symbol extends Identifier {
   }
 }
 
-export class Comment extends Entry {
+export class BaseComment extends Entry {
   constructor(content) {
     super();
-    this.type = 'Comment';
+    this.type = 'BaseComment';
     this.content = content;
   }
 }
 
-export class Section extends Entry {
-  constructor(name, comment = null) {
-    super();
-    this.type = 'Section';
-    this.name = name;
-    this.comment = comment;
+export class Comment extends BaseComment {
+  constructor(content) {
+    super(content);
+    this.type = 'Comment';
+  }
+}
+
+export class GroupComment extends BaseComment {
+  constructor(content) {
+    super(content);
+    this.type = 'GroupComment';
+  }
+}
+export class ResourceComment extends BaseComment {
+  constructor(content) {
+    super(content);
+    this.type = 'ResourceComment';
   }
 }
 
--- a/fluent-syntax/src/ftlstream.js
+++ b/fluent-syntax/src/ftlstream.js
@@ -109,7 +109,7 @@ export class FTLParserStream extends ParserStream {
     return ((cc >= 48 && cc <= 57) || cc === 45); // 0-9
   }
 
-  isPeekNextLineComment() {
+  isPeekNextLineZeroFourStyleComment() {
     if (!this.currentPeekIs('\n')) {
       return false;
     }
@@ -122,6 +122,39 @@ export class FTLParserStream extends ParserStream {
         this.resetPeek();
         return true;
       }
+    }
+
+    this.resetPeek();
+    return false;
+  }
+
+  // -1 - any
+  //  0 - comment
+  //  1 - group comment
+  //  2 - resource comment
+  isPeekNextLineComment(level = -1) {
+    if (!this.currentPeekIs('\n')) {
+      return false;
+    }
+
+    let i = 0;
+
+    while (i <= level || (level === -1 && i < 3)) {
+      this.peek();
+      if (!this.currentPeekIs('#')) {
+        if (i !== level && level !== -1) {
+          this.resetPeek();
+          return false;
+        }
+        break;
+      }
+      i++;
+    }
+
+    this.peek();
+    if ([' ', '\n'].includes(this.currentPeek())) {
+      this.resetPeek();
+      return true;
     }
 
     this.resetPeek();
--- a/fluent-syntax/src/parser.js
+++ b/fluent-syntax/src/parser.js
@@ -21,7 +21,7 @@ function withSpan(fn) {
     }
 
     // Spans of Messages and Sections should include the attached Comment.
-    if (node.type === 'Message' || node.type === 'Section') {
+    if (node.type === 'Message') {
       if (node.comment !== null) {
         start = node.comment.span.start;
       }
@@ -42,7 +42,7 @@ export default class FluentParser {
 
     // Poor man's decorators.
     [
-      'getComment', 'getSection', 'getMessage', 'getAttribute',
+      'getComment', 'getMessage', 'getAttribute',
       'getIdentifier', 'getVariant', 'getSymbol', 'getNumber', 'getPattern',
       'getTextElement', 'getPlaceable', 'getExpression',
       'getSelectorExpression', 'getCallArg', 'getString', 'getLiteral',
@@ -52,8 +52,6 @@ export default class FluentParser {
   }
 
   parse(source) {
-    let comment = null;
-
     const ps = new FTLParserStream(source);
     ps.skipBlankLines();
 
@@ -62,8 +60,16 @@ export default class FluentParser {
     while (ps.current()) {
       const entry = this.getEntryOrJunk(ps);
 
-      if (entry.type === 'Comment' && entries.length === 0) {
-        comment = entry;
+      if (entry === null) {
+        // That happens when we get a 0.4 style section
+        continue;
+      }
+
+      if (entry.type === 'Comment' &&
+        entry.zeroFourStyle && entries.length === 0) {
+        const comment = new AST.ResourceComment(entry.content);
+        comment.span = entry.span;
+        entries.push(comment);
       } else {
         entries.push(entry);
       }
@@ -72,7 +78,7 @@ export default class FluentParser {
       ps.skipBlankLines();
     }
 
-    const res = new AST.Resource(entries, comment);
+    const res = new AST.Resource(entries);
 
     if (this.withSpans) {
       res.addSpan(0, ps.getIndex());
@@ -117,7 +123,7 @@ export default class FluentParser {
   getEntry(ps) {
     let comment;
 
-    if (ps.currentIs('/')) {
+    if (ps.currentIs('/') || ps.currentIs('#')) {
       comment = this.getComment(ps);
 
       // The Comment content doesn't include the trailing newline. Consume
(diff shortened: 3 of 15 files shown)


fluent-syntax/src/ast.js
```diff
@@ -194,27 +193,49 @@ export class Identifier extends SyntaxNode {
...
+}
+export class ResourceComment extends BaseComment {
+  constructor(content) {
+    super();
+    this.type = 'ResourceComment';
+    this.content = content;
+  }
+}
+
 export class Section extends Entry {
```
stasm: We should remove `Section` from `ast.js`.

fluent-syntax/src/ftlstream.js
```diff
@@ -109,14 +109,50 @@ export class FTLParserStream extends ParserStream {
     return ((cc >= 48 && cc <= 57) || cc === 45); // 0-9
   }
 
-  isPeekNextLineComment() {
+  // -1 - any
+  //  0 - comment
+  //  1 - group comment
+  //  2 - resource comment
+  isPeekNextLineComment(level = -1, oldStyle) {
```
stasm: I know this is a temporary, but how about we rename `oldStyle` to `zeroFourSyntax` to make it explicit?

fluent-syntax/src/ast.js
```diff
@@ -194,27 +193,49 @@ export class Identifier extends SyntaxNode {
...
   constructor(content) {
     super();
     this.type = 'Comment';
     this.content = content;
   }
 }
 
+export class Comment extends BaseComment {
+  constructor(content) {
+    super();
```
stasm: You could pass `content` to `super` and avoid setting it in again in the constructor here.

fluent-syntax/src/ftlstream.js
```diff
@@ -109,14 +109,50 @@ export class FTLParserStream extends ParserStream {
...
 
-    if (this.currentPeekIs('/')) {
+    if (oldStyle === false && this.currentPeekIs('#')) {
+      this.peek();
+      if (level === -1 &&
+        (this.currentPeekIs(' ') || this.currentPeekIs('#'))) {
+        this.resetPeek();
+        return true;
+      }
+      if (level === 0 && this.currentPeekIs(' ')) {
```
stasm: It looks like this could use a `while` loop where you decrement `level` in each pass.

fluent-syntax/src/ast.js
```diff
@@ -194,27 +193,49 @@ export class Identifier extends SyntaxNode {
...
-    this.type = 'Symbol';
+    this.type = 'VariantName';
   }
 }
 
-export class Comment extends Entry {
+export class BaseComment extends Entry {
   constructor(content) {
     super();
     this.type = 'Comment';
```
stasm: `BaseComment`

fluent-syntax/src/parser.js
```diff
@@ -153,20 +173,40 @@ export default class FluentParser {
...
         content += '\n';
         ps.next();
-        ps.expectChar('/');
-        ps.expectChar('/');
-        ps.takeCharIf(' ');
+        if (ps.currentIs('#')) {
+          ps.next();
+          if (ps.currentIs('#')) {
+            ps.next();
+            if (ps.currentIs('#')) {
```
stasm: Perhaps put this in a `for` loop iterating from 1 to `level`?

fluent-syntax/src/parser.js
```diff
@@ -117,7 +115,8 @@ export default class FluentParser {
   getEntry(ps) {
     let comment;
 
-    if (ps.currentIs('/')) {
+    if (ps.currentIs('/') ||
+      (ps.currentIs('#') && (ps.peekCharIs('#') || ps.peekCharIs(' ')))) {
```
stasm: Why does this peek one char ahead? Isn't a single `#` enough to branch into `getComment`?

fluent-syntax/src/parser.js
```diff
@@ -179,7 +226,7 @@ export default class FluentParser {
     ps.expectChar(']');
     ps.expectChar(']');
 
-    return new AST.Section(symb, comment);
+    return new AST.Section(symb);
```
stasm: `getSection` should return a `GroupComment` with the section's comment or `null` (or `undefined`) if the comment is missing.

fluent-syntax/test/fixtures_structure/resource_comment.json
```diff
@@ -1,16 +1,12 @@
...
-    "content": "This is a resource wide comment\nIt's multiline",
-    "span": {
-      "type": "Span",
-      "start": 0,
-      "end": 52
+  "body": [
+    {
+      "type": "ResourceComment",
+      "annotations": [],
+      "content": "This is a resource wide comment\nIt's multiline"
```
stasm: Why is there no `span` here?

fluent-syntax/src/parser.js
zbraniecki: Unfortunately, not.
```
key = Value
#word
```

should not branch into getComment since the comment starting code is `"# "|"## "|"### "`.

stasm: I feel like that should throw a `SyntaxError` inside of `getComment`.

fluent-syntax/src/ftlstream.js
```diff
@@ -128,6 +128,39 @@ export class FTLParserStream extends ParserStream {
...
+  // -1 - any
+  //  0 - comment
+  //  1 - group comment
+  //  2 - resource comment
+  isPeekNextLineComment(level = -1) {
+    if (!this.currentPeekIs('\n')) {
+      return false;
+    }
+
+    let ptr = 0;
```
stasm: In other parts of this file, `ptr` is used to store a peek index. Perhaps rename this simply to `i`?

fluent-syntax/src/parser.js
```diff
@@ -126,21 +134,22 @@ export default class FluentParser {
...
+    if (ps.isIDStart() && (!comment || comment.type === 'Comment')) {
       return this.getMessage(ps, comment);
     }
 
     if (comment) {
       return comment;
     }
+
+    if (ps.currentIs('[')) {
+      return this.getSection(ps);
```
stasm: Shouldn't we skip sections and return a `GroupComment` if `comment` is present?

fluent-syntax/src/parser.js
```diff
@@ -163,23 +172,79 @@ export default class FluentParser {
...
+        Comment = AST.GroupComment;
+        break;
+      case 2:
+        Comment = AST.ResourceComment;
+        break;
+    }
+    return new Comment(content);
+  }
+
+  getSection(ps) {
```
stasm: Perhaps rename to `skipSection` and don't return anything for it?

fluent-syntax/test/fixtures_behavior/section_starts_with_one_bracket.ftl
```diff
@@ -1,2 +0,0 @@
-[ This is a broken section ]]
-//~ ERROR E0003, pos 1, args "["
```
stasm: This should still throw a syntax error, no?

fluent-syntax/test/fixtures_behavior/section_with_no_nl_after_it.ftl
```diff
@@ -1 +0,0 @@
-[[ This is a correct section ]]
```
stasm: It would be good to still test that this parses alright, even if it parses to an empty AST at this point.

fluent-syntax/src/serializer.js
```diff
@@ -40,10 +40,12 @@ export default class FluentSerializer {
         return serializeMessage(entry);
       case 'Section':
         return serializeSection(entry);
-      case 'Comment': {
-        const comment = serializeComment(entry);
-        return `\n${comment}\n\n`;
-      }
+      case 'Comment':
+        return `\n${serializeComment(entry)}\n\n`;
```
stasm: I think we should have a flag on the `FluentSerializer` class which starts as `isFirstEntry = true` and only add those leading `\n` if it's `false`.

fluent-syntax/src/parser.js
```diff
@@ -126,21 +134,26 @@ export default class FluentParser {
...
     }
 
-    if (ps.isIDStart()) {
-      return this.getMessage(ps, comment);
+    if (ps.currentIs('[')) {
+      this.skipSection(ps);
+      if (comment) {
+        return new AST.GroupComment(comment.content);
+      }
+      return null;
```
stasm: I might be missing something, but if you bring back the previous order of `ifs`, so first test for `[` and then test for identifier start, you could get away without this `return null` here and just let the code flow to the next `if`.

You could then remove the `node === null` check in `withSpan` and `parse`.

fluent-syntax/src/serializer.js
```diff
@@ -23,11 +24,13 @@ export default class FluentSerializer {
       parts.push(
         `${serializeComment(resource.comment)}\n\n`
       );
+      this.hasEntries = true;
     }
 
     for (const entry of resource.body) {
       if (entry.types !== 'Junk' || this.withJunk) {
         parts.push(this.serializeEntry(entry));
+        this.hasEntries = true;
```
stasm: I would avoid setting it in every pass of the loop and only do it if `hasEntries` is `false`.

fluent-syntax/src/serializer.js
```diff
@@ -40,10 +43,21 @@ export default class FluentSerializer {
...
+        }
+        return `${serializeComment(entry)}\n\n`;
+      case 'GroupComment':
+        if (this.hasEntries) {
+          return `\n${serializeGroupComment(entry)}\n\n`;
+        }
+        return `\n${serializeGroupComment(entry)}\n\n`;
+      case 'ResourceComment':
+        if (this.hasEntries) {
+          return `${serializeResourceComment(entry)}\n\n`;
```
stasm: This should start with `\n`.

fluent-syntax/src/parser.js
```diff
@@ -62,8 +65,13 @@ export default class FluentParser {
...
       const entry = this.getEntryOrJunk(ps);
 
-      if (entry.type === 'Comment' && entries.length === 0) {
-        comment = entry;
+      if (entry === null) {
+        // This is probably a section returning null.
+        continue;
+      }
+
+      if (entry.type === 'Comment' && entry.oldStyle && entries.length === 0) {
```
stasm: I don't see `oldStyle` anywhere else in the PR.

fluent-syntax/src/parser.js
zbraniecki: >  and just let the code flow to the next if.

Then it'll return regular `Comment` for old style comment that was attached to a section, instead of GroupComment.

> You could then remove the node === null check in withSpan and parse.

That's crashes if getEntry returns `null` (it does when section doesn't have a comment in 0.4 style).

stasm: I mean something like this:

```js
  getEntry(ps) {
    let comment;

    if (ps.currentIs('/') || ps.currentIs('#')) {
      comment = this.getComment(ps);

      // The Comment content doesn't include the trailing newline. Consume
      // this newline here to be ready for the next entry.  undefined stands
      // for EOF.
      ps.expectChar(ps.current() ? '\n' : undefined);
    }

    if (ps.currentIs('[')) {
      this.skipSection(ps);
      if (comment) {
        return new AST.GroupComment(comment.content);
      }
    }

    if (ps.isIDStart() && (!comment || comment.type === 'Comment')) {
      return this.getMessage(ps, comment);
    }

    if (comment) {
      return comment;
    }

    throw new ParseError('E0002');
  }
```

Then you never return `null`.

zbraniecki: I do in case the file is:

```
[[ This is a section ]]
```

which is one of our tests :(

What do you prefer here?

stasm: Ah, good point, thanks. Let's keep the tests.

Can we at least remove the `node === null` check in `withSpans`? I think we only return `null` from `getEntry` which isn't decorated with `withSpans` because it handles spans in an own custom way.

PR #110, opened 2017-12-13
Remove tags

zbraniecki:
Implements https://github.com/projectfluent/fluent/issues/67


fluent-syntax/src/parser.js
```diff
@@ -203,18 +202,11 @@ export default class FluentParser {
       attrs = this.getAttributes(ps);
     }
 
-    if (ps.isPeekNextLineTagStart()) {
-      if (attrs !== undefined) {
-        throw new ParseError('E0012');
```
stasm: Please remove `E0012` from `errors.js`.

fluent/test/tags_test.js
```diff
@@ -1,82 +0,0 @@
-'use strict';
-
-import assert from 'assert';
-
-import { MessageContext } from '../src/context';
```
stasm: Would you like me to land a follow-up which removes tags from `MessageContext`'s resolver?

PR #113, opened 2017-12-17
Coalesce all translations from MutationObserver into a single rAF

zbraniecki:
This changes the way we react to mutations by batching all localizations to happen once per animation frame irrelevant of the number of microtasked mutations fired.

### DIFF
--- a/fluent-dom/src/dom_localization.js
+++ b/fluent-dom/src/dom_localization.js
@@ -27,6 +27,11 @@ export default class DOMLocalization extends Localization {
 
     // A Set of DOM trees observed by the `MutationObserver`.
     this.roots = new Set();
+    // requestAnimationFrame handler.
+    this.pendingrAF = null;
+    // list of elements pending for translation.
+    this.pendingElements = new Set();
+    this.windowElement = windowElement;
     this.mutationObserver = new windowElement.MutationObserver(
       mutations => this.translateMutations(mutations)
     );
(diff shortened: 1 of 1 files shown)


stasm:
I need some context here :) What is this for? Does it make things faster? By how much? How did you test? Do you want me to prioritize this over 0.5?

zbraniecki:
> I need some context here :) 

Oh, stupid me. Sorry! Here it is: https://bugzilla.mozilla.org/show_bug.cgi?id=1389384

> What is this for?

Gecko and Fluent-DOM

> Does it make things faster?

It makes us do less things.

> How did you test?

I created a set of tests that synchronously and asynchronously updated UI in Preferences of Gecko and monitored the number of mutations and translations we perform.

With the patch we perform one batch per animation frame at most. Without it, we are happy to perform hundreds per millisecond if needed.

> Do you want me to prioritize this over 0.5?

I'd hope you can evaluate the code. The DOM API integration review is handled by :mossop in the bug. If you think you're to busy with other things I honestly believe that his review should be enough.

zbraniecki:
I got r+ from :mossop in https://bugzilla.mozilla.org/show_bug.cgi?id=1389384

@stasm do you want me to wait for you before landing this?

stasm:
No, please don't wait; go ahead and land.

stasm:
@zbraniecki How hard would it be to have tests for this?

zbraniecki:
I have no idea how to write tests that trigger the behavior I'm testing this on. Karma? I need to trigger tons of mutations per animation frame and verify I get just one translate elements out of it.

PR #125, opened 2018-01-09
Add mapContextAsync

stasm:
`mapContextAsync` asynchronously maps an identifier or an array of identifiers to the best `MessageContext` instance(s) from an async iterable.

The implementation is the easy part.

  - [x] Implement `mapContextAsync`.
  - [x] Write tests.

The hard part is getting our build and testing toolchain to parse `for await` :( Here's the list of things that need fixing:

  - [x] Support `for await` in Rollup which bundles the dist files.

    The [async iterators propsal](https://github.com/tc39/proposal-async-iteration) is still at [Stage 3](https://github.com/tc39/proposals/blob/2f701f68c48f4044c161c3ad290ce0350a09b457/README.md#stage-3). Acorn, the parser that Rollup uses, only supports Stage 4 proposals. There's a [plugin](https://github.com/adrianheine/acorn-async-iteration) which adds support for `for await` but it's currently not possible to extend Acorn via Rollup options. I submitted a PR to Rollup to allow adding plugins to Acorn (https://github.com/rollup/rollup/pull/1857).

  - [x] Support `for await` in Babel which transpiles dist files to compat files.

    Once Rollup bundles the dist files, we use Babel to transpile it to an older version of ECMAScript. There's a [transform-async-generator-functions plugin](https://www.npmjs.com/package/babel-plugin-transform-async-generator-functions) which hopefully can help Babel parse and transpile `for await` properly.

  - [x] Support `for await` when running tests.

    We use `babel-register` when running tests so that node imports ES modules properly. If we continue using it for this reason, we also need Babel to parse `for await` properly (cf. `transform-async-generator-functions` from above). Or, we could try using [@std/esm](https://github.com/standard-things/esm) which is designed just for this purpose.

  - [x] Support `for await` in the linter.

    Eslint also doesn't support non-Stage 4 proposals. There's [babel-eslint](https://github.com/babel/babel-eslint) which we might be able to use. Or, since Eslint uses Espree which is built on top of Acorn, maybe we can plug `acorn-async-iteration` into it somehow.

Yay, JavaScript!

### DIFF
--- a/fluent/src/fallback.js
+++ b/fluent/src/fallback.js
@@ -48,7 +48,6 @@
  * @param {string|Array<string>} ids
  * @returns {MessageContext|Array<MessageContext>}
  */
-
 export function mapContextSync(iterable, ids) {
   if (!Array.isArray(ids)) {
     return getContextForId(iterable, ids);
@@ -70,4 +69,44 @@ function getContextForId(iterable, id) {
   }
 
   return null;
+}
+
+/*
+ * Asynchronously map an identifier or an array of identifiers to the best
+ * `MessageContext` instance(s).
+ *
+ * @param {AsyncIterable} iterable
+ * @param {string|Array<string>} ids
+ * @returns {Promise<MessageContext|Array<MessageContext>>}
+ */
+export async function mapContextAsync(iterable, ids) {
+  if (!Array.isArray(ids)) {
+    for await (const context of iterable) {
+      if (context.hasMessage(ids)) {
+        return context;
+      }
+    }
+  }
+
+  let remainingCount = ids.length;
+  const foundContexts = new Array(remainingCount).fill(null);
+
+  for await (const context of iterable) {
+    // XXX Switch to const [index, id] of id.entries() when we move to Babel 7.
+    // See https://github.com/babel/babel/issues/5880.
+    for (let index = 0; index < ids.length; index++) {
+      const id = ids[index];
+      if (!foundContexts[index] && context.hasMessage(id)) {
+        foundContexts[index] = context;
+        remainingCount--;
+      }
+
+      // Return early when all ids have been mapped to contexts.
+      if (remainingCount === 0) {
+        return foundContexts;
+      }
+    }
+  }
+
+  return foundContexts;
 }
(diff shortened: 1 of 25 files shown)


stasm:
Is there a chance that the Async Iteration proposal will advance to Stage 4 at the next TC39 meeting ([23-25 January 2018](https://github.com/tc39/agendas/blob/master/2018/01.md))? All of the above would be magically solved in a matter of weeks.

zbraniecki:
Cool! Maybe this helps: https://github.com/tc39/ecma262/pull/1063

stasm:
Longer term, yes, it does, thanks. Short term, I just got a green build :)

stasm:
@zbraniecki This is ready for your review. You'll be most interested in the two last commits, but feel free to take a look at all of them. Thanks.

fluent/src/fallback.js
```diff
@@ -71,3 +70,36 @@ function getContextForId(iterable, id) {
...
+  for await (const context of iterable) {
+    // XXX Switch to const [index, id] of id.entries() when we move to Babel 7.
+    // See https://github.com/babel/babel/issues/5880.
+    for (let index = 0; index < ids.length; index++) {
+      const id = ids[index];
+      if (!foundContexts[index] && context.hasMessage(id)) {
+        foundContexts[index] = context;
+      }
+    }
+  }
```
zbraniecki: Should there be an early exit from the loop in case all contexts are found?

zbraniecki: If I'm correct, please add a test that verifies that contexts are not unnecessarily resolved.

stasm: Good point, thanks!

stasm:
My [PR to allow injecting Acorn plugins in Rollup](https://github.com/rollup/rollup/pull/1857) has been merged today and version [0.55.0](https://github.com/rollup/rollup/milestone/10) will likely be released soon. I'll wait for it before I merge this PR so that we don't have to use my branch as a dependency in `package.json`.

stasm:
rollup [0.55.0 has be released](https://github.com/rollup/rollup/releases/tag/v0.55.0). I updated the dependency to the upstream rollup repo.

PR #130, opened 2018-01-22
Add private messages

stasm:
Implement https://github.com/projectfluent/fluent/issues/62.


fluent-syntax/src/errors.js
```diff
@@ -45,6 +46,14 @@ function getErrorMessage(code, args) {
       return 'Expected literal';
     case 'E0015':
       return 'Only one variant can be marked as default (*)';
+    case 'E0016':
```
stasm: fluent-rs uses smaller errors numbers for these because it seems to be missing a few earlier ones.

fluent-syntax/src/parser.js
```diff
@@ -693,17 +718,25 @@ export default class FluentParser {
       throw new ParseError('E0014');
     }
 
+    if (ch === '$') {
```
stasm: I moved the `$` check to the top as this will likely be the most common literal.

fluent-syntax/src/serializer.js
```diff
@@ -28,7 +28,7 @@ export default class FluentSerializer {
     }
 
     for (const entry of resource.body) {
-      if (entry.types !== 'Junk' || this.withJunk) {
+      if (entry.type !== 'Junk' || this.withJunk) {
```
stasm: A nasty typo but fortunately didn't impact many tests. It made the serializer serialize Junks verbatim as they appear in the parsed text. Input magically always matched the output...

fluent/src/parser.js
```diff
@@ -886,7 +944,7 @@ class RuntimeParser {
 
         if ((cc >= 97 && cc <= 122) || // a-z
             (cc >= 65 && cc <= 90) || // A-Z
-             cc === 95 || cc === 47 || cc === 91) { // _/[
+             cc === 47 || cc === 91) { // /[
```
stasm: I think this shouldn't check for `/` not `[` anymore? I'll fix it in a follow-up

fluent-syntax/src/errors.js
zbraniecki: I'd like to make error naming part of the developer productivity review after 0.7.

fluent-syntax/src/ftlstream.js
```diff
@@ -93,20 +93,36 @@ export class FTLParserStream extends ParserStream {
...
-    const cc = this.ch.charCodeAt(0);
-    return ((cc >= 48 && cc <= 57) || cc === 45); // 0-9
+    if (this.currentIs('-')) {
+      this.peek();
+    }
+
+    const cc = this.currentPeek().charCodeAt(0);
+    const isDigit = cc >= 48 && cc <= 57; // 0-9
+    this.resetPeek();
+    return isDigit;
```
zbraniecki: what's the value of this change?

fluent-syntax/src/ftlstream.js
```diff
@@ -262,13 +278,20 @@ export class FTLParserStream extends ParserStream {
     }
   }
 
-  takeIDStart() {
-    if (this.isIDStart()) {
+  takeIDStart(allowPrivate) {
+    if (allowPrivate && this.currentIs('-')) {
+      this.next();
+      return '-';
+    }
```
zbraniecki: You should be able to `return this.next();` here.

stasm: We want to return `current` here but first, we advance the `index`.

fluent-syntax/src/ftlstream.js
stasm: It makes `isNumberStart` consistent with `isMessageIDStart` and more importantly, it considers the char after the leading `-`. The previous version would return true upon seeing a dash right away and more checking was required in `getLiteral`.

PR #132, opened 2018-01-23
Require = after the id (but not really)

stasm:
In Fluent Syntax 0.5, the identifier must be followed by an equals sign (=)
even if the message doesn't have a value.

    key =
        .attribute = Attribute Value

For some time fluent.js will allow both the new (0.5) and the old 0.4 syntax,
which means that the following will still parse but the Serializer will insert
the = after the key.

    key
        .attribute = Attribute Value

This patch also fixes Bug 1406880 - Decide where multiline Pattern spans should
start. The patterns now start when their first

    key =
        Value
        ^---- Pattern span start

Lastly, this patch forbids null variant values. Empty values should be written
with the {""} idiom:

    key =
        {
           *[valid] {""}
            [invalid]
        }


stasm:
@zbraniecki This isn't ready to land yet; I'm still working on implementing the same thing in the runtime parser. The full AST parser is good to go, however. Perhaps you'd like to take a look?

I had to rework most of `getMessage` because it used to always go to `getPattern` upon seeing a `=` which isn't the proper behavior for Syntax 0.5.

fluent/src/parser.js
```diff
@@ -121,60 +121,42 @@ class RuntimeParser {
...
-        this.skipBlankLines();
-        if (this._source[this._index] === ' ') {
-          this.skipInlineWS();
-          val = this.getPattern();
-        }
-      } else {
-        // This is a fast-path for the most common
-        // case of `key = Value` where the value
-        // is in the same line as the key.
-        val = this.getPattern();
```
stasm: This fast path wasn't needed because it's essentially almost the same as what `getPattern` does. Overall, I'm seeing a slight speed-up with my patch, around -7%.

fluent-syntax/src/ftlstream.js
```diff
@@ -46,12 +56,11 @@ export class FTLParserStream extends ParserStream {
     }
   }
 
-  skipInlineWS() {
-    while (this.ch) {
-      if (!includes(INLINE_WS, this.ch)) {
-        break;
-      }
-      this.next();
+  maybeExpectIndent() {
```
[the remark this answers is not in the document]
stasm: I ended up removing this method and replacing it with `ps.isPeekPatternStart` checks combined with a simple `ps.skipIndent`.

fluent/test/fixtures_structure/attribute_with_empty_pattern.json
```diff
@@ -0,0 +1,43 @@
+{
+  "key1": {
+    "val": "Value 1",
+    "attrs": {
+      "attr": {
+        "val": null
```
stasm: These are all syntax errors in the tooling parser. In the runtime parser I chose to parse them as `null` values, at least for now, to reduce the complexity of this PR and to keep the performance benefits.

stasm:
@zbraniecki This is now ready for your review. I wrestled a bit with the runtime parser; I think I finally managed to get it _acceptably right_ :) Let me know what you think, thanks!

fluent/src/resolver.js
```diff
@@ -294,7 +294,8 @@ function Type(env, expr) {
     }
     case undefined: {
       // If it's a node with a value, resolve the value.
-      if (expr.val !== undefined) {
+      // eslint-disable-next-line eqeqeq
+      if (expr.val != null) {
```
zbraniecki: why `!=` instead of `!==`?

stasm: To check both for `null` and `undefined`; see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness.

stasm: Just to make it clear in case it came across wrong: I don't suppose you don't know the difference :) Just wanted to link to the `Loose comparison` matrix for reference of what `!= null` evaluates to.

zbraniecki: since we usually use strict comparison, can you add a comment on why we use a different strategy here?

stasm: I'll just use the more explicit double comparison. Can't hurt the readability. Thanks!

PR #134, opened 2018-01-24
Expose FluentSerializer.serializeExpression

stasm:
@mathjazz wrote the following snippet for Pontoon and asked if we could expose it from `fluent-syntax`.

https://gist.github.com/mathjazz/4be883ea2562e2cbb59ca26beecf1e93

Yes, yes we could.

### DIFF
--- a/fluent-syntax/src/serializer.js
+++ b/fluent-syntax/src/serializer.js
@@ -18,6 +18,10 @@ export default class FluentSerializer {
   }
 
   serialize(resource) {
+    if (resource.type !== 'Resource') {
+      throw new Error(`Unknown resource type: ${resource.type}`);
+    }
+
     const parts = [];
 
     if (resource.comment) {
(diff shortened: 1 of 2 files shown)


stasm:
@mathjazz This exposes the `serializeExpression` method on instances of the `FluentSerializer` class. Let me know if this works for your use-case.

fluent-syntax/src/serializer.js
```diff
@@ -176,7 +184,13 @@ function serializePlaceable(placeable) {
...
       return `{${serializePlaceable(expr)}}`;
     case 'SelectExpression':
-      return `{${serializeSelectExpression(expr)}}`;
+      // Special-case select expression to control the whitespace around the
+      // opening and the closing brace.
+      return expr.expression
+        // A select expression with a selector.
+        ? `{ ${serializeSelectExpression(expr)}}`
+        // A variant list without a selector.
+        : `{${serializeSelectExpression(expr)}}`;
```
zbraniecki: should we use a single space before and after the expression?

stasm: This is special-cased so that there aren't any spaces around it. Consider:

```properties
-brand-name =
    {    ← no space after the opening {
       *[nominative] Firefox
    }    ← no extra space before the closing {, just the 4-space indent
```

PR #135, opened 2018-01-25
Runtime behavior tests

stasm:
This doesn't quite add behavior tests to the runtime parser because the runtime parser doesn't produce the same errors as the tooling one. What we can do, however, is parse all fixtures in `fluent-syntax` and check if the runtime parser parsed the same messages (check if keys are present) and if so, if their interface is the same (is value present or missing? which attributes are defined?).


common.mk
```diff
@@ -24,11 +24,13 @@ ifneq (,$(wildcard ./test/__setup.js))
 	@mocha --recursive --ui tdd \
 	    --require babel-register \
 	    --require babel-polyfill \
-	    --require ./test/__setup
+	    --require ./test/__setup \
+	    test/
```
stasm: I was getting some strange `unknown option --recursive` errors from `mocha` without this.

fluent-syntax/src/parser.js
```diff
@@ -692,10 +692,14 @@ export default class FluentParser {
     ps.expectChar('"');
 
     let ch;
-    while ((ch = ps.takeChar(x => x !== '"'))) {
+    while ((ch = ps.takeChar(x => x !== '"' && x !== '\n'))) {
       val += ch;
     }
 
+    if (ps.currentIs('\n')) {
+      throw new ParseError('E0020');
```
stasm: I submitted https://github.com/projectfluent/fluent/pull/83 to reflect this change in the spec.

fluent/src/parser.js
```diff
@@ -619,19 +619,18 @@ class RuntimeParser {
...
-    if (this._source[this._index] === ')') {
-      return args;
-    }
-
     while (this._index < this._length) {
       this.skipInlineWS();
 
+      if (this._source[this._index] === ')') {
+        return args;
+      }
```
stasm: Moving this here allows the last argument to be followed by a comma. I filed https://github.com/projectfluent/fluent/pull/84 to allow this in the spec.

fluent/test/parser_behavior_test.js
```diff
@@ -20,20 +19,40 @@ readdir(fixtures, function(err, filenames) {
...
+      const ftlpath = join(ftlFixtures, ftlfilename);
+      const jsonpath = join(jsonFixtures, jsonfilename);
+
+      test(ftlfilename, async function() {
+        const [source, json] = await Promise.all([ftlpath, jsonpath].map(readfile));
+
+        const [entries] = parse(source);
+        const expectedEntries = JSON.parse(json);
+
+        for (const [id, expected] of Object.entries(expectedEntries)) {
```
stasm: Iterating over `expectedEntries` here means that the runtime parser could possibly parse _more_ entries and we wouldn't learn about it from these tests. I'd like to fix this in a follow-up later on.

fluent/test/parser_structure_test.js
```diff
@@ -24,16 +24,15 @@ readdir(ftlFixtures, function(err, filenames) {
       const jsonfilename = ftlfilename.replace(/ftl$/, 'json');
       const ftlpath = join(ftlFixtures, ftlfilename);
       const jsonpath = join(jsonFixtures, jsonfilename);
-      test(ftlfilename, function() {
-        return Promise.all(
+      test(ftlfilename, async function() {
+        const [ftl, expected] = await Promise.all(
```
stasm: No functional changes here; just switching to `async`.

stasm:
This is PR consists of 3 commits: the first one adds the tests and the other two fix the differences in parsing between the  runtime and the tooling parser.

PR #136, opened 2018-01-29
Rename private messages to terms

stasm:
Implement https://github.com/projectfluent/fluent/pull/86.


fluent-syntax/src/ast.js
```diff
@@ -49,6 +49,17 @@ export class Message extends Entry {
   }
 }
 
+export class Term extends Entry {
+  constructor(id, value = null, attributes = [], comment = null) {
```
Pike: `value` isn't optional, I'd not give this a default value.

PR #141, opened 2018-02-01
fluent-react: Filter props with <Localized attrs={{…}}>

### LINKED ISSUE #139: Allow developers to create attribute filters for elements wrapped in <Localized>
Part of #103.

Right now `<Localized>` will apply all attributes found in the translation to the element wrapped inside of it. This may potentially override important props.

We should allow developers to control which attributes will be applied. For the initial release of overlays in `fluent-react` I think it's OK to err on the side of explicitness and _only allow_ attributes which are explicitly declared by the developer.

```properties
button = This button has a title attribute.
    .title = Title
    .other = This attribute will not be applied to the translation.
```

```jsx
<Localized id="button" attrs="title">
    <button>{'This button has a title attribute.'}</button>
</Localized>
```

What should be the type of the `attrs` prop?

  - A string, ssimilar to `className`: `attrs="title placeholder"`. 
  - An array: `attrs={["title", "placeholder"]}`.

stasm:
Fixes #139.

The <Localized> component now requires the `attrs` prop to set any localized attributes as props on the wrapped component. `attrs` should be an object with attribute names as keys and booleans as values.

```jsx
<Localized id="type-name" attrs={{placeholder: true}}>
    <input
        type="text"
        placeholder="Type your name"
        onChange={…}
        value={name}
    />
</Localized>
```

By default, if `attrs` is not passed, no attributes will be set. This is a breaking change compared to the previous behavior: in `fluent-react` 0.4.1 and before `<Localized>` would set _all_ attributes found in the
translation.

### DIFF
--- a/fluent-react/CHANGELOG.md
+++ b/fluent-react/CHANGELOG.md
@@ -50,8 +50,59 @@
     the element passed as a prop is cloned with the translated text content
     taken from the `DocumentFragment` used as `children`.
 
+  - Filter props set by translations with <Localized attrs={{…}}>.
 
- #### Migrating from `fluent-react` 0.4.1
+    The `<Localized>` component now requires the `attrs` prop to set any
+    localized attributes as props on the wrapped component. `attrs` should be
+    an object with attribute names as keys and booleans as values.
+
+    ```jsx
+    <Localized id="type-name" attrs={{placeholder: true}}>
+        <input
+            type="text"
+            placeholder="Localizable placeholder"
+            value={name}
+            onChange={…}
+        />
+    </Localized>
+    ```
+
+    By default, if `attrs` is not passed, no attributes will be set. This is
+    a breaking change compared to the previous behavior: in `fluent-react`
+    0.4.1 and before `<Localized>` would set _all_ attributes found in the
+    translation.
+
+#### Migrating from `fluent-react` 0.4.1 to 0.6.0
+
+If you're setting localized attributes as props of elements wrapped in
+`<Localized>`, in `fluent-react` 0.6.0 you'll need to also explicitly allow
+the props you're interested in using the `attrs` prop. This protects your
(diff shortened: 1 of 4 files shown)


stasm:
@hkasemir Would you mind reviewing this?

fluent-react/CHANGELOG.md
```diff
@@ -50,8 +50,56 @@
...
+    />
+</Localized>
+```
+
+```jsx
+// AFTER (fluent-react 0.6.0)
+<Localized id="type-name" attrs={{placeholder: true}}>
+    <input
+        type="text"
+        onChange={…}
```
hkasemir: might include a demo `placeholder` attribute here

fluent-react/test/localized_render_test.js
```diff
@@ -69,6 +133,69 @@ foo
...
+        <div existing={true} />
+      </Localized>,
+      { context: { l10n } }
+    );
+
+    assert.ok(wrapper.contains(
+      <div existing={true} />
+    ));
+  });
+
```
hkasemir: nice test coverage - makes the benefits of explicitly disallowing setting attrs clear.

fluent-react/src/localized.js
```diff
@@ -93,13 +93,29 @@ export default class Localized extends Component {
...
     const [args, elems] = toArguments(this.props);
-    const { value, attrs } = l10n.formatCompound(mcx, msg, args);
+    const {
+      value: messageValue,
+      attrs: messageAttrs
+    } = l10n.formatCompound(mcx, msg, args);
+
+    // The default is to forbid all message attributes. If the attrs prop exists
+    // on the Localized instance, only set message attributes which have been
+    // explicitly allowed by the developer.
```
hkasemir: 👍

Issue #128, opened 2018-01-17
Allow reordering child elements in translations in DOM overlays


stasm:
In `fluent-dom` it is currently not possible to reorder child elements found in the translation if they're of the same type. An example from [bug 1424682](https://bugzilla.mozilla.org/show_bug.cgi?id=1424682):

```properties
# This string is shown to notify the user how to enable an extension that they disabled.
extension-controlled-enable = To enable the extension go to <image/> Add-ons in the <image/> menu.
```

stasm:
The first design principle of Fluent is [_Control and Isolation_](https://github.com/projectfluent/fluent/wiki/Design-Principles#1-control-and-isolation). In particular I'd like to highlight two statements: 

  - _The localizers should have control over the translation._
  - _The localization logic of one language doesn't leak into other other languages nor into the source code_.

I think #143 is a good first attempt of the issue. It satisfies the first statement but it breaks the second one. The developers should not have to put anything in the source code for localizers to be able to reorder elements. Instead they should write their source code in an as agnostic manner as possible. Localizers interested in reordering should be able to do so via some additional syntax or markup.

For this reason, I don't think we should give child elements _names_ but instead, allow localizers to use CSS's [structural pseudo-classes](https://www.w3.org/TR/selectors-3/#structural-pseudos) to match desired elements.

I'd go for something like the following to make it easy to recognize the CSS:

```html
<a data-l10n-select=":last-child">bar</a> <a data-l10n-select=":first-child">foo</a>
<a data-l10n-select=":nth-child(2)">bar</a> <a data-l10n-select=":nth-child(1)">foo</a>
<a data-l10n-select=":last-of-type">bar</a> <a data-l10n-select=":nth-of-type(1)">foo</a>
```

In code, we could use something like:

```js
let localName = sourceChildElement.localName;
let selector = translationChildElement.getAttribute('data-l10n-select');
sourceElement.querySelector(`${localName}${selector}`);
```

`data-l10n-select` or `data-l10n-match` would be my picks for the attribute name.

Pike:
I have a different take on this.

I think we're trying to achieve two things with DOM overlays:
* localizers can use richer markup to do italic as quotes, or the like
* programmers of DOM-like targets can pass DOM-like nodes as parameters to localization, so allow to mix up localizable and non-localizable attributes.

The current code is pretty good at the first, but doesn't serve us well on the latter (parameters).

[...]

PR #143, opened 2018-02-03
Add data-l10n-order to allow child elements to be ordered in DOM Overlays

zbraniecki:
Fixes https://github.com/projectfluent/fluent.js/issues/128

### DIFF
--- a/fluent-dom/src/overlay.js
+++ b/fluent-dom/src/overlay.js
@@ -144,7 +144,7 @@ function sanitizeUsing(translationFragment, sourceElement) {
     // If a child of the same type exists in sourceElement, take its functional
     // (i.e. non-localizable) attributes. This also removes the child from
     // sourceElement.
-    const sourceChild = shiftNamedElement(sourceElement, childNode.localName);
+    const sourceChild = getNamedElement(sourceElement, childNode);
 
     // Find the union of all safe attributes: localizable attributes from
     // childNode and functional attributes from sourceChild.
@@ -157,7 +157,7 @@ function sanitizeUsing(translationFragment, sourceElement) {
     translationFragment.replaceChild(mergedChild, childNode);
   }
 
-  // SourceElement might have been already modified by shiftNamedElement.
+  // SourceElement might have been already modified by getNamedElement.
   // Let's clear it to make sure other code doesn't rely on random leftovers.
   sourceElement.textContent = '';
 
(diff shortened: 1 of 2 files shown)


Pike:
@zbraniecki, can you describe on how this is working?

I wonder if `order` is the right term here. I'm looking at this from the POV of how hard it is for localizers to not break this. Do we have to use a full blown `data-l10n-foo` attribute?

zbraniecki:
> @zbraniecki, can you describe on how this is working?

If the `data-l10n-order` is set on the translation, we look for a matching element. If it's not, we take the first in line.

Then if the attr is set and we can't find a matching element, we take the first of the type that matches.

I think we can improve the logic, throw early if there's no match (I'd prefer the algo to be stricter and dumber initially, than to try to be smart).

> I wonder if order is the right term here.

I'm happy to bikeshed the name! I'm not convinced that `order` is the right word here, but it's not the worst. I like using `data-l10n-` prefix as it keeps all our attrs together making it more consistent.

> Do we have to use a full blown data-l10n-foo attribute?

I think so. It makes it pretty composable. Do you have an idea how to improve it?

Pike:
First things second, I think we should do something like this. Let's figure out the details.

One thing I'd do is move the examples from "number-like" to "name-like". In the current use-case, I'd use <image data-l10n-name="addons-settings-icon"> <image data-l10n-name="extension-icon">

Which also made me pick a name just now, `name` ;-). Also matches what you used in the code, right?

As for resurrection, I'm pondering and rambling. Not an opinion yet. I guess just "randomly" assigning one element to one element would be bad, in particular if you're looking at the retranslation use-case. It'd not be idempotent for an ordering-changing locale.

Should we strip all elements of a local-name if we there are multiple, but not specified by name?

Follow-up thought, should we add  "l10n-name-extensions-icon" to the classList of the element? I guess we shouldn't assume in general that all elements of the original DOM survive throughout the lifetime of the element. That way they'd be easy to access from CSS? True dat, also works via attribute selectors.

Also should have compare-locales checks for this.

fluent-dom/src/overlay.js
```diff
@@ -144,7 +144,7 @@ function sanitizeUsing(translationFragment, sourceElement) {
     // If a child of the same type exists in sourceElement, take its functional
     // (i.e. non-localizable) attributes. This also removes the child from
     // sourceElement.
-    const sourceChild = shiftNamedElement(sourceElement, childNode.localName);
+    const sourceChild = getNamedElement(sourceElement, childNode);
```
stasm: The reason why this is called `shift...` is that it removes elements from `sourceElement`. I don't think `getNameElement` conveys this meaning well enough.

stasm:
@zbraniecki Can we close this in favor of a new solution in line with the design discussed in #128?

Issue #144, opened 2018-02-04
DOMOverlays should allow developers to put any elements into the fragment


zbraniecki:
At the moment we use a single list of translatable elements as a white list of elements allowed to be placed in DOM Overlay fragments.

While the list makes sense as a whitelist of elements allowed for use by the localizer, the current situation is rather limiting for the developer who might want to put elements from outside of the list in the fragment.

The reason we limit it this way is because we couldn't figure out how to do re-translation, so we'll need to solve that first before we can enable this.

stasm:
The reason why re-translation is an issue here is that a translation might remove one of the source children in which case we can never recover it until the user reloads the documents (or restarts the app). Note that by _re-translation_ I mean any situation in which an element gets new translated content; it might happen when the user changes the language, but also when the args to translation change and the language remains the same.

Because of this behavior, the current overlay logic clones source elements rather than use the originals. It makes it impossible to attach event handlers to them which in turn makes it explicit that developers should not rely on source children for app logic because they might be removed by translations.

It's interesting to consider how this doesn't apply to `fluent-react` at all: the JSX holds the source elements at all time while the DOM is only a manifestation of a state produced by a single translation. Any subsequent re-translation will use the source elements defined in the JSX—not the ones present in the DOM.

In `fluent-dom` we currently only use the present DOM state as the source for the overlay logic. To be more like `fluent-react` we could store the child elements we find in the DOM during the first translation on the target element (the one with `data-l10n-id`, i.e. the one being translated) or in a `WeakMap` where the target element is the key. None of these approaches would work if the element is removed from the DOM and then reinserted, however. Perhaps all we need is a `Map` of `data-l10n-ids` to source children collected during the first translation?

Pike:
It's perfectly legal to have multiple DOM nodes with the same data-l10n-id, right? So I'd not key off of that.

But yes, a cache of nodes would be great. There's interesting questions about when to cache, because developers might change non-localizable attributes on _their_ nodes, and we should try to retain that. So probably pre-translate, update the cache, then translate.

stasm:
> It's perfectly legal to have multiple DOM nodes with the same data-l10n-id, right? So I'd not key off of that.

Yes, but that's a feature. Same `data-l10n-id`, same child elements as arguments :)

stasm:
BTW, in server-side rendering scenarios which also run the translation logic, this would again create the risk of a translation removing source children from the DOM.

Pike:
[...]

PR #146, opened 2018-02-05
Allow for <img/> element in DOM Overlays

zbraniecki:
Required for https://bugzilla.mozilla.org/show_bug.cgi?id=1424682


stasm:
I agree with @Pike. Let's discuss how to allow elements like `` in #128 and #144.

Issue #145, opened 2018-02-05
string extraction from react components and `getString`


hkasemir:
I've been fiddling with a string extraction script to pull out localizable messages from react components that have been wrapped with `<Localized>`.

initial (and very messy) code is here: https://github.com/hkasemir/fluent-react-pull-strings

One thing I realized is that the `getString` function from the `withLocalization` HOC does not allow for any string extraction - meaning values for the strings identified by ID passed to `getString` need to be manually added to an ftl file. Is there a chance we could add a third arg to `getString` that would include a "fallback" value and to allow for optional inline defining the string for extraction scripts?

Or perhaps there's a better way to handle this, but to illustrate this idea:
```js
const name = 'Grace';
alert(getString('alert-name', {name}, 'This is an alert for { $name }');
```

stasm:
Good idea. We could also use this third argument as the penultimate fallback—just before falling back to the `id` itself. Even without any interpolations and formatting it's still better than the `id`.

@hkasemir Would you like to submit a PR for this? The relevant code is here: https://github.com/projectfluent/fluent.js/blob/f676215a4dab988001647999d3e63e7797d8f0ab/fluent-react/src/localization.js#L65-L78

PR #147, opened 2018-02-06
add(getString): add third argument to getString for fallback message

hkasemir:
addresses issue https://github.com/projectfluent/fluent.js/issues/145
adds a third argument to the `getString` function in `withLocalized` wrapped components to allow for definition of a fallback message in case the message id is not fount in the message context.

Additional future benefit to be including fallback string in extraction scripts.

### DIFF
--- a/fluent-react/src/localization.js
+++ b/fluent-react/src/localization.js
@@ -65,11 +65,11 @@ export default class ReactLocalization {
   /*
    * Find a translation by `id` and format it to a string using `args`.
    */
-  getString(id, args) {
+  getString(id, args, fallback) {
     const mcx = this.getMessageContext(id);
 
     if (mcx === null) {
-      return id;
+      return fallback || id;
     }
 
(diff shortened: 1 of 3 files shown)


hkasemir:
@stas - I was sort of wondering if there would be a way to format the fallback string, but it didn't look like there was a good place to put it in there. would that be possible, or more trouble than it's worth?

stasm:
We'd need to parse the fallback string to be able to format it. It's possible (although not trivial), but I don't think we should do it. It would quickly teach developers that they don't need to maintain the `en-US` file at all :)

hkasemir:
oh, the commit message is following a convention I sometimes use at work that follows:

`change_type(scope): commit message` - if it's confusing I can edit the title

stasm:
It was confusing to me at first :) No worries though.

stasm:
Thanks again, @hkasemir!

PR #148, opened 2018-02-07
Backport MessageContext parser changes from fluent 0.6.0

stasm:
In order to make it easier to upgrade to the new syntax, we'll release `fluent` 0.4.3 which will keep the API compatibility with 0.4.2 and will be able to read Fluent Syntax 0.5 files (just like `fluent` 0.6.0).

These changes make `fluent` 0.4.x able to read both Syntax 0.4 and Syntax 0.5 files:

  - Supported Syntax 0.4 features:
    - `//` comments.
    - Sections (`[[ Foo ]]`).
    - Message definitions may omit the `=` after the identifier if they don't have a value.

  - Supported Syntax 0.5 features:
    - `#`, `##` and `###` comments.
    - Identifiers starting with a dash (`-`) are parsed as terms.

The only feature of the old Syntax 0.4 which isn't supported anymore are Tags. In practice we didn't see any of the current consumers of `fluent` 0.4.2 take advantage of tags. In Syntax 0.5, [tags were replaced by attribute defined on terms](https://github.com/projectfluent/fluent/blob/master/spec/CHANGELOG.md#050-january-31-2018).


compat_config.js
```diff
@@ -13,6 +13,7 @@ export default {
       ],
       'plugins': [
         'external-helpers',
+        'babel-plugin-optimize-starts-with',
```
stasm: I wanted to keep the code as close to `fluent@0.6.0` as possible. We started using `startsWith('-')` in 0.6.0 to tell terms and messages apart. This plugin converts it to `charCodeAt(0) == 45`.

fluent/test/fixtures_structure/elements_indent.json
```diff
@@ -7,10 +7,5 @@
     }
   },
   "baz": "Baz",
-  "qux": {
-    "val": "Qux",
-    "tags": [
-      "tag1"
-    ]
-  }
+  "qux": "Qux\n#tag1"
```
stasm: Tags are no longer parsed by the runtime parsers so this simply becomes a multiline message value.

stasm:
See also: all the changes in the [`fluent@0.4.2...fluent@0.6.0`](https://github.com/projectfluent/fluent.js/compare/fluent@0.4.2...fluent@0.6.0) range.

fluent/test/fixtures_structure/elements_indent.json
Pike: I think it'd be good to unsupport tags in the tests, and probably relnotes.

Removing the tags from the ftl sources feels more discoverable of the change actually happening.

fluent/test/fixtures_structure/sparse-messages.json
```diff
@@ -3,22 +3,15 @@
...
     }
   },
-  "key4": {
-    "val": "Value",
-    "tags": [
-      "tag1",
-      "tag2"
-    ]
-  },
+  "key4": "Value\n\n\n#tag1\n\n\n#tag2",
```
Pike: ... here, too.

fluent/src/parser.js
```diff
@@ -72,7 +74,8 @@ class RuntimeParser {
     const ch = this._source[this._index];
 
     // We don't care about comments or sections at runtime
-    if (ch === '/') {
+    if (ch === '/' ||
+      (ch === '#' && [' ', '#'].includes(this._source[this._index + 1]))) {
```
zbraniecki: You may need to skip also if `ch === '#' && ch2 === '\n'`.

fluent/test/fixtures_structure/elements_indent.json
stasm: I cherry picked ca5f6040266fe70c7adb72d3d9ab5322e99c6fb9 as the first commit in the PR.

fluent/src/parser.js
stasm: This has to be fixed also on master in #149. I'll backport it together with other changes when master becomes 0.6.1.

PR #151, opened 2018-02-08
Backport MessageContext parser from fluent 0.6.2

stasm:
- Correctly parse empty comment lines. (#149)
  - Forbid null attribute and variant values. (part of #150)


stasm:
In order to test the new runtime parser I had to update the fixtures of the tooling parser in `fluent-syntax`. I don't think updating the tooling parser to 0.6.x is worth the effort so instead, I simply generated the expected parse results from these updated fixtures using the 0.4.x `fluent-syntax` parser. To some extent this codifies how the parsing behavior changed between 0.4.x and 0.6.x. I'm not sure if there's a better way to handle these test updates.

PR #154, opened 2018-02-09
Preserve children of wrapped component if translation value is null

stasm:


stasm:
Good question! It was a subject of a heated debate (see https://github.com/projectfluent/fluent/issues/71 and https://github.com/projectfluent/fluent/issues/63 if you're interested) :). The latter form is currently not allowed. There are good reasons for that, too, mostly related to messages which have more than one attribute and the edge-cases related to reordering or forgetting to update the id in only one of the attribute definitions.

Gregoor:
Uuh, quite the controversial topic. I don't want to chime in too much (and I liked the distinction and reasons you gave for simple vs easy) but it might also be interesting to look at what users' default assumption is. E.g. Michal and me expected the sugar version to exist and be equivalent.

But as long as we get an error for using the wrong version I'm good (which I guess we currently don't because we're still on 0.4.x).

zbraniecki:
@Gregoor - That's very useful for us. Please, pay attention to what are your instinctive expectations for the syntax. We have time till 1.0 to exactly collect such feedback and it may help us make better syntax for many years to come.

The fact that there are areas where Fluent team expectations differ is a sign for me that we may get more requests like this later.
We don't have to rush deciding if we want to listed to such requests, but we can make it possible for ourselves in the future by not locking ourselves out today.

To give you example, here's a number of syntax sugar id+attr pairs that I would like to see us support: https://gist.github.com/zbraniecki/1a273839a8babbfb0d938070d85cacfc#file-attrs-ftl

And since my opinion is currently in minority, I can only rely on feedback from users to bring more validity to my suggestions :)

PR #155, opened 2018-02-09
Protect void elements from translations which try to set their children

stasm:
A broken translation may accidentally set a value which `<Localized>` will normally insert as a child of its wrapped component. For void elements we need to actively forbid this because React throws when trying to set children of void elements, breaking the entire app.

### DIFF
--- a/fluent-react/src/localized.js
+++ b/fluent-react/src/localized.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 
 import { isReactLocalization } from './localization';
 import { parseMarkup } from './markup';
+import VOID_ELEMENTS from '../vendor/voidElementTags';
 
 /*
  * Prepare props passed to `Localized` for formatting.
@@ -111,7 +112,15 @@ export default class Localized extends Component {
       }
     }
 
-    // If the message has a null value, we're onl interested in its attributes.
+    // If the wrapped component is a known void element, explicitly dismiss the
+    // message value and do not pass it to cloneElement in order to avoi the
+    // "void element tags must neither have `children` nor use
+    // `dangerouslySetInnerHTML`" error.
+    if (elem.type in VOID_ELEMENTS) {
+      return cloneElement(elem, localizedProps);
+    }
+
+    // If the message has a null value, we're only interested in its attributes.
     // Do not pass the null value to cloneElement as it would nuke all children
     // of the wrapped component.
     if (messageValue === null) {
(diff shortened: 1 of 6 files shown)


stasm:
@hkasemir Would you mind reviewing this?

fluent-react/src/localized.js
```diff
@@ -111,7 +112,15 @@ export default class Localized extends Component {
       }
     }
 
-    // If the message has a null value, we're onl interested in its attributes.
+    // If the wrapped component is a known void element, explicitly dismiss the
+    // message value and do not pass it to cloneElement in order to avoi the
+    // "void element tags must neither have `children` nor use
+    // `dangerouslySetInnerHTML`" error.
+    if (elem.type in VOID_ELEMENTS) {
+      return cloneElement(elem, localizedProps);
```
hkasemir: any chance we'd want to log a warning to the console to indicate an invalid message format? Or do we prefer silently handling the error?

fluent-react/test/localized_void_test.js
```diff
@@ -0,0 +1,70 @@
...
+
+    const wrapper = shallow(
+      <Localized id="foo" attrs={{title: true}}>
+        <input />
+      </Localized>,
+      { context: { l10n } }
+    );
+
+    assert.ok(wrapper.contains(
+      <input title="TITLE" />
```
hkasemir: 👍

fluent-react/src/void_elements/index.js
```diff
@@ -0,0 +1,23 @@
...
+  'hr': true,
+  'img': true,
+  'input': true,
+  'link': true,
+  'meta': true,
+  'param': true,
+  'source': true,
+  'track': true,
+  'wbr': true
+};
```
hkasemir: also a quick search of void elements produced a larger list than this:
https://github.com/wooorm/html-void-elements/

any reason we're using this smaller list?

hkasemir: though this does look suspiciously exactly like what is in the React code: https://github.com/facebook/react/blob/master/packages/react-dom/src/shared/omittedCloseTags.js

with [the addition of `menuitem`](https://github.com/facebook/react/blob/master/packages/react-dom/src/shared/voidElementTags.js)

stasm: Ah, nice find in React's source. I'll use their file instead of the `void-elements` package because what we really care about here is not a complete list of void elements but rather the list that React uses to throw errors.

fluent-react/src/localized.js
stasm: I'd like to tackle the problem of error reporting in a separate PR.

PR #157, opened 2018-02-21
Switch linter to double quotes

zbraniecki:
Back in a day when I crafted the list of eslint rules we use I made a mistake - settled on single quotes.

Nobody likes that, and its incompatible with Mozilla eslinting rules. Let's fix it.

This patch exclusively focuses on quotes because that's the only incompatibility between us and mozilla-central. We may want to further align our rules with mozilla ruleset, but that'll be optional and may be worth more discussions.

I basically ran `eslint **/src --fix`.

### DIFF
--- a/eslint_src.json
+++ b/eslint_src.json
@@ -73,7 +73,7 @@
         "prefer-spread": 2,
         "quotes": [
             2,
-            "single",
+            "double",
             {
                 "avoidEscape": true
             }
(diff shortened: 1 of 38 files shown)


stasm:
Thanks for this. Should we also make fixes to `eslint_test.json` and the test files?

zbraniecki:
Since the test files already use double quotes, I'd prefer to handle any test changes in a follow up if that's ok with you :)

stasm:
Sure, sounds good.

stasm:
🎉

PR #158, opened 2018-02-21
Skip placing FSI/PDI if a placeable is the only element in the Pattern

zbraniecki:
Currently we wrap every placeable in FSI/PDI to help layout work with bidi code.
This is unnecessary in cases where the placeble is the only element in the Pattern like:

```
key1 = { -brand-short-name }

key2 =
    {  PLATFORM() ->
        [windows] Foo
       *[other] Faa
    }
```

This patch optimizes that scenario.

### DIFF
--- a/fluent/src/resolver.js
+++ b/fluent/src/resolver.js
@@ -448,6 +448,10 @@ function Pattern(env, ptn) {
   dirty.add(ptn);
   const result = [];
 
+  // Wrap interpolations with Directional Isolate Formatting characters
+  // only when the pattern has more than one element.
+  const useIsolating = ctx._useIsolating && ptn.length > 1;
+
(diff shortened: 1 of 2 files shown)


fluent/src/resolver.js
```diff
@@ -448,6 +448,9 @@ function Pattern(env, ptn) {
   dirty.add(ptn);
   const result = [];
 
+  // Skip isolating if the pattern has only one element.
+  const useIsolating = ctx._useIsolating && ptn.length > 1;
```
stasm: The comment describes the `false` value while the name of the variable corresponds to the `true` value. Can you update the comment to something like the following please?

    Wrap interpolations with Directional Isolate Formatting characters
    only when the pattern has more than one element.

PR #159, opened 2018-02-21
Fix parsing of whitespace-only values

stasm:
This PR fixes two bugs I discovered while working on the Playground code.

  1. Inline whitespace should be allowed in null values.
  1. Indented EOF should not make the parser try to parse a Pattern.


fluent-syntax/src/parser.js
```diff
@@ -269,6 +269,7 @@ export default class FluentParser {
     // XXX Replace with ps.expectChar('=').
     if (ps.currentIs("=")) {
       ps.next();
+      ps.skipInlineWS();
```
zbraniecki: I'm not sure if I understand this change. You're skipping inline WS, but then `isPeekPatternStart` also peeks through it and then `skipIndent` skips it.

So, in which scenario it is needed?

stasm: We need to `skipInlineWS` regardless of whether there's a Pattern or not. Would this be more readable if I moved `skipInlineWS` into an `else` block of the `isPeekPatternStart` check?

PR #160, opened 2018-02-23
Improve error debugging and recovery on missing messages

zbraniecki:
The current error reporting for errors from Fluent is very limited. While we do plan a more complete rework of the developer experience in Q2, I started receiving feedback that this is impacting people's first experience of working with Fluent in Firefox right now. I believe it's worth landing a smaller fix that fixes the most noticeable issues.

This PR aims to fix three scenarios:

- Overreporting missing strings when falling back (see https://bugzilla.mozilla.org/show_bug.cgi?id=1438608 )
 - Not falling back on message id as last resort (see https://bugzilla.mozilla.org/show_bug.cgi?id=1410849 ) 
 - Not reporting untranslated strings

This does not address reporting errors out of resolver and I see it more as a simple imminent fix that vastly improves experience of engineers trying to work with Fluent today, rather than a long term solution.
I'd like to this in a minor release, and leave a more complete overhaul for stas' planned refactor (see https://bugzilla.mozilla.org/show_bug.cgi?id=1410857 )

### DIFF
--- a/fluent-dom/src/localization.js
+++ b/fluent-dom/src/localization.js
@@ -249,7 +249,7 @@ function messageFromContext(ctx, errors, id, args) {
  */
 function keysFromContext(method, ctx, keys, translations) {
   const messageErrors = [];
-  let missingIds = null;
+  const missingIds = new Set();
 
   keys.forEach((key, i) => {
     if (translations[i] !== undefined) {
@@ -261,9 +261,6 @@ function keysFromContext(method, ctx, keys, translations) {
       translations[i] = method(ctx, messageErrors, key[0], key[1]);
       // XXX: Report resolver errors
     } else {
-      if (missingIds === null) {
-        missingIds = new Set();
-      }
       missingIds.add(key[0]);
     }
   });
--- a/fluent-dom/src/dom_localization.js
+++ b/fluent-dom/src/dom_localization.js
@@ -282,7 +282,9 @@ export default class DOMLocalization extends Localization {
     this.pauseObserving();
 
     for (let i = 0; i < elements.length; i++) {
-      overlayElement(elements[i], translations[i]);
+      if (translations[i] !== undefined) {
+        overlayElement(elements[i], translations[i]);
+      }
     }
 
     this.resumeObserving();
--- a/fluent-dom/test/dom_localization_test.js
+++ b/fluent-dom/test/dom_localization_test.js
@@ -1,0 +1,46 @@
+import assert from "assert";
+import { MessageContext } from "../../fluent/src/index";
+import DOMLocalization from "../src/dom_localization";
+
+function* mockGenerateMessages(resourceIds) {
+  const mc = new MessageContext(["en-US"]);
+  mc.addMessages("key1 = Key 1");
+  yield mc;
+}
+
+const mockWindow = {
+  MutationObserver: class MutationObserver {
+    takeRecords() {return new Set();}
+    disconnect() {}
+  }
+};
+
+suite("translateFragment", function() {
+  test("translates a node", async function() {
+    const domLoc = new DOMLocalization(mockWindow, ["test.ftl"], mockGenerateMessages);
+
+    const frag = document.createDocumentFragment();
+    const elem = document.createElement("p");
+    domLoc.setAttributes(elem, "key1");
+    frag.appendChild(elem);
+
+    await domLoc.translateFragment(frag);
+
+    assert.equal(elem.textContent, "Key 1");
+  });
+
+  test("does not inject content into a node with missing translation", async function() {
+    const domLoc = new DOMLocalization(mockWindow, ["test.ftl"], mockGenerateMessages);
+
+    const frag = document.createDocumentFragment();
+    const elem = document.createElement("p");
+    domLoc.setAttributes(elem, "missing_key");
+    elem.textContent = "Original Value";
+    frag.appendChild(elem);
+
+    await domLoc.translateFragment(frag);
+
+    assert.equal(elem.textContent, "Original Value");
+  });
+
+});
--- a/fluent-dom/test/localization_test.js
+++ b/fluent-dom/test/localization_test.js
@@ -1,0 +1,27 @@
+import assert from "assert";
+import { MessageContext } from "../../fluent/src/index";
+import Localization from "../src/localization";
+
+function* mockGenerateMessages(resourceIds) {
+  const mc = new MessageContext(["en-US"]);
+  mc.addMessages("key1 = Key 1");
+  yield mc;
+}
+
+suite("formatMessages", function() {
+  test("returns a translation", async function() {
+    const loc = new Localization(["test.ftl"], mockGenerateMessages);
+    const translations = await loc.formatMessages([["key1"]]);
+
+    assert.equal(translations[0].value, "Key 1");
+  });
+
+  test("returns undefined for a missing translation", async function() {
+    const loc = new Localization(["test.ftl"], mockGenerateMessages);
+    const translations = await loc.formatMessages([["missing_key"]]);
+
+    // Make sure that the returned value here is `undefined`.
+    // This allows bindings to handle missing translations.
+    assert.equal(translations[1], undefined);
+  });
+});


Pike:
I'm concerned about adding more code-paths w/ async iteration right now, given that we might need to back some of that out for the package size of compat builds. If I'm reading this right.

zbraniecki:
That's just me being tired, sorry. The async cleanup is not part of this patch. Removed.

stasm:
Thanks, I'll review this tomorrow. Sorry for keeping you waiting.

fluent-dom/src/localization.js
```diff
@@ -273,27 +294,26 @@ function messageFromContext(ctx, errors, id, args) {
...
+    if (ctx.hasMessage(key[0])) {
+      messageErrors.length = 0;
+      const translation = method(ctx, messageErrors, key[0], key[1]);
+      if (messageErrors.length === 0 ||
+          !messageErrors.some(e => e instanceof L10nError)) {
+        translations[i] = translation;
+      } else {
+        hasMissing = true;
+      }
     } else {
```
stasm: Nit: `else if (!hasMissing) {`

fluent-dom/src/localization.js
```diff
@@ -55,17 +55,49 @@ export default class Localization {
...
+    if (hasMissing) {
+      keys.forEach((key, i) => {
+        if (translations[i] === undefined) {
+          translations[i] = { value: key[0], attrs: null };
+        }
+      });
+    }
+    if (debugInfo !== null && typeof console !== "undefined") {
+      if (hasMissing) {
+        console.warn("[Fluent] Could not resolve all keys.");
```
stasm: Perhaps:

    [Fluent] Some translations were missing in all current locales.

fluent-dom/src/localization.js
```diff
@@ -55,17 +55,49 @@ export default class Localization {
...
+      } else {
+        console.warn("[Fluent] Had to use fallback to resolve all keys.");
```
stasm: Perhaps:

    [Fluent] Some translations were missing in the first locale but were found in fallback locales.

fluent-dom/src/localization.js
```diff
@@ -55,17 +55,49 @@ export default class Localization {
...
     for (let ctx of this.ctxs) {
       // This can operate on synchronous and asynchronous
       // contexts coming from the iterator.
       if (typeof ctx.then === "function") {
         ctx = await ctx;
       }
-      const errors = keysFromContext(method, ctx, keys, translations);
-      if (!errors) {
+      if (keysFromContext(method, ctx, keys, translations)) {
+        hasMissing = false;
```
stasm: This will only set `hasMissing` to `false` if the current context has _all_ the translations. However, if the sum of two or more contexts is also complete but none of the context is complete individually, `hasMissing` will still be `true`.

fluent-dom/src/localization.js
stasm: Or remove this (see my comment below).

fluent-dom/src/localization.js
stasm: Or remove this (see my comment below).

fluent-dom/src/localization.js
```diff
@@ -55,17 +55,49 @@ export default class Localization {
...
+      });
+    }
+    if (debugInfo !== null && typeof console !== "undefined") {
+      if (hasMissing) {
+        console.warn("[Fluent] Could not resolve all keys.");
+      } else {
+        console.warn("[Fluent] Had to use fallback to resolve all keys.");
+      }
+      for (const [locale, ids] of debugInfo.entries()) {
+        console.warn(`Tried "${locale}". Missing keys - ${ids.join(", ")}`);
```
stasm: This is the most important information here. I think the minimal desired output should be something like:

    Missing translations in es-MX: foo, bar.
    Missing translations in es: baz.
    Missing translations in en-US: qux.

IIUC, you could return the list of missing keys from `keysFromContext` and print them right in the `for (let ctx of this.ctxs)` loop.

stasm:
> Not falling back on message id as last resort (see https://bugzilla.mozilla.org/show_bug.cgi?id=1410849 )

This PR doesn't seem to do anything about this particular point.

zbraniecki:
Thanks stas! I like your approach and will apply it :)

zbraniecki:
> This PR doesn't seem to do anything about this particular point.

That's not true. There is a part of `formatWithFallback` that handles this. I would like to look into improving that in DOMLocalization for lables later.

zbraniecki:
Updated to your feedback.

fluent-dom/src/localization.js
```diff
@@ -273,27 +282,24 @@ function messageFromContext(ctx, errors, id, args) {
...
+      translations[i] = method(ctx, messageErrors, key[0], key[1]);
+      // XXX: Report resolver errors
     } else {
-      hasErrors = true;
-    }
-
-    if (messageErrors.length && typeof console !== "undefined") {
-      messageErrors.forEach(error => console.warn(error));
+      if (missingIds === null) {
+        missingIds = new Set();
```
stasm: Nit: I wonder if it might be cheaper (and less code) to just create this set upfront rather than check if `missingIds` is null for every translation.

fluent-dom/src/localization.js
```diff
@@ -258,8 +267,8 @@ function messageFromContext(ctx, errors, id, args) {
  * we return it. If it does, we'll try to resolve the key using the passed
  * `MessageContext`.
  *
- * In the end, we fill the translations array, and return if we
- * encountered at least one error.
+ * In the end, we fill the translations array, and return `true` if all
```
stasm: This now returns a set of missing ids.

fluent-dom/src/localization.js
```diff
@@ -55,17 +37,55 @@ export default class Localization {
...
+          ctx.locales[0],
+          missingIds
+        );
+      }
+    }
+
+    if (hasMissing) {
+      keys.forEach((key, i) => {
+        if (translations[i] === undefined) {
+          translations[i] = { value: key[0], attrs: null };
```
stasm: This is dangerous without coordination with `overlayElement` or `applyTranslations`. Some DOM elements might not expect a value. Others might have their own children and might be only interested in translating the attributes.

fluent-dom/src/localization.js
```diff
@@ -55,17 +37,55 @@ export default class Localization {
...
       // This can operate on synchronous and asynchronous
       // contexts coming from the iterator.
       if (typeof ctx.then === "function") {
         ctx = await ctx;
       }
-      const errors = keysFromContext(method, ctx, keys, translations);
-      if (!errors) {
+      const missingIds = keysFromContext(method, ctx, keys, translations);
+      if (missingIds === null) {
+        hasMissing = false;
```
stasm: As I stated in my previous review, I don't think `hasMissing` does what you'd like it to do. It will be `true` if all locales have missing translations but their sum is complete. In which case `Could not resolve all keys.` is incorrect.

I would remove `debugInfo` and `hasMissing` entirely from this patch. Would something like the following be enough?

```js
// Always return a set from keysFromContext, even if empty.
const missingIds = keysFromContext(method, ctx, keys, translations);

// Maybe factor out to a separate function?
if (typeof console !== "undefined") {
  for (const missingId of missingIds) {
    console.warn(
      `[Fluent] Missing translations in ${locale}: ${Array.from(ids).join(", ")}`);
  }
}
```
```

**Comment by stasm** [comment]
> That's not true. There is a part of formatWithFallback that handles this. I would like to look into improving that in DOMLocalization for lables later.

Ah, sorry, I missed that in the original review. Good thing that you brought my attention to it. I think we shouldn't do this right now because it might unintentionally replace important DOM trees with the identifier of a missing message.

Nit for the future: please don't amend the PR while the review is ongoing. It makes it harder to see the interdiffs. You can always squash/rebase before landing.

File: fluent-dom/src/localization.js (hunk 8 above)
Comment (stasm) [reply fluent-dom/src/localization.js]: Can you please add a test for this so that we don't regress in the future?

File: fluent-dom/src/localization.js (hunk 9 above)
Comment (stasm) [reply fluent-dom/src/localization.js]: I stand corrected: we're passing `translations` to `keysFromContext` so that subsequent iterations of the loop will only grow the list of retrieved translations. This doesn't change my suggestion however :)

File: fluent-dom/src/localization.js [hunk 10]
```diff
@@ -55,17 +37,55 @@ export default class Localization {
...
+          debugInfo = new Map();
+        }
+        debugInfo.set(
+          ctx.locales[0],
+          missingIds
+        );
+      }
+    }
+
+    if (hasMissing) {
```
stasm: By _remove `hasMissing`_ I meant removing this block. We still need the variable to break out of the loop, of course.

fluent-dom/src/localization.js
zbraniecki: My logic here is that this is an error scenario. Even if creating it is cheaper for when there's a single error in 100 strings, it is more expensive if there are zero errors which is 99% of production cases.

I'm optimizing for the scenario where this branch of `if/else` is never hit in production.

fluent-dom/src/localization.js
zbraniecki: I have second thought on whether I should fix it here. I've seen anecdotally cases where a missing string would turn:

```
<window data-l10n-id="key1">
   ... you know, whole browser.xul
</window>
```

into:

```
<window data-l10n-id="key1">
  key1
</window>
```

So I'm going to remove it here and let's continue conversation about how to approach it.

fluent-dom/src/localization.js
stasm: I understand the motivation. I don't agree with it. Creating a single empty Set per context is cheap and it spares you the trouble of having to check if it's `null` everywhere which is prone to errors. And which, by the way, the code in `formatWithFallback` doesn't do.

fluent-dom/src/localization.js
```diff
@@ -55,17 +37,22 @@ export default class Localization {
...
         ctx = await ctx;
       }
-      const errors = keysFromContext(method, ctx, keys, translations);
-      if (!errors) {
-        break;
+      const missingIds = keysFromContext(method, ctx, keys, translations);
+
+      if (typeof console !== "undefined") {
+        const locale = ctx.locales[0];
+        const ids = Array.from(missingIds).join(", ");
```
stasm: With `missingIds` being a nullable set, this should first check if `missingIds !== null`. I'd prefer if `keysFromContext` always returned a set, even an empty one.

fluent-dom/src/localization.js
```diff
@@ -268,32 +244,29 @@ function messageFromContext(ctx, errors, id, args) {
...
  * @param {{Array<{value: string, attrs: Object}>}} translations
  *
- * @returns {Boolean}
+ * @returns {Set<string>}
  * @private
  */
 function keysFromContext(method, ctx, keys, translations) {
   const messageErrors = [];
-  let hasErrors = false;
+  let missingIds = null;
```
stasm: The docstring says `keysFromContext` returns a Set but this is set to `null` here.

fluent-dom/src/localization.js
stasm: Yes, that's what I meant. Let's not touch it here. A test would be nice to have to make sure we don't regress in the future (like one of the iterations of this PR would have).

PR #163, opened 2018-03-07
Update fluent-web to preserve resource insertion order

zbraniecki:
Fixing issues raised by stas in https://github.com/projectfluent/fluent.js/commit/50d553796139abf2f07ae6ac3d04cf836c79edb0

I didn't add any code to handle network errors because it doesn't seem to be trivial - es2018 doesn't have any equivalent of `Promise.all` that would finish all promises even if one fails.

I think we can handle it in a separate PR.

### DIFF
--- a/fluent-web/src/index.js
+++ b/fluent-web/src/index.js
@@ -34,14 +34,24 @@ function getResourceLinks(elem) {
   );
 }
 
-async function generateContext(locale, resourceIds) {
+async function fetchResource(locale, id) {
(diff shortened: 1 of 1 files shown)


stasm:
> I didn't add any code to handle network errors because it doesn't seem to be trivial - es2018 doesn't have any equivalent of Promise.all that would finish all promises even if one fails.

Yeah, I meant adding error handling to `fetchResource` so that it only ever returns resolved Promises.

PR #164, opened 2018-03-08
Relax the engines support in package.json for node6 support for fluent-syntax

EnTeQuAk:
This blocks mozilla/addons-linter#1789 (and thus the langpack signing
for Firefox 60).

fluent-syntax is compatible with node 6 when using the compat build but
still can't be installed given that it requires node 8+ in the
packages.json.

Can we relax that a little bit?

Addons-linter is a dependency of `web-ext` which is being installed on
end-users machines and thus has a bit more LTS requirements and has to
support node6 while it's still LTS.

I added the `ignore-engines` option to the addons-linter tests to check that the build really does support node 6 and all tests are passing.

See https://github.com/projectfluent/fluent.js/pull/164 for more details.


stasm:
Hi @EnTeQuAk, thanks for the PR.

The Node 8+ requirement was introduced in #95 and the motivation was https://github.com/projectfluent/fluent.js/issues/92#issuecomment-344785444. Node 6 doesn't support `Object.entries()`. Even though this was specifically a problem in the `fluent` package, I made the change for all packages in the repo, including `fluent-syntax`, to make maintenance easier and avoid similar bug in the future. Note that the compat builds don't solve this issue: `Object.entries` needs to be polyfilled via `core-js/fn/object/entries` rather than transpiled.

While `fluent-syntax` currently doesn't use `Object.entries()` , `fluent` and `fluent-react` do. It would make the maintenance of `fluent-syntax` much harder if we had to remember that it's the only package in the repo where you can't use certain JS features :( Ideally we'd run the test suite on Node 6 as well to prevent bugs in the future, but then `fluent` and `fluent-react` would break.

I'll do my best to unblock mozilla/addons-linter#1789 in the short term. Perhaps we can relax the requirement for a few point releases?

Long term -- what's your timeline for phasing out Node 6 support? Please say Firefox 61 :)

stasm:
BTW Is this a yarn-only thing?

EnTeQuAk:
Ah, thanks for the hint regarding the compat transpilation. Well, it seems the compat build is working for the stuff that I use but it could break at other points it seems.

> Long term -- what's your timeline for phasing out Node 6 support? Please say Firefox 61 :)

That depends on how `web-ext` phases it out but currently I plan to remove Node 6 support when node6 goes into "Maintenance LTS"-mode which is April 30th 2018, so not too long.

Alternatively, I could fork the fluent-syntax package temporarily, rewrite whatever is node 8-only and publish that for the time being. I could even try using an older fluent-syntax package but I'd rather not go backwards here tbh. Or we could merge those changes temporarily but that increases maintainance burden on your end.

What do you think?

stasm:
> That depends on how web-ext phases it out but currently I plan to remove Node 6 support when node6 goes into "Maintenance LTS"-mode which is April 30th 2018, so not too long.

OK, this is great to know. I was mostly afraid of having different engine requirements across different packages in the repo. I'm sure down the road I'd forget and try to use some new JS feature which would break `fluent-syntax` on Node 6 :)

If we put a time limit on this, I think that's fine and easier to remember. I'll also try to set Travis up such that it runs `fluent-syntax` tests on Node 6.

stasm:
@Pike Does this look like a sane way forward until May?

fluent-syntax/makefile
```diff
@@ -23,6 +23,10 @@ compat.js: $(SOURCES)
 	    --output.file $@
 	@echo -e " $(OK) $@ built"
 
+.PHONY: test-without-babel-polyfill
+test-without-babel-polyfill:
+	@mocha --recursive --ui tdd --require babel-register test/
```
stasm: We still need `babel-register` to transpile ES6 syntax, just like the compat builds do. Thanks to this, `import`s work correctly, for instance.

.travis.yml
```diff
@@ -1,7 +1,10 @@
 sudo: false
 language: node_js
-script: make deps && make dist
+script:
+  - make deps && make dist
+  - if [[ $(node --version) = v6.* ]]; then make -C fluent-syntax test-without-babel-polyfill; fi
```
Pike: Any reason to only test this for v6 and not for v8 as well?

stasm: No reason, except that I fixated on Node 6 :)

PR #168, opened 2018-03-14
Overlay child elements by name 

### LINKED ISSUE #128: Allow reordering child elements in translations in DOM overlays

### LINKED ISSUE #144: DOMOverlays should allow developers to put any elements into the fragment

stasm:
### Background 

The overlay mechanic allows some safe HTML to be present in translations. Previously only text-level elements were allowed in translations. Additionally, if a element of the same type was present in the source HTML, its functional attributes were copied to the translated element.  This approach allowed translations to use the `<a>` element and have its `href` attribute copied from the source. It was limited, however, to text-semantic elements (`` or `<button>` were not supported) and to overlaying child elements in the order defined by the source.

### Use-Cases

This PR refactors the mechanism for sanitizing the HTML markup found in translations. It introduces two separate use-cases:

  - Text-level elements. Localizers may want to use some HTML markup to make the translation correct according to the rules of grammar and spelling. For instance, `<em>` may be used for words borrowed from foreign languages, `<sup>` may be used for ordinals or abbreviations, etc.

  - Functional elements. Developers may want to pass elements as arguments to the translation and create rich language-agnostic UIs. For instance, they may define an `` element which should be placed inline inside of the translation. Each language will need to decide where exactly the `` should go.

#### Text-Level Elements

For the first use-case, _fluent-dom_ will always allow a safe set of text-level elements such as `<em>` and `<strong>`. Their contents will always be converted to pure text (no nesting is allowed) and their attributes will be sanitized using a well-defined list of safe attributes.

    hello = Hello, <em title="It's a wonderful world.">world</em>!

#### Functional Elements

For the latter use-case, developers may put child elements inside of the element which is the target of the translation. These child elements must be annotated with the data-l10n-name attribute. _fluent-dom_ will look for corresponding child elements defined by the translation and clone them into the final translated result. The cloning preserves functional attributes defined in the source (`href`, `class`, `src`) but destroys any event listeners attached to the child element. (We may be able to relax this limitation in the future.) Safe attributes found on the matching child element from the translation will be copied to the resulting child.

    <p data-l10n-id="hello-icon">
        
    </p>

    hello-icon = Hello, !

Fixes #128. Fixes #144.


stasm:
I'm very happy with how this PR improves the code in `overlay.js`. The separation of the two use-cases made the code easier to follow and easier to test. Both of these qualities are important for the sake of the security of the implementation.

One consequence of the new design is that _all_ elements which want to use attributes defined in the source HTML must now have the `data-l10n-name` attribute. This also includes <a>. We'll need to change some existing Fluent strings as follows:

```properties
# BEFORE
update-application-info = Version { $version } <a>What's new</a>
```
```properties
# AFTER
update-application-info = Version { $version } <a data-l10n-name="whatsnew">What's new</a>
```

stasm:
I wonder if we should remove the `a` element from `LOCALIZABLE_ELEMENTS`. Even if it's defined as one of the [text-level elements by the HTML spec](https://www.w3.org/TR/html5/text-level-semantics.html), I don't see much value in letting it roam free in translations without `data-l10n-name` (which would be the only way an `a` element defined by the translation could get the `href` attribute).

stasm:
I chose `data-l10n-name` as the name of the attribute which defines functional argument-elements to translation. It makes the translation a bit more wordy:

    <a data-l10n-name="whatsnew">What's new</a>

I think that's OK given how much value this brings. It's also the simplest solution in terms of teaching it to developers. I'm open to discussing other means to uniquely identify child elements. We could do something similar to what `fluent-react` does, where the name of the prop passed to `<Localized>` becomes the name of the element in the translation:

```html
<p data-l10n-id="update-application-info" data-l10n-args="…">
    <a data-l10n-name="whatsnew" href="…"></a>
</p>
```

```properties
update-application-info = Version { $version } <whatsnew>What's new</whatsnew>
```

An alternative could be to construct the name of the element from the original child element (`a`) and its `data-l10n-name` attribute (`whatsnew`):

```properties
# Custom-element style
update-application-info = Version { $version } <a-whatsnew>What's new</a-whatsnew>
```
```properties
# BEM style
update-application-info = Version { $version } <a--whatsnew>What's new</a--whatsnew>
```
```properties
# I didn't even know dots were allowed as tag names in XML
update-application-info = Version { $version } <a.whatsnew>What's new</a.whatsnew>
```

stasm:
Looking for feedback from @flodolo, @Pike and @zbraniecki. Take your time :)

fluent-dom/src/overlay.js
```diff
@@ -43,152 +43,148 @@ const LOCALIZABLE_ATTRIBUTES = {
...
+ * attribute will be treated as arguments to the translation. If the
+ * translation defines the same children, their attributes and text contents
+ * will be used for translating the matching source child.
+ *
+ * @param   {Element} element
+ * @param   {Object} translation
  * @private
  */
-export default function overlayElement(targetElement, translation) {
+export default function translateElement(element, translation) {
```
stasm: The diff of `overlay.js` isn't very helpful, I'm afraid. I refactored most of the file. I may suggest opening the new version in a new tab and reading it as it is.

fluent-dom/src/overlay.js
```diff
@@ -43,152 +43,148 @@ const LOCALIZABLE_ATTRIBUTES = {
...
+      // We can't currently guarantee that a translation won't remove
+      // sourceChild from the element completely, which could break the app if
+      // it relies on an event handler attached to the sourceChild. Let's make
+      // this limitation explicit for now by breaking the identitiy of the
+      // sourceChild by cloning it. This will destroy all event handlers
+      // attached to sourceChild via addEventListener and via on<name>
+      // properties.
+      const clone = sourceChild.cloneNode(false);
+      return shallowPopulate(clone, childNode);
+    }
```
stasm: Should we add an `else` here with an early return producing `textContent`?

Right now the code moves to `if (isElementLocalizable(childNode))` which means that the following HTML in the translation:

    <em data-l10n-name="mismatched-name">Foo</em>

will produce the following result:

    <em>Foo</em>

If we return early on mismatched names, the result would instead be:

    Foo

fluent-dom/src/overlay.js
```diff
@@ -43,152 +43,148 @@ const LOCALIZABLE_ATTRIBUTES = {
...
+      // the same reference name again.
+      sourceElement.removeChild(sourceChild);
+      // We can't currently guarantee that a translation won't remove
+      // sourceChild from the element completely, which could break the app if
+      // it relies on an event handler attached to the sourceChild. Let's make
+      // this limitation explicit for now by breaking the identitiy of the
+      // sourceChild by cloning it. This will destroy all event handlers
+      // attached to sourceChild via addEventListener and via on<name>
+      // properties.
+      const clone = sourceChild.cloneNode(false);
```
stasm: I'd like to stay explicit in this iteration and try to remove this limitation in April. I think it should be possible to cache child elements in a WeakMap to prevent losing them to a bad translation. However, I'm not sure I fully understand the consequences of caching them like this and I also didn't want to make tests any more complex for now.

zbraniecki:
Great work Stas!

There's a saying that the most important innovations are those that after the fact feel very obvious. This is one of them. Glad to see it coming! :)

After reading through the patch I have three topics that I'd like to see addressed in one way or another:

1. Performance

I'm concerned about the impact on "simple" node translation since if I understand the code correctly it runs through each and every `data-l10n-id` element scanning for attributes to be removed against the whitelist.

That seems like a potential performance regression and I'm tbh quite shaky about the whole concept of sanitization of old attributes. I understand the idea but it doesn't seem to me like Fluent is developing in the direction where localizers can set arbitrary attributes that differ between locales, and in that case, I don't see any value out of this operation.

Additionally, the fast-path C++ API we use in Gecko does not perform this sanitization, which introduces inconsistency.

I'd like to propose that we drop this feature.

There's more in the patch that seems worthy of perf comparison (of course understanding that the code does more), but this one item stands out as it affects simple translation of childless elements.

2) Identity preservation

I'd like to at least test an approach which would allow us to preserve the identity of the element with this patch. I'm ok not landing it here, but I'd like to verify the hypothesis that we can move forward with the idea Stas has without having to touch the client code (ftl,xul,js files).

3) `data-l10n-name`

I'm torn here.

I understand the arguments for it, and I understand the logic behind it. I also like how it avoids the problem of matching.
But it feels like a significant regression in readability compared to the original approach.

I'd like to discuss if there is a path forward that would allow users to skip it in common cases.

Example of the readability difference from a patch I'm working on right now:

```xml
<label class="fxMobilePromo" data-l10n-id="sync-signedout-mobile-promo">
  <image class="androidLink" data-l10n-name="androidImage" />
  <label id="fxaMobilePromo-android" class="text-link" data-l10n-name="androidLink" />
  <image class="iOSLink" data-l10n-name="iosImage" />
  <label id="fxaMobilePromo-ios" class="text-link" data-l10n-name="iosLink" />
</label>
```

```properties
sync-signedout-mobile-promo =
    Download Firefox for <image data-l10n-name="androidImage" /> <label data-l10n-name="androidLink">Android</label> or <image data-l10n-name="iosImage" /> <label data-l10n-name="iosLink">iOS</label> to sync with your mobile device.
```
vs.

```xml
<label class="fxMobilePromo" data-l10n-id="sync-signedout-mobile-promo">
  <image class="androidLink"/>
  <label id="fxaMobilePromo-android" class="text-link"/>
  <image class="iOSLink"/>
  <label id="fxaMobilePromo-ios" class="text-link"/>
</label>
```

```properties
sync-signedout-mobile-promo =
    Download Firefox for <image/> <label>Android</label> or <image/> <label>iOS</label>
    to sync with your mobile device.
```

```xml
<label class="fxMobilePromo" data-l10n-id="sync-signedout-mobile-promo">
  <image class="androidLink" data-l10n-name="androidImage" />
  <label id="fxaMobilePromo-android" class="text-link" data-l10n-name="androidLink" />
  <image class="iOSLink" data-l10n-name="iosImage" />
  <label id="fxaMobilePromo-ios" class="text-link" data-l10n-name="iosLink" />
</label>
```

```properties
sync-signedout-mobile-promo =
    Download Firefox for <image data-l10n-name="androidImage" /> <label data-l10n-name="androidLink">Android</label> or <image data-l10n-name="iosImage" /> <label data-l10n-name="iosLink">iOS</label> to sync with your mobile device.
```

Stas suggested several alternatives to using `data-l10n-name`, but I'm not sure how user friendly they would be.

Using names as tags would give us:

```properties
sync-signedout-mobile-promo =
    Download Firefox for <androidImage /> <androidLink>Android</androidLink> or <iosImage /> <iosLink>iOS</iosLabel> to sync with your mobile device.
```

which I think brings back more readability, but at the cost of introducing unfamiliar tag names. I'd expect most localizers to understand `<image/>` or `<a></a>` but not necessarily `<iosLink></iosLink>`.

But maybe that is some way out?

I'm wondering if there's a way for us to design it so that in this step we can bring the functionality, and in the future lift restrictions to allow "easy things easy, complex things possible" kind of approach where if you just want to put an `` or an `<a></a>` you can do it without names, but once you have multiple you can somehow add distinguishers.

stasm:
> I'm concerned about the impact on "simple" node translation since if I understand the code correctly it runs through each and every data-l10n-id element scanning for attributes to be removed against the whitelist.

This also happens in the current implementation: https://github.com/projectfluent/fluent.js/blob/83b1d1df1e684ef9223a8d898bc6312de222d538/fluent-dom/src/overlay.js#L77-L83

> Additionally, the fast-path C++ API we use in Gecko does not perform this sanitization, which introduces inconsistency.

Yes, it's an inconsistency introduced by the C++ API and we should fix it. In the C++ API :) The feature of removing leaked attributes is important to me from the idempotence stand point. Ideally, translation would always create the same result, given the source HTML and the FTL file. Leaking attributes between retranslations breaks the idempotence and makes it harder to test the output.

Let's discuss the proposal to drop this feature in a separate issue. Can you file it please?

> I'd like to at least test an approach which would allow us to preserve the identity of the element with this patch. I'm ok not landing it here, but I'd like to verify the hypothesis that we can move forward with the idea Stas has without having to touch the client code (ftl,xul,js files).

The current implementation doesn't keep the identity of overlaid elements either, so this PR doesn't regress the status quo.

>  I'd expect most localizers to understand <image/> or <a></a> but not necessarily <iosLink></iosLink>.

I actually think this is a plus: it would leave the well-known names of elements for localizers to use as per the first use-case (text-level elements). And the functional elements defined by developer would look exotic enough to hint that they should not be tampered with :)

zbraniecki:
> The current implementation doesn't keep the identity of overlaid elements either, so this PR doesn't regress the status quo.

I did not claim it regresses. I said that I believe we should verify that the hypothesis that we can use WeakMap to add it later without changes to how the API is used is real.

> I actually think this is a plus: it would leave the well-known names of elements for localizers to use as per the first use-case (text-level elements). And the functional elements defined by developer would look exotic enough to hint that they should not be tampered with :)

Maaybe... I have to think more on this.

stasm:
> I did not claim it regresses. I said that I believe we should verify that the hypothesis that we can use WeakMap to add it later without changes to how the API is used is real.

You're right, sorry. I assume we'd _always_ want to verify hypotheses that we put forwad, so I'm not sure what you have in mind when you mention it here. Do you want to block this PR on the verification of the hypothesis?

zbraniecki:
Added: https://github.com/projectfluent/fluent.js/issues/170

Pike:
Yay, named arguments. On the design aspects:

I strongly feel that we should require named arguments in all cases.

I also think we shouldn't let react break the DOM. There's real semantics behind things being a button or not, or a link. We should persist that, and avoid magic element names.

Thus, my recommendation is to use a `data-l10n-name` attribute on any parameter node.

stasm:
> We should remove all cloneNode calls. They're all freshly generated nodes form template.innerHTML, let's use them as is.

I assume you mean the `cloneNode` in the case of a `TEXT_NODE`. Good point, let's remove it. The other one, cloning the `sourceChild`, should stay however. At least for now: I want to break the identity of elements in this PR. We don't need it for putting `<image>` elements into the translations.

> isElementLocalizable(childNode) is only ever called for HTML elements, maybe we should just make that explicit? 

Sorry, I don't follow. Do you mean renaming the function name?

> Also, as I'm on it, is Localizable bad grammar for what we intend here? Also, LOCALIZABLE_ATTRIBUTES sounds similar, but really isn't in its functionality? More like a data-l10n-attrs-whitelist?

Should we go back to using `ALLOWED_ELEMENTS` and `isElementAllowed`?

> I'm all for removing <a> from allowed localizer content. We should probably also remove the <ruby>-related elements, as they don't make any sense as long as we don't support nested text-level-semantics. 

Good point.

> Or maybe we should just do that? Now that named-arguments and text-level-semantics are properly distinguished, that wouldn't be too bad?

I'm against doing it in this PR. The goal of this PR is to allow the use case of putting two `<image>` elements in the translation.

stasm:
> I assume you mean the cloneNode in the case of a TEXT_NODE. Good point, let's remove it. 

Actually, sorry, we can't do this without changing more code. If we return the exact same node, it's then appended to the `DocumentFragment` and thus immediately _removed_ from the constructed `<template>` which breaks the `for…of` iteration. I'll document it in a comment.

fluent-dom/src/overlay.js
```diff
@@ -43,152 +46,153 @@ const LOCALIZABLE_ATTRIBUTES = {
...
-      targetElement.removeAttribute(attr.name);
-    }
-  }
+/**
+ * Replace child nodes of an element with child nodes of another element.
+ *
+ * The contents of the target element will be cleared and fully replaced with
+ * sanitized contents of the source element.
+ *
+ * @param   {Element} fromElement - The source of child nodes to overlay.
```
Pike: Nit-pickin', this is actually a `DocumentFragment` instead of an `Element`.

fluent-dom/src/overlay.js
```diff
@@ -263,19 +267,15 @@ function isAttrNameLocalizable(name, element, explicitlyAllowed = null) {
...
  */
-function shiftNamedElement(element, localName) {
-  for (const child of element.children) {
-    if (child.localName === localName) {
-      element.removeChild(child);
-      return child;
-    }
-  }
-  return null;
+function shallowPopulate(element, fromElement) {
```
Pike: Can you revert the order of the args here to be consistent with the other methods? I had trouble following assertions of what's safe and what's not because this methods flips that going in and going out to overlayAttributes.

PR #172, opened 2018-03-19
BaseNode.equals support, and other ast goodness from python-fluent

Pike:
This is basically a port of
https://github.com/projectfluent/python-fluent/blob/90301381e9b4d56b317bd6f15adf38936569f023/fluent/syntax/ast.py#L76-L120

Creating a PR early on to ensure that node 6 and 8 pass on the js features. For example, the use of `Object.keys`, which is rumored to be OK per http://node.green/#ES2015-misc-Object-static-methods-accept-primitives-Object-keys.

### DIFF
--- a/fluent-syntax/src/ast.js
+++ b/fluent-syntax/src/ast.js
@@ -7,6 +7,67 @@
  */
 class BaseNode {
   constructor() {}
+
+  equals(other, ignoredFields = ["span"]) {
+    const thisKeys = new Set(Object.keys(this));
+    const otherKeys = new Set(Object.keys(other));
+    if (ignoredFields) {
+      for (const fieldName of ignoredFields) {
+        thisKeys.delete(fieldName);
+        otherKeys.delete(fieldName);
+      }
+    }
+    if (thisKeys.size !== otherKeys.size) {
+      return false;
+    }
+    for (const fieldName of thisKeys) {
+      if (!otherKeys.has(fieldName)) {
+        return false;
+      }
+      const thisVal = this[fieldName];
+      const otherVal = other[fieldName];
+      if (typeof thisVal !== typeof otherVal) {
+        return false;
+      }
+      if (thisVal instanceof Array) {
+        if (thisVal.length !== otherVal.length) {
+          return false;
+        }
+        // Sort elements of order-agnostic fields to ensure the
+        // comparison is order-agnostic as well. Annotations should be
+        // here too but they don't have sorting keys.
+        if (["attributes", "variants"].indexOf(fieldName) >= 0) {
+          thisVal.sort(sorting_key_compare);
+          otherVal.sort(sorting_key_compare);
+        }
+        for (let i = 0, ii = thisVal.length; i < ii; ++i) {
+          if (!scalars_equal(thisVal[i], otherVal[i], ignoredFields)) {
+            return false;
+          }
+        }
+      } else if (!scalars_equal(thisVal, otherVal, ignoredFields)) {
+        return false;
+      }
+    }
+    return true;
+  }
+}
+
+function scalars_equal(thisVal, otherVal, ignoredFields) {
+  if (thisVal instanceof BaseNode) {
+    return thisVal.equals(otherVal, ignoredFields);
+  }
+  return thisVal === otherVal;
+}
+
+function sorting_key_compare(left, right) {
+  if (left.sorting_key < right.sorting_key) {
+    return -1;
+  }
+  if (left.sorting_key === right.sorting_key) {
+    return 0;
+  }
+  return 1;
 }
 
 /*
@@ -188,6 +249,10 @@ export class Attribute extends SyntaxNode {
     this.id = id;
     this.value = value;
   }
+
+  get sorting_key() {
+    return this.id.name;
+  }
 }
 
 export class Variant extends SyntaxNode {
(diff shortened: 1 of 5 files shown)


Pike:
stas, I think this is ready for review.

I gave all those things from python-fluent a go (w/out traverse).

Tests are heavily inspired by the python tests, too.

Questions I know of: `ignoreFields` vs `ignore_fields`. And is `Object.create` the right way to instantiate a new object of the same type, given that the `kwargs` trick in python doesn't work in js? (I don't think it does).

fluent-syntax/src/visitor.js
```diff
@@ -0,0 +1,60 @@
...
+  visit(node) {
+    if (Array.isArray(node)) {
+      node.forEach(child => this.visit(child));
+      return;
+    }
+    if (!(node instanceof BaseNode)) {
+      return;
+    }
+    const nodename = node.type;
+    const visit = this[`visit_${nodename}`] || this.generic_visit;
```
stasm: Nit: I'd just interpolate `node.type` here, without the interim `nodename` variable. It reads better, too.

fluent-syntax/src/ast.js
```diff
@@ -5,8 +5,86 @@
...
+      if (thisVal instanceof Array) {
+        if (thisVal.length !== otherVal.length) {
+          return false;
+        }
+        // Sort elements of order-agnostic fields to ensure the
+        // comparison is order-agnostic as well. Annotations should be
+        // here too but they don't have sorting keys.
+        if (["attributes", "variants"].includes(fieldName)) {
+          thisVal.sort(sortingKeyCompare);
+          otherVal.sort(sortingKeyCompare);
```
stasm: Should we make copies of both arrays here, just in case someone tries to serialize the AST (or work with it in some other fashion) after running `equals` on it?

fluent-syntax/src/visitor.js
```diff
@@ -0,0 +1,58 @@
...
+export class Visitor {
+  visit(node) {
+    if (Array.isArray(node)) {
+      node.forEach(child => this.visit(child));
+      return;
+    }
+    if (!(node instanceof BaseNode)) {
+      return;
+    }
+    const visit = this[`visit_${node.type}`] || this.genericVisit;
```
stasm: After some more thought, as a consumer of this API I'd find it strange that the underscore is required here. Plus, it might trigger linter warnings, e.g. https://eslint.org/docs/rules/camelcase. I'm not going to block on this but I'd prefer `visit${node.type}`, e.g. `visitTextElement`.

fluent-syntax/test/ast_test.js
```diff
@@ -0,0 +1,78 @@
...
+            }
+      `);
+      const thisNode = thisRes.body[0];
+      const otherNode = otherRes.body[0];
+      assert.strictEqual(thisNode.equals(otherNode), false);
+      assert.strictEqual(thisNode.equals(otherNode, ['span', 'comment']), true);
+      assert.strictEqual(thisNode.value.equals(otherNode.value), true);
+      assert.strictEqual(thisNode.value.equals(otherNode.value, []), false);
+      assert.strictEqual(thisRes.equals(thisRes.clone(), []), true);
+      assert.notStrictEqual(thisRes, thisRes.clone());
```
stasm: Let's add a test which verifies that the original nodes have their variants in unchanged order after these tests run, please.

fluent-syntax/src/ast.js
Pike: I'd rather avoid the cost for that.

stasm: Do I understand it correctly, that not doing it introduces a bug: the AST is mutated when `equals` is called?

Pike: Thinking about this, we should just not make nodes equal if the order changes.

I filed https://github.com/projectfluent/python-fluent/issues/111 on this.

fluent-syntax/test/ast_test.js
Pike: I'm dropping the ordering completely, per comment above, and update the tests accordingly. Let me know if there's more to do here.

Pike:
There's still an open question I have on `sortingKey`. It's really useful in tooling as a canonical representation of a VariantKey, in compare-locales, and also in Pontoon for example.

stasm:
Thanks for addressing my comments, @Pike! This looks good to land.

Regarding the question about `sortingKey`: `serializedKey` which we discussed last night would work for me. I think a helpful excerise is to imagine what this field would be for variants with `StringLiterals` as keys (https://github.com/projectfluent/fluent/issues/90). Would we want to serialize them wrapped in double quotes, to differentiate them from keys which are `Identifiers`? Or should `[foo]` and `["foo"]` serialize to the same thing?

I'm a bit concerned that having this field is mixing concerns. The use-cases present in Pontoon should really just use `FluentSerializer` or a custom serialization layer. The use-case of finding duplicate variants could also be solved with a one-line function in `compare-locales`. But, since this is an extra API of `python-fluent`'s AST rather than part of the reference AST, I see the tooling value here and I don't want to block.

Pike:
I've taking your comments and mulled over them, and then ended up removing the `sorting_key` in python. I replaced the use in `compare-locales` in https://phabricator.services.mozilla.com/D23813, which had a couple of arguments to change the way that code worked to begin with. For variant keys, I went for serialization. For attributes, I just hard-coded the AST structure accessor.

PR #177, opened 2018-04-06
Report an error when attempting to override an existing message or term.

zbraniecki:
It's a pure developer productivity improvement. In our experience there has never been a case where overwriting an ID in the same context was intentional and useful, but it can be a really hard to debug and can lead to a very entangled code if someone accidentally starts depending on it.

Let's cut it short early. No overrides.

### DIFF
--- a/fluent/src/context.js
+++ b/fluent/src/context.js
@@ -113,8 +113,16 @@ export class MessageContext {
       if (id.startsWith("-")) {
         // Identifiers starting with a dash (-) define terms. Terms are private
         // and cannot be retrieved from MessageContext.
+        if (this._terms.has(id)) {
+          errors.push(`Attempt to override an existing term: "${id}"`);
+          continue;
+        }
         this._terms.set(id, entries[id]);
       } else {
+        if (this._messages.has(id)) {
+          errors.push(`Attempt to override an existing message: "${id}"`);
+          continue;
+        }
         this._messages.set(id, entries[id]);
       }
     }
--- a/fluent/test/context_test.js
+++ b/fluent/test/context_test.js
@@ -53,16 +53,19 @@ suite('Context', function() {
     });
 
 
-    test('overwrites existing messages if the ids are the same', function() {
-      ctx.addMessages(ftl`
+    test('does not overwrite existing messages if the ids are the same', function() {
+      const errors = ctx.addMessages(ftl`
         foo = New Foo
       `);
+
+      // Attempt to overwrite error reported
+      assert.equal(errors.length, 1);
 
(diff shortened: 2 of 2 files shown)


zbraniecki:
or, should we throw (to avoid having to specify `console`?)

fluent/src/context.js
```diff
@@ -113,8 +113,20 @@ export class MessageContext {
       if (id.startsWith("-")) {
         // Identifiers starting with a dash (-) define terms. Terms are private
         // and cannot be retrieved from MessageContext.
+        if (this._terms.has(id)) {
+          if (typeof console !== "undefined") {
```
stasm: In which environments can we expect `console` to be undefined?

stasm:
> or, should we throw (to avoid having to specify console?)

`addMessages` doesn't throw currently and I think it's the right design for it. We've put a lot of work to have good error recovery in the parser and throwing on duplicate messages seems to be at odds with the defensive nature of this API.

fluent/src/context.js
zbraniecki: I don't think there's any. Can we add it to eslint to not throw and assume `console` is available? :)

stasm: I throws to protect against `console.logs` sprinkled across the code after some debugging. `// eslint-disable-line no-console` should do the trick.

stasm: If `console` is available in all envs, then what's the purpose of the `if typeof` check? Can we remove it?

fluent/src/context.js
```diff
@@ -113,8 +115,22 @@ export class MessageContext {
...
+            // eslint-disable-next-line no-console
+            console.error(`Attempt to override an existing l10n-id: "${id}"`);
```
stasm: `…an existing term…`

fluent/src/context.js
```diff
@@ -113,8 +115,22 @@ export class MessageContext {
...
+          }
+          continue;
+        }
         this._terms.set(id, entries[id]);
       } else {
+        if (this._messages.has(id)) {
+          if (typeof console !== "undefined") {
+            // eslint-disable-next-line no-console
+            console.error(`Attempt to override an existing l10n-id: "${id}"`);
```
stasm: `…an existing message…`

stasm:
Actually, I think we should do it differently. Let's _return_ an array of errors from `addMessages` and let the user code report them as they see fit. @zbraniecki what do you think?

zbraniecki:
Hmm, no strong opinion. I'm fine with either.

PR #179, opened 2018-04-10
Validate locales passed to MessageContext constructor if Intl.getCanonicalLocales is available

zbraniecki:


fluent/src/context.js
```diff
@@ -48,7 +50,10 @@ export class MessageContext {
    * @returns {MessageContext}
    */
   constructor(locales, { functions = {}, useIsolating = true } = {}) {
-    this.locales = Array.isArray(locales) ? locales : [locales];
+    // eslint-disable-next-line no-nested-ternary
+    this.locales = Intl.getCanonicalLocales ?
+      Intl.getCanonicalLocales(locales) :
+      Array.isArray(locales) ? locales : [locales];
```
stasm: I think `eslint` was right here :) I'd reformat this as:

```js
if (Intl.getCanonicalLocales) {
    this.locales = …
} else {
    this.locales = …
}
```

If you want to keep the nested ternary, I'd suggest the following formatting:

```js
this.locales = Intl.getCanonicalLocales
  ? Intl.getCanonicalLocales(locales)
  : Array.isArray(locales)
    ? locales
    : [locales];
```

Another alternative would be to add a dumb polyfill of `getCanonicalLocales` in `fluent/src/index` which would only make the `isArray` check. You'd then only use `this.locales = Intl.getCanonicalLocales()` here.

stasm:
@zbraniecki How hard would it be to polyfill the correct version of `getCanonicalLocales`? Can we make it part of `fluent-langneg` (to be renamed `fluent-locale`) or `fluent-intl-polyfill` and make `fluent` depend on it? Or at least strongly recommend using it outside of Gecko.

My worry with the approach taken in this PR is that `MessageContext` might start throwing for some users depending on their browser support for `Intl` methods. We can't really solve this by versioning of the `fluent` API, either.

zbraniecki:
> How hard would it be to polyfill the correct version of getCanonicalLocales?

Seems like this polyfill may work - https://www.npmjs.com/package/ietf-language-tag-regex ?

Would you like to vendor it in, or copy (in which case we may need to cross-check the licensing)

PR #191, opened 2018-04-26
Separate out CachedIterable and CachedAsyncIterable, and add a param to touchNext

zbraniecki:


fluent/src/cached_iterable.js
```diff
@@ -4,17 +4,15 @@
  * It can be used to iterate over an iterable many times without depleting the
  * iterable.
  */
-export default class CachedIterable {
+export class CachedIterable {
```
stasm: Please rename this to `CachedSyncIterable` to make the distinction explicit. We already have a precedent for using `Sync` in API names in form of `mapContextSync`, so this will be consistent with it.

Issue #126, opened 2018-01-12
Upgrade to Babel 7


stasm:
Babel 7 is still in beta but very much usable. Let's use this issue to track what changes are needed on our side when we upgrade.

Improvements:

  - [x] Remove the `transform-builtin-extend` plugin. (https://github.com/babel/babel/pull/7020)
  - [x] OK to use destructuring in `for await` loops. (https://github.com/babel/babel/pull/5932)
  - [x] Maybe drop `stage-3` in favor of [`@babel/plugin-proposal-async-generator-functions`](https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-async-generator-functions)?

Open questions:

  - [x] Will the logic of reading `.babelrc` files change? (https://github.com/babel/babel/issues/6766)

PR #193, opened 2018-05-07
Only use compat dependencies in fluent-react/compat

### LINKED ISSUE #192: fluent-react shouldn't force usage of fluent/compat
As seen here:

https://github.com/projectfluent/fluent.js/blob/b44b80205ae1dca0ed24e34d2e96842c4ee93550/fluent-react/src/localization.js#L1

`fluent/compat` adds a polyfill for async generator functions, thus requiring projects to use `babel-polyfill` (or something like it), even if their browser supports async generators natively.

stasm:
Fix #192.

### DIFF
--- a/babel_config.js
+++ b/babel_config.js
@@ -1,0 +1,21 @@
+export default {
+  babelrc: false,
+  presets: [
+    ["@babel/preset-env", {
+      // Cf. https://github.com/rollup/rollup-plugin-babel#modules
+      modules: false,
+      targets: {
+        browsers: [
+          ">1%",
+          "last 4 versions",
+          "Firefox ESR",
+          "not ie < 9"
+        ]
+      }
+    }]
+  ],
+  plugins: [
+    "@babel/plugin-proposal-async-generator-functions",
+    "@babel/plugin-proposal-object-rest-spread"
+  ]
+};
--- a/compat_config.js
+++ b/compat_config.js
@@ -1,30 +1,11 @@
 import babel from 'rollup-plugin-babel';
+import babelConfig from './babel_config';
 
 export default {
   output: {
     format: 'umd'
   },
   plugins: [
-    babel({
-      'babelrc': false,
-      'presets': [
-        ['@babel/preset-env', {
-          // Cf. https://github.com/rollup/rollup-plugin-babel#modules
-          'modules': false,
-          'targets': {
-            'browsers': [
-              '>1%',
-              'last 4 versions',
-              'Firefox ESR',
-              'not ie < 9'
-            ]
-          }
-        }]
-      ],
-      plugins: [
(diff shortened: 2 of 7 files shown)


stasm:
@jimporter I was able to use `babel-plugin-transform-rename-import` to fix the issue you reported in #192. Thanks for the pointer! I tested it locally with a simple React app but I wasn't able to verify that it fixes https://github.com/mozilla-lockbox/lockbox-extension/pull/696. Would you mind trying it out by cloning this branch, `cd fluent-react; npm link` and then `npm link fluent-react` in `lockbox-extension`?

jimporter:
@stasm It works, though I couldn't use `npm link` because that made Babel unhappy with the stage-3 features used in fluent (this might be an issue in our build config, but Webpack+Babel is really brittle in this area, so I'm going to try to avoid messing with it too much). Manually copying fluent-react to my `node_modules` works just fine though.

stasm:
I'm having trouble making the tests pass because now `localization.js` imports from `fluent` rather than `fluent/compat` and the JSX transpilation of `fluent-react/test` fails. I tried using the `syntax-async-generators` plugin but to no avail.

stasm:
I think this is a bug in Babel 6. `babel-register` [ignores `node_modules` by default](https://babeljs.io/docs/usage/babel-register/#ignores-node_modules-by-default) and I can't seem to tell it to also transpile the `node_modules/fluent/fluent.js` dependency. It looks like this issue might be [fixed in Babel 7](https://github.com/babel/babel/pull/5487). #126 is about upgrading to Babel 7, but it's a more involved task and I'd also prefer to wait for the stable release rather than use the betas.

I considered dropping `babel-register` altogether. We currenly use it for two purposes:

  - Transpile `import` statements in tests so that mocha can run them.
  - Transpile async generators in `fluent/src` and in `fluent/test` so that, again, mocha can run them.

We could use [esm](https://github.com/standard-things/esm) for the first goal. For the second one, a tempting way forward would be to switch to Node 10 for development (i.e. running tests and building the compat versions, which would continue to work in older Node versions). Node 10 came out 2 weeks ago, so it might be a bit too early and I'm also not sure what _other_ issues we might run into if we switch.

Perhaps I can simply hook the `rename-import` plugins into mocha as well?

stasm:
After more research I bit the bullet and upgraded to Babel 7 in #200. This PR is now green. I'll test a bit more and land today.

jimporter:
@stasm That's pretty much what I did for Lockbox desktop too. Glad to see it works here as well!

stasm:
Alright, I was finally able to test this locally. Yay, it works :)

stasm:
I published [`fluent-react` 0.7.0](https://github.com/projectfluent/fluent.js/releases/tag/fluent-react%400.7.0) today. It includes this change.

PR #196, opened 2018-05-14
Re-render withLocalization-Components when l10n changes

Gregoor:
I took a stab it. Happy to make changes, if I went off-convention somewhere 🙃 

Also tested it in Common Voice and it indeed fixes https://github.com/mozilla/voice-web/issues/986.

### DIFF
--- a/fluent-react/src/with_localization.js
+++ b/fluent-react/src/with_localization.js
@@ -9,6 +9,31 @@ export default function withLocalization(Inner) {
       this.getString = this.getString.bind(this);
     }
 
+    componentDidMount() {
+      const { l10n } = this.context;
+
+      if (l10n) {
+        l10n.subscribe(this);
+      }
+    }
+
+    componentWillUnmount() {
+      const { l10n } = this.context;
+
+      if (l10n) {
+        l10n.unsubscribe(this);
+      }
+    }
+
+    /*
+     * Rerender this component in a new language.
+     */
+    relocalize() {
+      // When the `ReactLocalization`'s fallback chain changes, update the
+      // component.
+      this.forceUpdate();
+    }
+
     /*
      * Find a translation by `id` and format it to a string using `args`.
      */
@@ -25,7 +50,7 @@ export default function withLocalization(Inner) {
     render() {
       return createElement(
         Inner,
-        Object.assign({ getString: this.getString }, this.props)
+        Object.assign({ getString: (...args) => this.getString(...args) }, this.props)
       );
     }
   }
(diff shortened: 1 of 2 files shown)


stasm:
Yay, thanks! I'll review this tomorrow.

Gregoor:
Great thanks :) Looks like the linter doesn't like my code, but it talks about a line that doesn't exist. Plus locally `make lint` doesn't report anything to me. Hmm

fluent-react/src/with_localization.js
```diff
@@ -25,7 +50,7 @@ export default function withLocalization(Inner) {
     render() {
       return createElement(
         Inner,
-        Object.assign({ getString: this.getString }, this.props)
+        Object.assign({ getString: (...args) => this.getString(...args) }, this.props)
```
stasm: I think the linter complains about this line. Not that this change shouldn't be needed because `this.getString` is bound to `this` in line 9. That said, I'd much rather get rid of that `bind` and the whole `constructor` and fix the line length here :)

Gregoor: Oh, that line does exist, I was on the wrong branch, sorry 🙈 I pushed a new commit, which also added a comment and removes the constructor.

fluent-react/src/with_localization.js
```diff
@@ -50,7 +45,11 @@ export default function withLocalization(Inner) {
     render() {
       return createElement(
         Inner,
-        Object.assign({ getString: (...args) => this.getString(...args) }, this.props)
+        Object.assign(
+          // getString needs to be re-bound on updates to trigger a re-render of Inner
```
Gregoor: line length fail no 2

stasm:
Sweet, thanks! Can I squash this while merging?

Gregoor:
Sure! Always feel free to squash my commits 😁

stasm:
Thanks again, @Gregoor! This fixes #194. I'll publish a new version of `fluent-react` tomorrow. I'm still hoping I can find a solution to #192 to include in the same release.

stasm:
I published [`fluent-react` 0.7.0](https://github.com/projectfluent/fluent.js/releases/tag/fluent-react%400.7.0) today.

Gregoor:
Awesome, thanks for getting a new release out 🙂

stasm:
Thanks again for the PR and the patience. I fixed a few build system-related issues in this release (hence 0.7.0 rather than 0.6.2) and wanted to test them thoroughly before publishing.

stasm:
(If you're using the `/compat` builds, however, you shouldn't notice any changes.)

Gregoor:
Sure, happy to have had an excuse to contribute to fluent 🙃

PR #195, opened 2018-05-09
Put FluentDOM in line with Gecko's nsIDocument binding changes.

zbraniecki:
This is a straight port of Dave's work from https://bugzilla.mozilla.org/show_bug.cgi?id=1455649

I think that both bring value straight to FluentDOM, and I'm happy to see how little fine-tuning it needs.

Let me know if you're ok with them Stas.

PR #197, opened 2018-05-14
Allow adding and removing resource Ids from a Localization.

zbraniecki:
Separated out of #195

### DIFF
--- a/fluent-dom/src/localization.js
+++ b/fluent-dom/src/localization.js
@@ -26,12 +26,12 @@ export default class Localization {
 
   addResourceIds(resourceIds) {
     this.resourceIds.push(...resourceIds);
-    this.onLanguageChange();
+    this.onChange();
   }
 
   removeResourceIds(resourceIds) {
     this.resourceIds = this.resourceIds.filter(r => !resourceIds.includes(r));
-    this.onLanguageChange();
+    this.onChange();
   }
 
   /**
@@ -144,14 +144,14 @@ export default class Localization {
   }
 
   handleEvent() {
-    this.onLanguageChange();
+    this.onChange();
   }
 
   /**
    * This method should be called when there's a reason to believe
    * that language negotiation or available resources changed.
    */
-  onLanguageChange() {
+  onChange() {
     this.ctxs =
       new CachedAsyncIterable(this.generateMessages(this.resourceIds));
   }
--- a/fluent-dom/src/localization.js
+++ b/fluent-dom/src/localization.js
@@ -24,6 +24,16 @@ export default class Localization {
       new CachedAsyncIterable(this.generateMessages(this.resourceIds));
   }
 
+  addResourceIds(resourceIds) {
+    this.resourceIds.push(...resourceIds);
+    this.onLanguageChange();
+  }
+
+  removeResourceIds(resourceIds) {
+    this.resourceIds = this.resourceIds.filter(r => !resourceIds.includes(r));
+    this.onLanguageChange();
+  }
+
   /**
    * Format translations and handle fallback if needed.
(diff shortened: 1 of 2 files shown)


fluent-dom/src/localization.js
```diff
@@ -24,6 +24,16 @@ export default class Localization {
       new CachedAsyncIterable(this.generateMessages(this.resourceIds));
   }
 
+  addResourceIds(resourceIds) {
+    this.resourceIds.push(...resourceIds);
```
stasm: I'm a bit worried about non-deterministic resource injection by web components, depending on the order in which they load. The order of resource ids in Localization is significant because it governs how duplicate translations are handled. The code which performs the injection might want to position the `<link>` in a precisely defined place. Consequently, the code listening for injections might want to add the resource ids at the desired position, too.

Should we consider changing this to something like the following?

```js
insertResourceIds(resourceIds, offset = this.resourceIds.length) {
  this.resourceIds = [
    ...this.resourceIds.slice(0, offset),
    ...resourceIds,
    ...this.resourceIds.slice(offset)
  ];
  // Or simply:
  // this.resourceIds.splice(offset, 0, ...resourceIds);
}

File: fluent-dom/src/localization.js [hunk 2]
```diff
@@ -24,6 +24,16 @@ export default class Localization {
...
+    this.onLanguageChange();
```
stasm: Now that `onLanguageChange` is called also for resource ids changes, should we rename it to something more general?

fluent-dom/src/localization.js
zbraniecki: >  The order of resource ids in Localization is significant because it governs how duplicate translations are handled.

Duplicate IDs throw - https://github.com/projectfluent/fluent.js/blob/master/fluent/src/context.js#L123

fluent-dom/src/localization.js
zbraniecki: Yeah, we probably want to rename it. Any preference for the new name? :)

fluent-dom/src/localization.js
zbraniecki: I would hope that the duplicate ID throwing is sufficient to make the order irrelevant. Just the composition of resources matters. Is that ok with you?

stasm: Right, but still, the first one wins.

fluent-dom/src/localization.js
stasm: Not really. Is `onChange` too generic? :)

fluent-dom/src/localization.js
stasm: It's not really throwing, is it? It logs the error and continues.

zbraniecki: True, should it be throwing instead? I assume that everything declared in the `<head/>` or `<linkset/>` will come first, and everything later will be just added.

I'm ok with throwing as I don't think we should allow for a state where two resources loaded together register the same ID.

Would you be ok with that?

stasm: Where would you like to throw? In `MessageContext.addMessages` or in `Localization`? In any case, I don't think a duplicate translation should break the UI.

Two web components with the same translation id and injecting resources into `document`'s `Localization` might race against each other. I'd prefer to avoid non-determinism as much as we can. For instance, we could report the error as we do right now _and_ remove the previously added message. (Both actions would happen in `MessageContext`.) 

I realize now that inserting in a specific position doesn't really help here, does it? In which case I think it's OK to keep `addResourceIds` as it's implemented in the PR, and discuss solutions to non-deterministic resource injection in a separate issue.

PR #198, opened 2018-05-14
Localization.formatValues should accept an array of object.

zbraniecki:
Separated out of #195


fluent-dom/src/localization.js
```diff
@@ -249,17 +249,17 @@ function keysFromContext(method, ctx, keys, translations) {
   const messageErrors = [];
   const missingIds = new Set();
 
-  keys.forEach((key, i) => {
+  keys.forEach(({id, args = undefined}, i) => {
```
stasm: Nit: `undefined` is the default value for missing parameters anyways, so it's not needed here.

PR #203, opened 2018-05-18
Remove support for Node.js 6.9

### LINKED ISSUE #201: Remove Node 6.9 support from fluent-syntax
We relaxed the engine requirement for `fluent-syntax` in #164. it shipped in `fluent-syntax` 0.6.5+ and 0.7.0. As of April 30, 2018, [Node 6 is no longer in _Active_ LTS cycle](https://github.com/nodejs/Release).

Starting from the yet-to-be-released `fluent-syntax` 0.8.0 I'd like to remove the support for Node 6.9.

@EnTeQuAk, have you started phasing out Node 6.9 support on your side?

stasm:
Fix #201.


stasm:
I'm not very happy with how we must use `babel-polyfill` for tests, but I haven't been able to drop this requirement without investing much more time. I don't think it's worth it.

As they are right now, our tests test the business logic of each package. They do not test whether packages are runnable under a certain version of Node.

PR #213, opened 2018-06-01
Add transformation options to MessageContext

zbraniecki:
Based on the work in https://bugzilla.mozilla.org/show_bug.cgi?id=1450781 this adds the `transform` option to `MessageContext` which can be used to provide pseudolocalization capabilities.

This may be replaced later with internal pseudolocalization, once we gain more experience with testing this in Firefox.

### DIFF
--- a/fluent/src/resolver.js
+++ b/fluent/src/resolver.js
@@ -257,7 +257,10 @@ function SelectExpression(env, {exp, vars, def}) {
 function Type(env, expr) {
   // A fast-path for strings which are the most common case, and for
   // `FluentNone` which doesn't require any additional logic.
-  if (typeof expr === "string" || expr instanceof FluentNone) {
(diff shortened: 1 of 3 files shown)


fluent/src/resolver.js
```diff
@@ -257,7 +257,9 @@ function SelectExpression(env, {exp, vars, def}) {
 function Type(env, expr) {
   // A fast-path for strings which are the most common case, and for
   // `FluentNone` which doesn't require any additional logic.
-  if (typeof expr === "string" || expr instanceof FluentNone) {
+  if (typeof expr === "string") {
+    return env.ctx._transform(expr);
+  } else if (expr instanceof FluentNone) {
```
stasm: No need for the `else` here, let's make this two early-returning `ifs`, please.

Issue #206, opened 2018-05-23
fluent-syntax: Implement Syntax 0.6


stasm:
A tracking issue for implementing [Syntax 0.6](https://github.com/projectfluent/fluent/projects/2) in `fluent-syntax`.

- [x] Port fixtures from the reference parser. (#239)
- [x] Named arguments to `CallExpression` must now follow all positional ones. (#215)
- [x] Forbid duplicate named arguments to `CallExpressions`. (#215)
- [x] Add the `TermReference` expression. (#226)
- [x] Store the referent expression in `AttributeExpression`. (#224)
  - This is already true in `VariantExpression`.
- [x] Rename `ExternalArgument` to `VariableReference`. (#214)
- [x] Introduce the `VariantList` value type. (#232)
- [x] Rename `NumberExpression` to `NumberLiteral`. (#214)
- [x] Rename `StringExpression` to `StringLiteral`. (#214)
- [x] Remove the abstract `Expression` AST node. (2d5b48c81849406574465b800696bc1d49bceaf8)
- [x] Rename `SelectExpression.expression` to `selector`. (8fc69440fcc091821d45aca4e90b8ee51b604569)
- [x] Parse `Comments` as standalone when the `Message` they're attached to is `Junk`. (#235)

zbraniecki:
I started working on that and hit two incompatibilities between fluent-syntax and reference parser that I'd like to flag:

 - TermReference's name is `-term`. It seems a bit redundant and inconsistent. VariableReference's name is `var, not `$var`.
 - When a comment is attached to a message it's part of its AST, but when there's an error in that message, its comment is added to AST separately and only the message is turned into junk. That seems inconsistent.

I'd like to get your opinion @stasm - is that an intentional change?

Pike:
I think I can unblock us here before Tuesday, thus:

> TermReference's name is -term. It seems a bit redundant and inconsistent. VariableReference's name is var, not $var.

I think that's good that way. The IDs of a Term also include the '-' as they are represented in the source. Variables are actually a bit odd, in that their API-level name is without the `$`, but their l10n-level name is with it. If we wanted to make that consistent, I'd lean towards changing the `VariableReference` rather than `Term` and `TermReference`.

I don't think this changed the way referencing Terms works, the existing behavior just got a dedicated name.

> When a comment is attached to a message it's part of its AST, but when there's an error in that message, its comment is added to AST separately and only the message is turned into junk. That seems inconsistent.

[...]

PR #214, opened 2018-06-03
Rename AST nodes for Fluent 0.6

zbraniecki:
Simple renames for #206.

### DIFF
--- a/fluent-syntax/src/ast.js
+++ b/fluent-syntax/src/ast.js
@@ -91,18 +91,18 @@ export class Expression extends SyntaxNode {
   }
 }
 
-export class StringExpression extends Expression {
+export class StringLiteral extends Expression {
   constructor(value) {
     super();
-    this.type = "StringExpression";
+    this.type = "StringLiteral";
     this.value = value;
   }
 }
 
-export class NumberExpression extends Expression {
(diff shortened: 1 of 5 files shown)


fluent-syntax/src/ast.js
```diff
@@ -115,10 +115,10 @@ export class MessageReference extends Expression {
   }
 }
 
-export class ExternalArgument extends Expression {
+export class VariableExpression extends Expression {
```
stasm: This should be called `VariableReference`.

fluent-syntax/src/serializer.js
```diff
@@ -227,7 +227,7 @@ function serializeMessageReference(expr) {
 }
 
 
-function serializeExternalArgument(expr) {
+function serializeVariableExpression(expr) {
```
stasm: `…VariableReference` here as well.

fluent-syntax/src/ast.js
zbraniecki: Ahh, great catch! I was following https://github.com/projectfluent/fluent.js/issues/206 which has it as `VariableExpression`. Fixing :)

stasm: My bad. I fixed the checklist in #206.

PR #215, opened 2018-06-03
Call expr 0.6

zbraniecki:
Addresses two items from #206 list:

- Named arguments to CallExpression must now follow all positional ones.
- Forbid duplicate named arguments to CallExpressions.

### DIFF
--- a/fluent-syntax/src/serializer.js
+++ b/fluent-syntax/src/serializer.js
@@ -277,18 +277,14 @@ function serializeVariantExpression(expr) {
 
 function serializeCallExpression(expr) {
   const fun = serializeFunction(expr.callee);
-  const args = expr.args.map(serializeCallArgument).join(", ");
-  return `${fun}(${args})`;
-}
-
-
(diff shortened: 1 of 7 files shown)


zbraniecki:
ugh, git isn't very good with patchsets. I don't know how to submit only the latter :/

fluent-syntax/src/serializer.js
```diff
@@ -277,18 +277,14 @@ function serializeVariantExpression(expr) {
...
-  switch (arg.type) {
-    case "NamedArgument":
-      return serializeNamedArgument(arg);
-    default:
-      return serializeExpression(arg);
+  const positional = expr.positional.map(serializeExpression).join(", ");
+  const named = expr.named.map(serializeNamedArgument).join(", ");
+  if (expr.positional.length > 0 && expr.named.length > 0) {
+    return `${fun}(${positional}, ${named})`;
+  } else if (expr.positional.length > 0) {
```
stasm: Nit: no need for the `else` here.

PR #221, opened 2018-06-06
Remove 0.4 support from fluent-syntax

zbraniecki:
This will make it easier to implement the preserve-msg-comment on error change, and overall red > green :)


fluent-syntax/src/parser.js
```diff
@@ -61,21 +60,7 @@ export default class FluentParser {
     while (ps.current()) {
       const entry = this.getEntryOrJunk(ps);
 
-      if (entry === null) {
-        // That happens when we get a 0.4 style section
```
stasm: I don't think I see where `null` is returned in `getEntryOrJunk`. Did we change it and forget to remove from here?

zbraniecki: idk :(

PR #226, opened 2018-06-21
Add TermReference

stasm:


fluent-syntax/src/parser.js
```diff
@@ -179,32 +190,20 @@ export default class FluentParser {
...
-      } else {
-        ps.skipInlineWS();
-      }
-    }
-
-    if (id.name.startsWith("-") && pattern === undefined) {
-      throw new ParseError("E0006", id.name);
+    if (ps.isPeekPatternStart()) {
+      ps.skipIndent();
+      var pattern = this.getPattern(ps);
```
zbraniecki: not a huge fan of hoisting via `var` over explicit local definitions via `let`, so I'd prefer the latter.

fluent-syntax/src/parser.js
```diff
@@ -214,6 +213,26 @@ export default class FluentParser {
...
 
+  getTerm(ps, comment) {
+    const id = this.getTermIdentifier(ps);
+
+    ps.skipInlineWS();
+    ps.expectChar("=");
+
+    if (ps.isPeekPatternStart()) {
+      ps.skipIndent();
+      var pattern = this.getPattern(ps);
```
zbraniecki: same here.

fluent-syntax/src/parser.js
stasm: I'd like to keep `var`, it's what it does best.

stasm: To elaborate: `var` is well suited for use inside of `if…else` or `try…catch` blocks and keeps to code more concise. As you know, I usually advocate for using `let` and `const` but here I feel like it's the perfect tool for the job.

I'm going AFK. If you agree and if this PR blocks you, feel free to merge it as-is. Thanks!

zbraniecki: I don't, but that's because I value code readability above conciseness in the `explicit over implicit` manner.

Not blocking.

Issue #175, opened 2018-03-30
The use of <template> prevents server-side rendering


jamesreggio:
For context, I'm trying to get `fluent-react` working within [Next.js](https://github.com/zeit/next.js/).

Next.js provides seamless server-side rendering of the React component hierarchy, which means that dependencies utilized during React rendering need to avoid dependencies upon the browser's runtime environment.

It appears that the only major issue preventing the use of Fluent during server-side rendering is the dependency upon the HTML `<template>` element within `fluent-react` for quick parsing of HTML. It'd be great if there was another approach that could do away with this lone dependency upon `document`.

zbraniecki:
Would `DOMParser`[0] be an acceptable alternative? Also, is there no polyfill for `<template>`?

[0] https://developer.mozilla.org/en-US/docs/Web/API/DOMParser

jamesreggio:
@zbraniecki — not quite. The problem here is during server-side rendering, which happenes in the absence of all DOM APIs. Every `template` polyfill I've found leverages some other DOM APIs, which are not available.

Conceivably, we could use [`xmldom`](https://www.npmjs.com/package/xmldom) as a server-side `DOMParser` implementation, and then use the built-in API in the browser.

stasm:
There might be a way to allow passing custom markup sanitization functions to `<LocalizationProvider>`. I'll take a closer look at this in a week, when I'm back from vacation.

The code in `<Localized>` expects the parsed translation to have an interface of a `DocumentFragment`. Custom sanitization functions would still need to conform to this interface. On server-side, you could use `jsdom` to return a `DocumentFragment` with sanitized translation.

We actually use `jsdom` in our tests to be able to run them in node:

https://github.com/projectfluent/fluent.js/blob/b44b80205ae1dca0ed24e34d2e96842c4ee93550/fluent-react/test/__setup.js#L3-L13

Maybe you could do something similar in your app?

stasm:
#229 has a proposed fix for this.

stasm:
I closed #229 in favor of #233, also by @blushingpenguin. I think we might be able to land it today.

stasm:
I just merged #233 which fixes this. Thanks, @blushingpenguin!

I need to fix a few more things before I publish a new version of `fluent-react`. The theme of this next version is _compatibility_ and it requires some changes in the core `fluent` package as well. I've created a [project board](https://github.com/projectfluent/fluent.js/projects/4) to track it.

[...]

PR #229, opened 2018-06-21
add jsdom as a peer dependency

blushingpenguin:
This adds jsdom as a peer dependency, and in fluent-react uses it when not running in a browser environment.  This is useful for SSR as the get the code to work all that you need to do is to install jsdom.

### DIFF
--- a/fluent-react/package.json
+++ b/fluent-react/package.json
@@ -50,6 +50,7 @@
     "prop-types": "^15.6.0"
   },
   "peerDependencies": {
+    "jsdom": "^10.0.0 || ^11.0.0",
     "react": "^0.14.9 || ^15.0.0 || ^16.0.0"
   },
   "devDependencies": {
@@ -57,7 +58,7 @@
     "babel-plugin-transform-rename-import": "^2.2.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
-    "jsdom": "^11.6.2",
+    "jsdom": "^11.10.0",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
     "sinon": "^4.2.2"
--- a/fluent-react/src/markup.js
+++ b/fluent-react/src/markup.js
@@ -1,6 +1,12 @@
-/* eslint-env browser */
-
-const TEMPLATE = document.createElement("template");
+/* global window, require */
+let doc = null;
+if (typeof(window) === "undefined" || !window || !window.document) {
+  const jsdom = require("jsdom");
+  doc = new jsdom.JSDOM("<html><body></body></html>").window.document;
+} else {
+  doc = window.document;
+}
+const TEMPLATE = doc.createElement("template");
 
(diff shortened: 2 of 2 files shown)


fluent-react/src/markup.js
```diff
@@ -1,6 +1,12 @@
-/* eslint-env browser */
-
-const TEMPLATE = document.createElement("template");
+/* global window, require */
+let doc = null;
+if (typeof(window) === "undefined" || !window || !window.document) {
```
stasm: Could we check for the existence of the `document` global rather than `window` and then `document`?

fluent-react/src/markup.js
```diff
@@ -1,6 +1,12 @@
...
+  const jsdom = require("jsdom");
+  doc = new jsdom.JSDOM("<html><body></body></html>").window.document;
```
stasm: How about we use the minimal valid HTML5 document here?

    <!doctype html><title>JSDOM</title>

fluent-react/src/markup.js
```diff
@@ -1,6 +1,12 @@
-/* eslint-env browser */
-
-const TEMPLATE = document.createElement("template");
+/* global window, require */
+let doc = null;
```
stasm: Nit: I'd prefer to use a `var` in both branches of the `if…else`, but this is of course fine as well.

fluent-react/package.json
```diff
@@ -50,14 +50,15 @@
     "prop-types": "^15.6.0"
   },
   "peerDependencies": {
+    "jsdom": "^10.0.0 || ^11.0.0",
```
stasm: Why support both v10 and v11?

stasm: Actually, should we add this here at all? For the SSR use-case, `jsdom` is more of an optional peer dependency and I'm worried this will result in warnings printed by `npm install fluent-react`.

mdw211:
I added this:
https://github.com/projectfluent/fluent.js/pull/233
hopefully it is something like your alternative solution proposed in #175

jamesreggio:
Just my two cents, but JSDom is a very expensive dependency to take when the only thing depending upon DOM APIs is the use of `template`.

The expense may not be a problem when performing server-side rendering, but it's of critical important on platforms like React Native.

stasm:
@jamesreggio Thanks, that's valuable feedback. I haven't considered React Native. Let's go with the alternative solution @blushingpenguin implemented in #233, which is to allow custom markup parsers to be passed to `<LocalizationProvider>` when `document` is not available. For SSR, the custom parser might want to require `jsdom`. For React Native something more complex might be necessary. Or not—the consumers of the library can now choose to ignore parsing altogether, as long as the custom parser returns an array of `Node` objects. I'll document what the return value must be exactly in https://github.com/projectfluent/fluent.js/wiki/React-Overlays later today.

Thanks for this PR, @blushingpenguin. I'm going to closing in favor of #233.

PR #230, opened 2018-06-21
allow Localized with no children

blushingpenguin:
This allows <Localized id="xxx"/> which returns a string.  This saves having to wrap all text content in an element, which is sometimes undesirable.  (I think that returning strings from a components render function was added in React 16).


fluent-react/src/localized.js
```diff
@@ -82,7 +82,8 @@ export default class Localized extends Component {
   render() {
     const { l10n } = this.context;
     const { id, attrs, children } = this.props;
-    const elem = Children.only(children);
+    const elem = typeof(children) !== "undefined" ?
+      Children.only(children) : null;
```
stasm: We should still allow fallback strings in the JSX:

```jsx
<Localized id="hello">
    Hello, world!
</Localized>
```

In such cases, `Localized` still receives a single child.

fluent-react/src/localized.js
```diff
@@ -116,6 +117,10 @@ export default class Localized extends Component {
       }
     }
 
+    if (!elem) {
```
stasm: Given my comment above, I think it would be better to test `!isValidElement(elem)`. You can also move this up a bit, above the logic which translates attributes which doesn't apply for strings.

fluent-react/test/localized_valid_test.js
```diff
@@ -45,13 +45,11 @@ suite('Localized - validation', function() {
...
-        <Localized />,
-        { context: { l10n: new ReactLocalization([]) } }
-      );
-    }
-    assert.throws(render, /a single React element child/);
+    const wrapper = shallow(
+      <Localized />,
+      { context: { l10n: new ReactLocalization([]) } }
+    );
+    assert.equal(wrapper.length, 1);
```
stasm: Please bring this test back and add a new one in `localized_render_test.js` which tests `<Localized>` with a string-typed child.

fluent-react/src/localized.js
blushingpenguin: Ok, that makes sense -- I've changed this to use just the message if children isn't a valid react element as suggested.

fluent-react/test/localized_valid_test.js
blushingpenguin: The test is still there -- I've just changed it not to check for an exception as it doesn't throw one anymore.

I've added a few test cases for fallback string children (with/without message) and empty children (with/without message) as suggested.

blushingpenguin:
I've updated the PR with the requested changes.  Sorry for the delay!

fluent-react/src/localized.js
```diff
@@ -82,18 +82,23 @@ export default class Localized extends Component {
...
+    // Validate that children isn't an array
+    if (Array.isArray(children)) {
+      throw new Error("<Localized/> expected to receive a single " +
+        "React node child");
+    }
 
     if (!l10n) {
       // Use the wrapped component as fallback.
-      return elem;
+      return children;
```
stasm: I'd prefer to keep the previous version with `elem`, please. It make it clearer that there should only be one child element. Same comment applies to other places in this PR.

blushingpenguin:
I've (at last!) renamed children back to elem as requested

PR #231, opened 2018-06-21
add typescript definitions

blushingpenguin:
This adds typescript definition files (.d.ts) which allow fluent-react to be used from typescript easily


stasm:
@blushingpenguin Sorry I haven't looked at this earlier. Is there some way to make sure type definitions stay up to date as we continue making changes to `fluent.js`? E.g. is some kind of CI testing possible?

demurgos:
@stasm 
To keep the definitions up-to-date, the best is to generate them from source: rename the source files from `.js` to `.ts` and use the Typescript compiler to generate both the `.d.ts` and `.js` files.

Another solution would be to just use Typescript for the test files and use the manually written definitions.

Overall, I agree that type definitions are very valuable, especially for consumers: they improve code completion and enable static checks.

If you're interested, I could try to send a minimal PR to convert the project to TS: do not change the code, just add the compiler.

stasm:
> To keep the definitions up-to-date, the best is to generate them from source: rename the source files from .js to .ts and use the Typescript compiler to generate both the .d.ts and .js files.

I'd describe myself as _cautious_ when it comes to converting `fluent` to TypeScript :) I'm not opposed to it per se, but sticking to vanilla JS gives us a direct control over what the source code looks like. It also makes it easy to land [`fluent` code in `mozilla-central`](https://searchfox.org/mozilla-central/source/intl/l10n/Fluent.jsm), which continues to be a very important driver behind the project :)

> Another solution would be to just use Typescript for the test files and use the manually written definitions.

I like this idea a lot. The API surface is small enough that it might just work.

> If you're interested, I could try to send a minimal PR to convert the project to TS: do not change the code, just add the compiler.

I'm sorry for not replying earlier. I'm very interested in doing _something_ to help TypeScript users use Fluent. I also have more time to tend to this project, now that Fluent Syntax 0.8 has been released.

If you add the compiler, what will the consequences for the code in `fluent/src` be? Can we start with just `fluent/test` and try your second suggested approach first?

demurgos:
Hi,
Thank you for your reply. Agreed, using Typescript only for the tests can be a good first step. I used this kind of setup for some of the Gulp plugins.

I'll try to send a PR building on top of this one.

blushingpenguin:
These definitions are out of date and subsequently definitions were submitted to DefinitelyTyped (of which those definitions are now also out of date!).  I have submitted some up to date definitions to DT so this PR isn't useful anymore.

PR #232, opened 2018-06-21
Add VariantList

zbraniecki:
This is not yet complete, but I'd like to get first feedback round.

I'm in particular a bit confused about the Messages only accepting Patterns, while Terms accept VariantList | Pattern.

Should I remove null-selector SelectExpression and replace it with VariantList? Does it mean that VariantList is a valid Expression that can show up in Pattern? If so, what's the difference between a Term with VariantList and a Term with a Pattern with a VariantList?

What's the value of preventing Messages from having VariantList if they can have a Pattern with a Placeable with a VariantList?

3 tests fail, pending answers to my questions :)


zbraniecki:
Fixed #206

stasm:
Thanks for looking into this, @zbraniecki!

> I'm in particular a bit confused about the Messages only accepting Patterns, while Terms accept VariantList | Pattern.

We're exploring `VariantLists` as values of `Messages` in https://github.com/projectfluent/fluent/issues/131. It's not part of Syntax 0.6 because it would require a new syntax to access the variants: `message[variant]`.

> Should I remove null-selector SelectExpression and replace it with VariantList? Does it mean that VariantList is a valid Expression that can show up in Pattern? If so, what's the difference between a Term with VariantList and a Term with a Pattern with a VariantList?

Yes, please remove `null` selector `SelectExpression`. However, don't replace it with `VariantList`. `VariantList` is a value type, not an expression. It's only valid as a value of `Terms` and `Variants` for now. It may not be used as a child of a `Placeable` in a `Pattern`.

> What's the value of preventing Messages from having VariantList if they can have a Pattern with a Placeable with a VariantList?

They can't. The [EBNF only allows `SelectExpression` or `InlineExpression` as `Placeables`](https://github.com/projectfluent/fluent/blob/d2a5363f413d9b963f9b2299329760156e6a7ba9/spec/fluent.ebnf#L29), and[ `SelectExpressions` must have selectors](https://github.com/projectfluent/fluent/blob/d2a5363f413d9b963f9b2299329760156e6a7ba9/spec/fluent.ebnf#L61).

Note that since `VariantLists` can be values of `Variants`, and `Variants` are used both in other `VariantLists` and in `SelectExpressions`, we needed to add a [post-processing check](https://github.com/projectfluent/fluent/blob/d2a5363f413d9b963f9b2299329760156e6a7ba9/syntax/abstract.mjs#L62-L68) in the reference parser which forbids `VariantLists` as descendants of `SelectExpressions`. I mentioned this in https://github.com/projectfluent/fluent/issues/102#issuecomment-385069390.

fluent-syntax/src/parser.js
```diff
@@ -221,7 +222,7 @@ export default class FluentParser {
 
     if (ps.isPeekPatternStart()) {
```
stasm: We may need to rename this to `isPeekValueStart`.

fluent-syntax/src/parser.js
```diff
@@ -394,6 +395,17 @@ export default class FluentParser {
     return new AST.NumberLiteral(num);
   }
 
+  getValue(ps) {
+    let pattern = this.getPattern(ps);
+
+    if (pattern.elements.length === 1 &&
+        pattern.elements[0].type === "Placeable" &&
+        pattern.elements[0].expression.type === "VariantList") {
+      return pattern.elements[0].expression;
```
stasm: Could we instead remove the following code from `getExpression` and move it up here? 

https://github.com/projectfluent/fluent.js/blob/108ed2438fa37416f38f5786323f4b89d2d38d58/fluent-syntax/src/parser.js#L542-L548

`getPattern` should not be able to successfully parse selector-less `SelectExpressions`.

stasm: We'll also need to change `getVariant` to use `getValue` rather than `getPattern`:

https://github.com/projectfluent/fluent.js/blob/108ed2438fa37416f38f5786323f4b89d2d38d58/fluent-syntax/src/parser.js#L382

And then, in `getSelectorExpression` ensure that none of the parsed variants has a `VariantList` for value:

https://github.com/projectfluent/fluent.js/blob/108ed2438fa37416f38f5786323f4b89d2d38d58/fluent-syntax/src/parser.js#L582

zbraniecki:
Ok, I updated as much as I could to your feedback, but am confused with tests. There's some minor issue with serialized adding a space in the closing bracket, but more importantly, there are some tests that expect select expression on message without selector, and I seem to get a pattern every time I try to get a variant list in my own test.

@stasm - if you have time on Monday, feel free to take it over. If not, I'll resume on Tuesday.

stasm:
I gave this a try yesterday and another one today. I'll comment on the PR in its current state and then I'll push my attempt to fix things.

fluent-syntax/src/parser.js
```diff
@@ -394,6 +395,17 @@ export default class FluentParser {
     return new AST.NumberLiteral(num);
   }
 
+  getValue(ps) {
+    if (ps.isPeekNextLineVariantStart()) {
```
stasm: This doesn't work because `isPeekNextLineVariantStart` expect to be called at the end of line. We need to first check for the opening `{` here.

fluent-syntax/src/parser.js
```diff
@@ -394,6 +395,17 @@ export default class FluentParser {
...
+      const variants = this.getVariants(ps);
+
+      ps.expectIndent();
```
stasm: …and expect the closing `}` here.

fluent-syntax/src/serializer.js
```diff
@@ -207,6 +215,8 @@ function serializeExpression(expr) {
       return serializeCallExpression(expr);
     case "SelectExpression":
       return serializeSelectExpression(expr);
+    case "VariantList":
+      return serializeVariantList(expr);
```
stasm: This isn't needed. `VariantLists` are not expressions and may not be children of `Pleceables`.

fluent-syntax/src/serializer.js
```diff
@@ -249,6 +259,17 @@ function serializeSelectExpression(expr) {
   return parts.join("");
 }
 
+function serializeVariantList(expr) {
```
stasm: This function should add the opening and the closing braces around the `VariantList`.

fluent-syntax/test/fixtures_structure/term.json
```diff
@@ -19,8 +19,7 @@
           {
             "type": "Placeable",
             "expression": {
-              "type": "SelectExpression",
-              "selector": null,
+              "type": "VariantList",
```
stasm: `VariantLists` may not be children of `Placeables`.

stasm:
@zbraniecki OK, I pushed a few commits to your branch. They should fix all my feedback. Let me know if they look good. @Pike, would you mind taking a look at this as well?

fluent-syntax/test/fixtures_behavior/variant_lists.ftl
```diff
@@ -0,0 +1,41 @@
...
+        *[one] One
+    }
+
+#~ ERROR E0014, pos 59
+message2 =
+    {
+        *[one] {
+            *[two] Two
+         }
+    }
```
Pike: This test doesn't seem to do anything beyond message1, AFAICT. At least, it dies with the same error message on the same code point.

stasm: Good point, I'll remove it.

fluent-syntax/src/serializer.js
```diff
@@ -131,23 +131,47 @@ function serializeAttribute(attribute) {
...
+    pattern.elements.some(isSelectExpr) ||
+    pattern.elements.some(includesNewLine);
 
   if (startOnNewLine) {
-    return `\n    ${content}`;
+    return `\n    ${indent(content)}`;
   }
 
-  return ` ${content}`;
+  return ` ${indent(content)}`;
```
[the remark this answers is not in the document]
stasm: It's not, good catch. All cases which need `indent()` are handled in the `startOnNewLine` branch.

zbraniecki:
Thanks stas! This looks great. I struggled to understand how VariantList fits into the Pattern which with your fixes looks very clean to me now.

Requesting r?s to land it.

PR #233, opened 2018-06-22
added parseMarkup to LocalizationProvider's props and context

blushingpenguin:
Did you mean something like this rather than https://github.com/projectfluent/fluent.js/pull/229#pullrequestreview-131109139?

### DIFF
--- a/fluent-react/src/localized.js
+++ b/fluent-react/src/localized.js
@@ -1,6 +1,5 @@
 import { isValidElement, cloneElement, Component, Children } from "react";
 import PropTypes from "prop-types";
-
 import { isReactLocalization } from "./localization";
 import { parseMarkup } from "./markup";
 import VOID_ELEMENTS from "../vendor/voidElementTags";
@@ -84,6 +83,8 @@ export default class Localized extends Component {
     const { id, attrs, children } = this.props;
     const elem = Children.only(children);
 
+    const markupParser = this.context.parseMarkup || parseMarkup;
+
     if (!l10n) {
       // Use the wrapped component as fallback.
       return elem;
@@ -139,7 +140,7 @@ export default class Localized extends Component {
 
     // If the message contains markup, parse it and try to match the children
     // found in the translation with the props passed to this Localized.
-    const translationNodes = Array.from(parseMarkup(messageValue).childNodes);
+    const translationNodes = Array.from(markupParser(messageValue).childNodes);
     const translatedChildren = translationNodes.map(childNode => {
       if (childNode.nodeType === childNode.TEXT_NODE) {
         return childNode.textContent;
@@ -172,7 +173,8 @@ export default class Localized extends Component {
 }
 
 Localized.contextTypes = {
-  l10n: isReactLocalization
+  l10n: isReactLocalization,
+  parseMarkup: PropTypes.func
 };
 
 Localized.propTypes = {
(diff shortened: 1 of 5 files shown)


stasm:
Thanks for the PR! I was away for a few days; I'll review this this week.

stasm:
This dropped off my radar because of the Fluent Syntax 0.6 release. Sorry about that. A number of projects have reported this problem so I'm going to look at this PR today.

fluent-react/src/markup.js
```diff
@@ -1,6 +1,7 @@
 /* eslint-env browser */
 
-const TEMPLATE = document.createElement("template");
+const TEMPLATE = typeof(document) !== "undefined" ?
+  document.createElement("template") : null;
 
 export function parseMarkup(str) {
```
stasm: I'd like to provide a meaningful error message in case we reach the default `parseMarkup` but `document` is `undefined`. In the current PR, if no custom markup parser function is given by the developer, the default `parseMarkup` will throw a `TypeError` trying to assign `str` to `null.innerHTML`.

Rather than check if `TEMPLATE === null` in every call to the default `parseMarkup`, I suggest to go one level up on the ladder of abstraction and define a markup parser _creator_. Here's a default one which is defensive; something like this:

```js
function createParseMarkup() {
  if (typeof(document) === "undefined") {
    // We can't use <template> to sanitize translations.
    throw new Error(
      "Without `document` translations cannot be safely sanitized. " +
      "Pass a custom `function parseMarkup(str: string): string;` prop " +
      "to the <LocalizationProvider>."
    );
  }

  const template = document.createElement("template");

  return function parseMarkup(str) {
    template.innerHTML = str;
    return template.content;
  };
}
```

fluent-react/src/localized.js
```diff
@@ -84,6 +83,8 @@ export default class Localized extends Component {
     const { id, attrs, children } = this.props;
     const elem = Children.only(children);
 
+    const markupParser = this.context.parseMarkup || parseMarkup;
```
stasm: With a `createParseMarkup` creator, this can become:

```js
const parseMarkup = this.context.parseMarkup || createParseMarkup;
```

Or even, using default values and destructuring:

```js
const {parseMarkup = createParseMarkup()} = this.context;
```

stasm: (I'd prefer for the variable used throughout the file to be called `parseMarkup` rather than `markupParser`.)

stasm: Actually, this isn't such a great idea :) We'd end up re-creating the default `parseMarkup` function for every `Localized`. Sorry for the confusion.

blushingpenguin: You could just cache the function once created to avoid that (personally I'd just move the check for document being missing into Localized)

blushingpenguin:
sure, just merge your modified version and close the PR?

fluent-react/src/localized.js
stasm: I added caching :)

stasm:
I pushed a commit to your branch. Would you mind taking a look? I added `createParseMarkup` in a way which I decribed in my earlier comments.

I have another idea: maybe it would make sense to move this logic entirely into the `<LocalizationProvider>` constructor and _always_ use `context.parseMarkup` inside `<Localized>`? What do you think?

blushingpenguin:
Your changes look good.  I do agree with your proposed tweak though -- it would seem better to call createParseMarkup from LocalizationProvider and pass the resulting function down through context -- that way the Localized instances aren't all creating markup parsers.

stasm:
OK, thanks for the review :) And more importantly, for doing most of the work here. Let's merge this in.

stasm:
I added some docs at https://github.com/projectfluent/fluent.js/wiki/React-Overlays#custom-markup-parsers.

Issue #239, opened 2018-07-03
 Port fixtures from the reference parser


stasm:
The reference parser has a [bunch of fixtures](https://github.com/projectfluent/fluent/tree/master/test/fixtures) in its tests. I'd like to copy them into `fluent-syntax/test/fixtures_reference` and set up a test suite to use them in order to test `fluent-syntax`'s parser.

We'll need to ignore spans and annotations for now.

stasm:
Hey @zbraniecki, I know you said you wrote some code for this. Let me know if you'd like to take this issue. If not, I can work on it tomorrow.

stasm:
This turns out to be much more work because of incompatibilities in the parsing behaviors of the tooling parser and the reference parser. Here's the inventory of differences I've found so far. I'm treating the behavior of the reference parser as the desired one.

- [ ] Blank lines should end `Junk`.
- [ ] Comments' `content` should include the trailing `\n`.
- [ ] Nested placeables should be possible.
- [ ] Broken `Attributes` should not break the whole message. (#237)
- [ ] Multiline `Patterns` should keep the indents of each line except the first one.
- [ ] Trailing whitespace should be kept.
- [ ] The `message[variant]` syntax should be forbidden.
- [ ] Escape sequences should be preserved in the raw value.
- [ ] `CallExpression` should allow newlines and blanklines between arguments.
- [ ] `NamedArgument`'s `val` field should be called `value`.

I've checked those which I have fixes for.

stasm:
The discrepancies which I'd like to fix in the next version of `fluent-syntax` have been fixed on the `feature/syntax-0.6` branch. In https://github.com/projectfluent/fluent.js/commit/b32a35dfc5e4d2b3031caea9a97a67195c6e5a87 I unskipped most of the reference fixtures.

I moved #248 and #237 out to a future milestone. I'm going to file new issues for the remaining incompatibilities (currently skipped in tests). I'll use https://github.com/projectfluent/fluent.js/projects/3 to track them.

We'll close this issue when the `feature/syntax-0.6` branch is merged into `master`.

stasm:
Actually, let's close it now. We still have #206 tracking all of the `feature/syntax-0.6` work.

PR #241, opened 2018-07-04
Reference multiline patterns

zbraniecki:
Fixes the multiline inlines in #239.

### DIFF
--- a/fluent-syntax/src/parser.js
+++ b/fluent-syntax/src/parser.js
@@ -480,7 +480,6 @@ export default class FluentParser {
         }
 
         ps.next();
-        ps.skipInlineWS();
 
         // Add the new line to the buffer
         buffer += ch;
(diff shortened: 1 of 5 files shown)


zbraniecki:
This is on top of  #232 so really only the last commit should be reviewed.

Pike:
As I said on IRC, I think this is an issue in the reference parser more than an actual spec part. At least in the scope of 0.6.

zbraniecki:
Rebased on top of syntax-0.6 to simplify assessing the code.

stasm:
This is being discussed in https://github.com/projectfluent/fluent/issues/150.

stasm:
As per the decision in https://github.com/projectfluent/fluent/issues/150, we won't need this for now. Closing the PR. Thanks, @zbraniecki, for submitting it in the first place.

PR #242, opened 2018-07-04
Reference nested placeables

zbraniecki:
Fixes nested placeables for #239

### DIFF
--- a/fluent-syntax/test/fixtures_behavior/placeable_in_placeable.ftl
+++ b/fluent-syntax/test/fixtures_behavior/placeable_in_placeable.ftl
@@ -8,7 +8,7 @@
 #   }
 
 key4 = {  { foo }
-#~ ERROR E0014, pos 90
+#~ ERROR E0003, pos 97, args "}"
 
 
 # key5 = { foo } }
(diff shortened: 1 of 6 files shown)


stasm:
Hey @zbraniecki, thanks for the patch. The change looks good, but it's a bit hard to review without at least one test testing the desired parse output. Also, for small changes like this basing the PR on top of `feature/syntax-0.6` would make it easier to review and land. Can you make these changes, please?

zbraniecki:
> The change looks good, but it's a bit hard to review without at least one test testing the desired parse output.

My assumption is that we're landing this to enable reference tests which will replace our current tests. Do you want to duplicate tests between them?

stasm:
Yes, at least one test please. I'm still not sure we'll be able to unskip the reference fixtures.

fluent-syntax/test/fixtures_behavior/placeable_in_placeable.ftl
```diff
@@ -8,7 +8,7 @@
 #   }
```
stasm: Can we un-comment `key1`-`key3` now?

zbraniecki:
Hmm, `key1` and `key2` - yes, for `key3` - no, and it seems to me that it should not work based on https://github.com/projectfluent/fluent/blob/master/syntax/grammar.mjs#L162 - it only allows for inline ws between first `{` and second `{`. Did I miss something?

stasm:
No, you're right, thanks. `key3` is related to https://github.com/projectfluent/fluent/issues/106 which is planned for Syntax 0.7.

PR #243, opened 2018-07-04
Reference call args

zbraniecki:
Part of #239.


stasm:
Please rebase this on top of `feature/syntax-0.6` and add a test.

zbraniecki:
I don't understand why travis fails here. The json file is in the patch.

stasm:
It looks for a JSON fixture in the `fluent/` directory. `make -C fluent fixtures` will generate it.

fluent-syntax/test/fixtures_structure/expressions_call_args.json
```diff
@@ -0,0 +1,120 @@
...
+                    },
+                    "type": "Identifier"
+                  },
+                  "span": {
+                    "end": 19,
+                    "start": 12,
+                    "type": "Span"
+                  },
+                  "type": "NamedArgument",
+                  "val": {
```
stasm: If you rebase this PR on top of `feature/syntax-0.6` and re-generate the fxitures, `val` will become `value` and the test will pass. Travis tests the result of the merge rather than the actual PR, which is why this fails right now.

fluent/test/fixtures_structure/expressions_call_args.json
```diff
@@ -0,0 +1 @@
+{}
```
stasm: This means that the runtime parser wasn't able to parse the fixture. I'd like to see it fixed. I think it's OK to not enforce the indent in the runtime parser and I'm hopeful that a simple replacement of `skipInlineWS` to `skipWS` will suffice.

PR #244, opened 2018-07-06
Add FluentResource

zbraniecki:
This implements the first part of https://bugzilla.mozilla.org/show_bug.cgi?id=1455649 in fluent.js allowing for caching of parsed resources.

### DIFF
--- a/fluent/src/context.js
+++ b/fluent/src/context.js
@@ -1,16 +1,5 @@
 import resolve from "./resolver";
-import parse from "./parser";
-
-/**
- * Fluent Resource is a structure storing a map
- * of localization entries.
- */
-class FluentResource extends Map {
-  constructor(entries, errors = []) {
-    super(entries);
-    this.errors = errors;
-  }
-}
+import FluentResource from "./resource";
 
 /**
  * Message contexts are single-language stores of translations.  They are
@@ -107,11 +96,6 @@ export class MessageContext {
     return this._messages.get(id);
   }
 
-  static parseResource(source) {
-    const [entries, errors] = parse(source);
-    return new FluentResource(Object.entries(entries), errors);
-  }
-
   /**
    * Add a translation resource to the context.
    *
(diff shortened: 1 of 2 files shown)


fluent/src/context.js
```diff
@@ -115,22 +131,45 @@ export class MessageContext {
    * @returns {Array<Error>}
    */
   addMessages(source) {
-    const [entries, errors] = parse(source);
-    for (const id in entries) {
+    const res = MessageContext.parseResource(source);
```
stasm: Please use `this` here rather than `MessageContext` to allow for subclassing.

zbraniecki: Using `this` throws on babel unfortunately here :(

zbraniecki:
> Given that this is supposed to help the Gecko use-case, could you weigh on which approach is the best from the Gecko POV?

I'm not strongly opinionated here. Happy to go either way.

Eventually there may be some complication when we try to transfer `FluentResource` between processes. I'm not sure how much hassle it adds to have methods defined on the data structure there, but we can cross that bridge when we come to it.

zbraniecki:
Should I move the parser to `FluentResource` then, @stasm?

stasm:
> Eventually there may be some complication when we try to transfer FluentResource between processes. I'm not sure how much hassle it adds to have methods defined on the data structure there, but we can cross that bridge when we come to it.

My hope and the intent behind the `extends Map` design is that we will be able to "downgrade" to a regular `Map` in IPC scenarios, which the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) supports. This would happen automatically whenever a `FluentResource` instance is `postMessaged`.

Once its FTL is parsed, `FluentResource` doesn't really need any parsing capabilities. `MessageContext.addResource` can effectively take anything with an interface of a `Map`.

So yeah, let's move parsing to `FluentResource`, please. It's OK to keep it the code in `parser.js` and import it into `resource.js`. We don't need to expose parsing as a public method of `FluentResource` AFAICT.

PR #245, opened 2018-07-06
Update FluentDOM classes to work with XPIDL

zbraniecki:
This implements the second part of https://bugzilla.mozilla.org/show_bug.cgi?id=1455649 in fluent.js allowing for caching of parsed resources.

The core shift is that now it is possible to construct the objects without arguments (optionally) - except of generateMessages which in Gecko scenario is supplemented by the `defaultGenerateMessages`.

### DIFF
--- a/fluent-dom/src/dom_localization.js
+++ b/fluent-dom/src/dom_localization.js
@@ -16,13 +16,12 @@ const L10N_ELEMENT_QUERY = `[${L10NID_ATTR_NAME}]`;
  */
 export default class DOMLocalization extends Localization {
   /**
-   * @param {Window}           windowElement
    * @param {Array<String>}    resourceIds      - List of resource IDs
    * @param {Function}         generateMessages - Function that returns a
    *                                              generator over MessageContexts
    * @returns {DOMLocalization}
    */
-  constructor(windowElement, resourceIds, generateMessages) {
+  constructor(resourceIds, generateMessages) {
     super(resourceIds, generateMessages);
 
     // A Set of DOM trees observed by the `MutationObserver`.
@@ -31,10 +30,8 @@ export default class DOMLocalization extends Localization {
     this.pendingrAF = null;
     // list of elements pending for translation.
     this.pendingElements = new Set();
-    this.windowElement = windowElement;
-    this.mutationObserver = new windowElement.MutationObserver(
-      mutations => this.translateMutations(mutations)
-    );
+    this.windowElement = null;
+    this.mutationObserver = null;
 
     this.observerConfig = {
       attribute: true,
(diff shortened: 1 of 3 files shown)


fluent-dom/src/dom_localization.js
```diff
@@ -132,6 +129,19 @@ export default class DOMLocalization extends Localization {
       }
     }
 
+    if (this.windowElement) {
+      if (this.windowElement !== newRoot.ownerGlobal) {
+        throw new Error(`Cannot connect a root:
```
stasm: Grammar nit: `a root` doesn't sound right here; I don't know how to make it right. `Cannot connect root` maybe?

fluent-dom/src/localization.js
```diff
@@ -17,21 +17,25 @@ export default class Localization {
    *
    * @returns {Localization}
    */
-  constructor(resourceIds, generateMessages) {
+  constructor(resourceIds = [], generateMessages) {
     this.resourceIds = resourceIds;
     this.generateMessages = generateMessages;
-    this.ctxs =
-      new CachedAsyncIterable(this.generateMessages(this.resourceIds));
+    if (resourceIds.length) {
```
stasm: Do we need this check here and in `onChange`? `generateMessages` and `CachedAsyncIterable` should work with all iterables, including empty ones. I think I'd prefer to not perform the check to signal that empty `resourceIds` is a valid scenario and not a special case.

fluent-dom/src/dom_localization.js
```diff
@@ -150,11 +160,17 @@ export default class DOMLocalization extends Localization {
...
-    this.resumeObserving();
 
-    return this.roots.size === 0;
+    if (this.roots.size === 0) {
+      this.mutationObserver = null;
+      this.windowElement = null;
+      return true;
+    }
+
+    this.resumeObserving();
```
stasm: Please add a comment here: `// Resume observing all other roots.`

fluent-dom/src/localization.js
```diff
@@ -152,8 +156,11 @@ export default class Localization {
    * that language negotiation or available resources changed.
    */
   onChange() {
-    this.ctxs =
-      new CachedAsyncIterable(this.generateMessages(this.resourceIds));
+    if (this.resourceIds.length) {
```
stasm: I'd remove this check. It's not needed; empty iterables are iterables too.

fluent-dom/src/dom_localization.js
```diff
@@ -150,11 +160,17 @@ export default class DOMLocalization extends Localization {
...
     this.roots.delete(root);
-    // Pause and resume the mutation observer to stop observing `root`.
+    // Pause the mutation observer to stop observing `root`.
     this.pauseObserving();
-    this.resumeObserving();
 
-    return this.roots.size === 0;
+    if (this.roots.size === 0) {
+      this.mutationObserver = null;
+      this.windowElement = null;
```
stasm: I'm not sure if this would be useful, but should we clear `this.pendingRAF` here as well, just to be safe?

fluent-dom/src/dom_localization.js
zbraniecki: Verified with bz - `a root` is correct.

PR #247, opened 2018-07-10
Trim trailing whitespace in TextElements

stasm:


fluent-syntax/test/fixtures_structure/whitespace_trailing.ftl
```diff
@@ -0,0 +1,5 @@
+#           < whitespace >
+key1 = Value              
+
+key2 = Value              {placeable}.
```
Pike: I guess you didn't mean to have the trailing `.` here?

stasm: Doch doch. I only meant to test that `TextElements` not in the last position are not affected by the trimming logic.

stasm:
> Can you add a test for nbsp or zwnj?

Ah, I see what you mean. Let me see.

> Also, if it is, use trimEnd, as that's what's standardized?

It's only available in Firefox 61+ and Chrome 66+, and not in node. I don't want to get bitten by its being missing in a `compat` build or something.

fluent/src/parser.js
```diff
@@ -310,8 +312,13 @@ class RuntimeParser {
       eol = this._length;
     }
 
-    const firstLineContent = start !== eol ?
-      this._source.slice(start, eol) : null;
+    // If there's any text between the = and the EOL, store it for now. The next
+    // non-empty line will decide what to do with it.
+    const firstLineContent = start !== eol
+      // Trim the trailing whitespace in case this is a single-line pattern.
+      // Multiline patterns are parsed anew by getComplexPattern.
```
zbraniecki: It would be nice to one day be able to pass first line to complex string and go from there.

stasm: It's doable, but keep in mind that we'd need to parse `firstLineContent`  in `getComplexPattern` anyways in case it contains placeables or escape sequences. The current approach of rewinding the index to the beginning of the first line works pretty well IMO.

PR #249, opened 2018-07-11
WIP Broken Attributes should not break the whole Entry 

### LINKED ISSUE #237: Broken Attributes should not break the whole Message
We should change the parser to comply with the decision in https://github.com/projectfluent/fluent/issues/132.

stasm:
_This is a WIP._ Fix #237.

To do:

- [ ] Runtime parser.
- [ ] Report the proper error.

### DIFF
--- a/fluent-syntax/src/parser.js
+++ b/fluent-syntax/src/parser.js
@@ -287,9 +287,16 @@ export default class FluentParser {
     const attrs = [];
 
     while (true) {
+      const lineStart = ps.getIndex() + 1;
       ps.expectIndent();
-      const attr = this.getAttribute(ps);
-      attrs.push(attr);
+
+      try {
+        const attr = this.getAttribute(ps);
+        attrs.push(attr);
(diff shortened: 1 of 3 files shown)


stasm:
@zbraniecki - Asking for your feedback on this. I implemented a naive approach which keeps the attributes parsed so far and the rewinds the index to the beginning of the line in which an error was found. The problem with this approach is that the error reporting is then performed by `getEntryOrJunk`. The original error message is lost. Do you have a recommendation for how to solve this given the current architecture of the tooling parser?

Pike:
We could store the attribute error on the parser, and compare its error position with the entry error position, and whoever got further wins?

stasm:
Yeah, that was my assumption too. I wanted to hear what @zbraniecki thinks of this approach as he is the original author of this code. I'm also not clear if it's better to store the error on the instance of `FluentParser` or perhaps `FTLParserStream`.

PR #254, opened 2018-07-13
Add an IntlMemoizer as an argument that can be shared between MessageContexts

zbraniecki:
This will allow us to implement https://bugzilla.mozilla.org/show_bug.cgi?id=1475356 and utilize a shared intl memoizer between contexts.

### DIFF
--- a/fluent/src/context.js
+++ b/fluent/src/context.js
@@ -1,5 +1,6 @@
 import resolve from "./resolver";
 import FluentResource from "./resource";
+import IntlMemoizer from "./intl_memoizer";
 
 /**
  * Message contexts are single-language stores of translations.  They are
@@ -52,7 +53,8 @@ export class MessageContext {
   constructor(locales, {
     functions = {},
     useIsolating = true,
-    transform = v => v
+    transform = v => v,
+    intlMemoizer = new IntlMemoizer(),
   } = {}) {
     this.locales = Array.isArray(locales) ? locales : [locales];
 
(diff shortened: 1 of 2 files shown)


fluent/src/context.js
```diff
@@ -210,14 +212,6 @@ export class MessageContext {
...
-    const cache = this._intls.get(ctor) || {};
-    const id = JSON.stringify(opts);
-
-    if (!cache[id]) {
-      cache[id] = new ctor(this.locales, opts);
-      this._intls.set(ctor, cache);
-    }
-
-    return cache[id];
+    return this._intlMemoizer.get(ctor, this.locales, opts);
```
stasm: Let's remove this method completely and use `ctx._intlMemoizer` directly in the resolver (in `types.js`).

fluent/src/intl_memoizer.js
```diff
@@ -0,0 +1,17 @@
+export default class IntlMemoizer {
```
stasm: We can remove some boilerplate by subclassing `WeakMap` here.

fluent/src/intl_memoizer.js
```diff
@@ -0,0 +1,17 @@
...
+  constructor() {
+    this._intls = new WeakMap();
+  }
+
+  get(ctor, locales, opts) {
+    const cache = this._intls.get(ctor) || new Map();
+    const id = {locales, ...opts};
+
+    if (!cache.has(id)) {
```
stasm: This looks rather Pythonic to me :) Sadly, it won't work in JS. `id` is a new object literal every time this code runs. `cache.has(id)` will always be `false`.

fluent/src/intl_memoizer.js
stasm: Also, should we export `IntlMemoizer` in `index.js`?

PR #256, opened 2018-07-17
Fix span.start for Messages/Terms with Comments, fixes #255

### LINKED ISSUE #255: spans for message with comment start at message start, not at comment start
For a message like

```
# leading comment
leading = message
```

The AST for the Message starts at the ID, not the message.

As this is a single AST node, I think the global span.start should be the earliest AST start. In https://github.com/projectfluent/fluent.js/blob/feature/syntax-0.6/fluent-syntax/test/fixtures_structure/whitespace_leading.json, the `key1` Message span should start at 0.

I'm having the same problem in the python impl, and that breaks compare-locales in the end. Ripples.

Pike:
Move the message.span.start logic from the wrapper to the point
where the comment is hooked up to the message.
As the message already has the span when it gets there, the
wrapper refuses to do it again.

Back-porting a similar fix I have for python.

### DIFF
--- a/fluent-syntax/src/parser.js
+++ b/fluent-syntax/src/parser.js
@@ -14,20 +14,13 @@ function withSpan(fn) {
       return fn.call(this, ps, ...args);
     }
 
-    let start = ps.getIndex();
+    const start = ps.getIndex();
     const node = fn.call(this, ps, ...args);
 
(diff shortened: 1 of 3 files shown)


stasm:
Feel free to push the Python fix to the `feature/syntax-0.6` branch in `python-fluent` :)

Issue #212, opened 2018-05-30
Move Cached*Iterable and mapContext out of the fluent package


stasm:
The `fluent` package currently exposes the following exports (among others):

  - `CachedSyncIterable`
  - `CachedAsyncIterable`
  - `mapContextSync`
  - `mapContextAsync`

The async exports have been a source of upgrading pains for web projects targeting older browsers. The transpiled `fluent/compat` build requires the `regeneratorRuntime` provided by `babel-polyfill` which can significantly increase the bundle size of a web app. This was the reason why we created the `fluent` 0.4.x branch which uses a Syntax 0.5-compatible parser but doesn't add the async iterables to the API.

I'd like to fix this by moving the iterables and the `mapContext` methods out of the `fluent` package, so that `fluent` can be updated independently of these iteration helpers.

stasm:
As the first step, I'd like to move `Cached*Iterable` to a new package called `cached-iterable`. It is agnostic to Fluent and may be useful for other codebases as well. I'll move the code into `projectfluent/cached-iterable`.

As far as `mapContext*` methods go, I think it should be possible to rewrite them without using the `for await` syntax. This would avoid requiring `babel-polyfill` with the `regeneratorRuntime`. `mapContextAsync` would be broken in older browsers, but at least it would be possible to import the package safely if only `mapContextSync` is needed. I'm still not sure if we should move both methods to a new package. I don't expect them to change much so perhaps it's better to keep them separate. Proposed names of the package: `fluent-map`, `fluent-fallback`, `fluent-sequence`.

@zbraniecki I'm looking for a second opinion here. Does this look like a good plan forward?

zbraniecki:
I'm totally with you on moving `CachedIterable` to a new package. It's a micropackage, but not very different from https://www.npmjs.com/package/intl-format-cache 

As to `mapContext*` I struggle a bit more. I'm not sure where they should live, and I'd like to minimize the damage the temporary state we're in imposes on our code base.
In other words - assuming that by the end of this year, all environments we target will support async iterators, will we want to keep it in a separate package? If not, how can we make it easy to bring it back once the reason disappears?

stasm:
[...]

PR #260, opened 2018-07-19
Add the cached-iterable dependency

stasm:
In #212 I suggested to move `CachedIterable` out of the `fluent` package. To this end, I have created [cached-iterable](https://github.com/projectfluent/cached-iterable). This PR makes `fluent`, `fluent-dom` and `fluent-react` use it.


stasm:
@zbraniecki Can you advise how to make `fluent-gecko` work with `cached-iterable`? Should it be inlined into `Localization.jsm` and `DOMLocalization.jsm`? Or rather, exported by its own module file, `CachedIterable.jsm`?

zbraniecki:
It will likely be imported into `Localization.jsm`. I'd prefer to avoid having to use a new jsm for this.

stasm:
Thanks, I'll make the necessary changes to `fluent-gecko`.

PR #261, opened 2018-07-19
Remove fluent-web

stasm:
fluent-web landed in June 2017 in #44. It was an example of an
end-to-end localization solution for the Web based on Fluent.

Since then we've been focused on fluent-dom which is used in Firefox,
and on fluent-react used by a number of Mozilla and non-Mozilla
websites.

fluent-web has had 0 releases in the last year. It's not used by any
Mozilla project. It's also currently broken, which we don't even know
about, because fluent-web has 0 tests.

And yet, every maintenance task that we perform in the repo must also
account for the existence of fluent-web. This adds a maintenance burden
without much benefit.


stasm:
Good idea. Let's land this now and we can find the best way to preserve this code in a separate repo later. I'm not sure if we should preserve the history, too, and we'll need to fix a few dependencies and the build system.

PR #262, opened 2018-07-26
make acceptedLanguages order the header by q values descending

blushingpenguin:
I'm not sure if anything actually sends this header with languages listed out of descending preference order, but according to the [specification](https://tools.ietf.org/html/rfc7231#section-5.3.5) you can't rely on it.

### DIFF
--- a/fluent-langneg/src/accepted_languages.js
+++ b/fluent-langneg/src/accepted_languages.js
@@ -1,7 +1,24 @@
-export default function acceptedLanguages(string = "") {
-  if (typeof string !== "string") {
+export default function acceptedLanguages(acceptLanguageHeader = "") {
+  if (typeof acceptLanguageHeader !== "string") {
     throw new TypeError("Argument must be a string");
   }
-  const tokens = string.split(",").map(t => t.trim());
-  return tokens.filter(t => t !== "").map(t => t.split(";")[0]);
+  const tokens = acceptLanguageHeader.split(",").map(t => t.trim());
+  const langsWithQ = [];
+  tokens.filter(t => t !== "").forEach((t, index) => {
+    const langWithQ = t.split(";").map(u => u.trim());
+    if (langWithQ[0].length > 0) {
+      let q = 1.0;
+      if (langWithQ.length > 1) {
+        const qVal = langWithQ[1].split("=").map(u => u.trim());
+        if (qVal.length === 2 && qVal[0].toLowerCase() === "q") {
+          const qn = Number(qVal[1]);
+          q = !isNaN(qn) ? qn : q;
+        }
+      }
+      langsWithQ.push({ index: index, lang: langWithQ[0], q });
+    }
+  });
+  // order by q descending, keeping the header order for equal weights
+  langsWithQ.sort((a, b) => a.q === b.q ? a.index - b.index : b.q - a.q);
+  return langsWithQ.map(t => t.lang);
 }
--- a/fluent-langneg/test/headers_test.js
+++ b/fluent-langneg/test/headers_test.js
@@ -29,6 +29,53 @@ suite('parse headers', () => {
     );
   });
 
+  test('with out of order quality values', () => {
+    assert.deepStrictEqual(
+      acceptedLanguages('en;q=0.8, fr;q=0.9, de;q=0.7, *;q=0.5, fr-CH'), [
+        'fr-CH',
+        'fr',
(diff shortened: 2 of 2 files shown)


fluent-langneg/src/accepted_languages.js
```diff
@@ -1,7 +1,24 @@
...
-  if (typeof string !== "string") {
+export default function acceptedLanguages(acceptLanguageHeader = "") {
+  if (typeof acceptLanguageHeader !== "string") {
     throw new TypeError("Argument must be a string");
   }
-  const tokens = string.split(",").map(t => t.trim());
-  return tokens.filter(t => t !== "").map(t => t.split(";")[0]);
+  const tokens = acceptLanguageHeader.split(",").map(t => t.trim());
+  const langsWithQ = [];
+  tokens.filter(t => t !== "").forEach((t, index) => {
```
zbraniecki: Can you please move the filter to line 5 together with trimming?

fluent-langneg/src/accepted_languages.js
```diff
@@ -1,7 +1,24 @@
...
+    const langWithQ = t.split(";").map(u => u.trim());
+    if (langWithQ[0].length > 0) {
```
zbraniecki: String.prototype.split always returns at least one element, so no need for this check I think?

fluent-langneg/src/accepted_languages.js
```diff
@@ -1,7 +1,24 @@
...
+      let q = 1.0;
+      if (langWithQ.length > 1) {
+        const qVal = langWithQ[1].split("=").map(u => u.trim());
+        if (qVal.length === 2 && qVal[0].toLowerCase() === "q") {
+          const qn = Number(qVal[1]);
+          q = !isNaN(qn) ? qn : q;
```
zbraniecki: That means that any non-numerical `q` value will get replaced with `1.0`, is that what we should do? Seems like bogus input should result in low priority, not high.

blushingpenguin:
Any chance of picking this one up?

Thanks,

Mark

stasm:
@zbraniecki ping :)

fluent-langneg/src/accepted_languages.js
```diff
@@ -1,7 +1,26 @@
...
   }
-  const tokens = string.split(",").map(t => t.trim());
-  return tokens.filter(t => t !== "").map(t => t.split(";")[0]);
+  const tokens = acceptLanguageHeader.split(",").map(t => t.trim())
+    .filter(t => t !== "");
+  const langsWithQ = [];
+  tokens.forEach((t, index) =>
+    langsWithQ.push(parseAcceptLanguageEntry(t, index)));
+  // order by q descending, keeping the header order for equal weights
+  langsWithQ.sort((a, b) => a.q === b.q ? a.index - b.index : b.q - a.q);
```
zbraniecki: Non-blocking cleanup suggestion:

```js
function parseAcceptLanguageEntry(entry) {
  const langWithQ = entry.split(";").map(u => u.trim());
  let q = 1.0;
  if (langWithQ.length > 1) {
    const qVal = langWithQ[1].split("=").map(u => u.trim());
    if (qVal.length === 2 && qVal[0].toLowerCase() === "q") {
      const qn = Number(qVal[1]);
      q = isNaN(qn) ? 0.0 : qn;
    }
  }
  return { lang: langWithQ[0], q };
}

const langsWithQ = Array.from(tokens.map(parseAcceptLanguageEntry).entries());
// order by q descending, keeping the header order for equal weights
langsWithQ.sort(([aidx, aval], [bidx, bval]) => aval.q === bval.q ? aidx - bidx : bval.q - aval.q);
return langsWithQ.map(([idx, val]) => val.lang);
```

I hand-written it so it may not launch but I hope the intention is expressed in my snippet. I'm trying to save us from reallocating the array in the loop. Alternatively, you could also likely do `const langsWithQ = Array(tokens.length);` which I guess is simpler.
The other thing I wanted to improve in my snippet is to not carry around the index, since the `langsWithQ` preserves the order of `tokens`, so there's no value in storing it on the `langWithQ` struct.

@blushingpenguin - do you like the proposed changes or would you prefer to land as-is?

blushingpenguin: Nicer your way I think -- I wasn't aware of Array.entries.  I have updated the code and retested it.

PR #264, opened 2018-07-31
Specify URL for JSDOM

stasm:
JSDOM 11.12.0 added support for localStorage. Without the url set to
a safe domain like localhost, tests fail due to security reasons.

See https://github.com/jsdom/jsdom/issues/2304


stasm:
I should note that the whole JSDOM setup could use a do-over. See https://github.com/jsdom/jsdom/wiki/Don't-stuff-jsdom-globals-onto-the-Node-global.

PR #270, opened 2018-08-09
Add node 10 to test matrix

Pike:


Pike:
Resurrecting old branches ...

This breaks because node 10 has `Intl`, but doesn't ship any ICU data.

I guess there are two ways to fix this, one is to figure out how to get ICU data into the node 10 tests, the other would be to skip the test that checks the polyfill. Condition could be `new Intl.PluralRules('pl').resolvedOptions().locale !== 'pl'`

@stasm ?

stasm:
For option 1, there's https://nodejs.org/api/intl.html#intl_providing_icu_data_at_runtime and https://www.npmjs.com/package/full-icu.

Pike:
Yeah, that's what I found as well. What I haven't found is how to hook that up to travis.

stasm:
We should look at the larger picture: `fluent-intl-polyfill` is going away (#293). I wouldn't bother about fixing this test; let's just remove the test (we can rely on the [upstream tests](https://github.com/eemeli/intl-pluralrules/blob/master/src/plural-rules.test.js) instead), or fix #293 first.

stasm:
I'd like to add node 9, 11, and 12 to the matrix, too, while we're at it.

stasm:
Or, actually, let's stick to the even numbers. 8, 10, and 12.

eemeli:
At least for Intl.PluralRules, the `intl-pluralrules` polyfill [will verify](https://github.com/eemeli/intl-pluralrules/blob/30b37fce4621a1ce1865d1d4a0e2d1cf5ca98300/src/polyfill.js#L16) that that an existing implementation actually works for multiple locales. But that's obviously not the only Intl object that Fluent cares about.

Pike:
@stasm might know why we're using a fork of an old version of `intl-pluralrules` ;-)

stasm:
Because I didn't have time to fix #293 yet :) Any takers?

Issue #269, opened 2018-08-09
Change how the compat builds are produced


stasm:
#133 changed the browser compatibility matrix of our `compat` builds. It was a good decision for Fluent and for the web. Unfortunately, while most browser support modern JavaScript, it turns out a significant portion of build tools like Webpack and more specifically, UglifyJS, do not 🤦‍♂️ .

I'm afraid we might need to re-implement strict ES5 compatibility in one way or another. I don't think we can expect existing projects to go through the whole ordeal of updating their build pipeline just to be able to use a newer version of `fluent` and friends.

Right now we produce two builds for each package:

- `index` - no transpilation at all, and
- `compat` - transpiled according to the modern compatibility matrix.

The moder compatiblity matrix expects the engine to support async functions, classes and generators, among others.

I propose we switch to the following strategy:

- `index` - transpiled according to the modern compatibility matrix,
- `compat` - transpiled to strictly ES5-compatible code,
- `vanilla` - not transpiled at all. This could also be called `modern` or `ecma`.

The `vanilla` builds will be allowed to use cutting-edge features like async iteration, implemented only in Firefox 57 and Chrome 63.

stasm:
@Pike, @zbraniecki - go / no go?

Pike:
`index` being `fluent/fluent.js`, right?

What I'd like us to have is some testing infrastructure that tells us if we're going to run into trouble while we're working on PRs. Asking a couple of stupid questions, because I know Jack about how JS works.

Is the ES5-compatible transpilation going to tell us that? Can we run tests against it? Track its size?

Also, do those entry points publish code we generate, or are they instructing other folks on how to generate it, and could they fail because of their own setup?

stasm:
> index being fluent/fluent.js, right?

Yes. I've used `index` because it's accessible as `import ... from "fluent"` rather than `from "fluent/something"`. I also think I'd change `fluent/fluent.js` to `fluent/index.js` and the same for all other packages as part of this change. `index` is usually how the main entry point is called in node-world.

> Is the ES5-compatible transpilation going to tell us that? Can we run tests against it? Track its size?

We could have a test which tries to run the ES5-compatible version and makes sure all known required polyfills are accounted for. Does this answer your question?

> Also, do those entry points publish code we generate, or are they instructing other folks on how to generate it, and could they fail because of their own setup?

I think we should publish all of them. Cosumers of the library will be then able to choose any of:

  - `import {MessageContext} from "fluent";`
[...]

PR #273, opened 2018-08-17
Create the fluent-sequence package

stasm:
The PR moves `mapContextSync` and `mapContextAsync` out of the `fluent` package. The goal is to simplify the dependency graph of higher-level packages like `fluent-react` as well as to reduce the build-time burden of `for await` used by `mapContextAsync`. We'll still need to deal with it in the `fluent-sequence` package which I'll do in #269.


stasm:
@Pike would you mind taking a look at this? It's a straight-forward move&rename PR. This doesn't yet solve the issue of the `regeneratorRuntime`. I'll deal with it in #269.

fluent-sequence/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "plural",
+    "gender",
+    "locale",
+    "language",
+    "formatting",
+    "translate",
+    "translation",
+    "format",
+    "parser"
+  ],
```
Pike: I suggest to review the tags here, many of them don't apply to this package.

Issue #280, opened 2018-08-27
Use markup of children if message is missing


cvle:
When we are writing integration tests, we don't provide translation files but rely on the fallback which is the children of the element to be localized. This works great, allowing us to write simpler tests. 

```jsx
<Localized button={<Button onClick={props.onGotoSignIn} />}>
  <Typography>
    {"Already have an account? <button>Sign In</button>"}
  </Typography>
</Localized>
```

However when we are using html markup overlays it becomes difficult to get e.g. in this case the rendered button to simulate clicks because as you can see in the image below, it is actually not being rendered if we don't provide the translation.

![image](https://user-images.githubusercontent.com/14221600/44691131-ab743400-aa5d-11e8-96a1-e54ca88b7861.png)

I propose to use the markup inside of the element's child in case the translation is missing essentially being rendered as this, allowing our tests to easily find and use the button.

![image](https://user-images.githubusercontent.com/14221600/44691338-93e97b00-aa5e-11e8-8a56-d28216f37772.png)

I'm attaching a PR with some changes that makes this possible. Please tell me what you think.

stasm:
Thanks, @cvle, for the report!

I understand not using translation files for unit testing, but I'm not sure I agree with not using them for integration testing. Are they too hard to set up?

I'm slightly afraid of making fallback _too good_, which can be a very effective way to hide localizability issues from developers using English as their main language.

cvle:
@stasm you have a point! We integrated actual translations into our integration tests now and see how it goes. I close for now and will reopen if we still need this 👍

PR #281, opened 2018-08-27
Use children markup for missing message

cvle:
Implements https://github.com/projectfluent/fluent.js/issues/280

### DIFF
--- a/fluent-react/src/localized.js
+++ b/fluent-react/src/localized.js
@@ -87,20 +87,24 @@ export default class Localized extends Component {
       return elem;
     }
 
-    const bundle = l10n.getBundle(id);
-
-    if (bundle === null) {
-      // Use the wrapped component as fallback.
-      return elem;
-    }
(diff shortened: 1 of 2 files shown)


fluent-react/src/localized.js
```diff
@@ -87,20 +87,24 @@ export default class Localized extends Component {
...
+      messageValue = result.value;
+      messageAttrs = result.attrs;
+    } else {
+      if (typeof elem.props.children !== "string") {
+        // Use the wrapped component as fallback.
+        return elem;
+      }
+      // Use the child string as our messageValue.
+      messageValue = elem.props.children;
+      messageAttrs = null;
```
stasm: In case of fallback, I think the attributes present in the source should also be rendered in the fallback.

cvle:
Closing for now, see https://github.com/projectfluent/fluent.js/issues/280

PR #272, opened 2018-08-16
Relax indentation WIP

unclenachoduh:
This PR implements the changes from https://github.com/projectfluent/fluent/pull/159/commits/a116b8ff4733ae8cc8fdeb2c9d6dea741791b41b Only require indentation before TextElements

Note: This PR removes call_expressions and select_expressions from the skip list in reference tests.

PR #284, opened 2018-09-14
Indentation/Whitespace in 0.7

zbraniecki:
This PR supersets #272. It covers all of the 0.7 indentation changes I could find, aligning the parser with the grammar.mjs as close I dared.

I *think* all the resulting test changes make sense as well. :)

### DIFF
--- a/fluent-syntax/src/ftlstream.js
+++ b/fluent-syntax/src/ftlstream.js
@@ -151,7 +151,9 @@ export class FTLParserStream extends ParserStream {
     return !includes(SPECIAL_LINE_START_CHARS, ch);
   }
 
-  isPeekValueStart() {
+  isValueStart({skip = true} = {}) {
+    if (skip === false) throw new Error("Unimplemented");
+
     this.peekBlankInline();
     const ch = this.currentPeek();
 
@@ -161,7 +163,7 @@ export class FTLParserStream extends ParserStream {
       return true;
     }
 
-    return this.isPeekNextLineValue();
+    return this.isNextLineValue();
   }
 
   // -1 - any
@@ -216,7 +218,9 @@ export class FTLParserStream extends ParserStream {
     return false;
   }
 
-  isPeekNextLineAttributeStart() {
+  isNextLineAttributeStart({skip = true} = {}) {
+    if (skip === false) throw new Error("Unimplemented");
+
     this.peekBlank();
 
     if (this.currentPeekIs(".")) {
@@ -228,7 +232,7 @@ export class FTLParserStream extends ParserStream {
     return false;
   }
 
-  isPeekNextLineValue(skipToPeek = true) {
+  isNextLineValue({skip = true} = {}) {
     if (!this.currentPeekIs("\n")) {
       return false;
     }
@@ -251,7 +255,7 @@ export class FTLParserStream extends ParserStream {
       }
     }
 
-    if (skipToPeek) {
+    if (skip) {
       this.skipToPeek();
     } else {
       this.resetPeek();
--- a/fluent-syntax/src/ftlstream.js
+++ b/fluent-syntax/src/ftlstream.js
@@ -4,14 +4,14 @@ import { ParserStream } from "./stream";
 import { ParseError } from "./errors";
 import { includes } from "./util";
 
-const INLINE_WS = [" "];
-const ANY_WS = [" ", "\r", "\n"];
+const INLINE_WS = " ";
+const ANY_WS = [INLINE_WS, "\n"];
 const SPECIAL_LINE_START_CHARS = ["}", ".", "[", "*"];
 
 export class FTLParserStream extends ParserStream {
   skipBlankInline() {
     while (this.ch) {
-      if (!includes(INLINE_WS, this.ch)) {
+      if (this.ch !== INLINE_WS) {
         break;
       }
       this.next();
@@ -21,7 +21,7 @@ export class FTLParserStream extends ParserStream {
   peekBlankInline() {
     let ch = this.currentPeek();
     while (ch) {
-      if (!includes(INLINE_WS, ch)) {
+      if (ch !== INLINE_WS) {
         break;
       }
       ch = this.peek();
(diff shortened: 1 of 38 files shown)


fluent-syntax/src/ftlstream.js
```diff
@@ -5,10 +5,11 @@ import { ParseError } from "./errors";
 import { includes } from "./util";
 
 const INLINE_WS = [" ", "\t"];
+const ANY_WS = [" ", "\t", "\r", "\n"];
```
Pike: Not sure if you want to do this as part of this patch or not, but `\t` is a regular char now.

fluent-syntax/test/fixtures_behavior/placeable_in_placeable.ftl
```diff
@@ -8,7 +8,7 @@ key2 = {  { foo }  }
 #   }
 
 key4 = {  { foo }
-# ~ERROR E0003, pos 93, args "}"
+# ~ERROR E0003, pos 96, args "}"
```
Pike: This is weird. Why is `#` OK in a placeable?

fluent-syntax/test/fixtures_behavior/indent.ftl
```diff
@@ -8,7 +8,6 @@ a }
 
 key3 = { a
 }
-# ~ERROR E0003, pos 36, args "}"
 
 key4 = {
 { a }}
```
Pike: The key4 error should also go away, I think.

fluent-syntax/src/ftlstream.js
zbraniecki: oh, cool! updated!

fluent-syntax/test/fixtures_behavior/placeable_in_placeable.ftl
zbraniecki: It's not. The error is exactly on that character:

```
! E0003 on line 2:
  | key4 = {  { foo }
  | # ~ERROR E0003, pos 93, args "}"
  … ^----- Expected token: "}"
```

zbraniecki:
Updated to your feedback! Thank you! :)

fluent-syntax/test/fixtures_behavior/placeable_in_placeable.ftl
zbraniecki: Ok, so we dove into this with Axel. The reason this annotation stops at 96 now, is that the test infra removes the `# ~ERROR` comment, so the parser continues and stops on the first character of `# key5` two `\n`'s later which is 96 :)

fluent-syntax/src/ftlstream.js
```diff
@@ -193,70 +202,49 @@ export class FTLParserStream extends ParserStream {
...
+      this.skipToPeek();
       return true;
     }
 
     this.resetPeek();
     return false;
   }
 
-  isPeekNextLineValue() {
+  isPeekNextLineValue(skipToPeek = true) {
```
stasm: Given that all `isPeek...` methods skip by default now, maybe we should call them simply `isNextLineValue()` etc?

Perhaps it's just me, but I often forget which methods skip and which do not. Making it a bit explicit would certainly make the code easier to follow for me:

```js
isPeekNextLineValue({skip}) { ... }
```

Then, we could use it _everywhere_ in a self-documenting fashion:

```js
ps.isNextLineValue({skip: true})
ps.isNextLineValue({skip: false})
```

stasm: For consistency, I'd make this change to all `isSomething` methods.

fluent-syntax/src/parser.js
```diff
@@ -416,34 +415,37 @@ export default class FluentParser {
...
+      ps.resetPeek();
     }
 
     return this.getPattern(ps);
   }
 
   getVariantList(ps) {
     ps.expectChar("{");
-    ps.skipInlineWS();
+    ps.skipBlank();
```
stasm: I think this needs to ensure at least one line break before moving on to parsing variants. Same as you do in `getExpression`.

fluent-syntax/src/parser.js
```diff
@@ -416,34 +415,37 @@ export default class FluentParser {
...
+    ps.skipBlankInline();
 
     let ch;
     while ((ch = ps.current())) {
 
       // The end condition for getPattern's while loop is a newline
       // which is not followed by a valid pattern continuation.
-      if (ch === "\n" && !ps.isPeekNextLineValue()) {
+      if (ch === "\n" && !ps.isPeekNextLineValue(false)) {
         break;
```
stasm: We shouldn't break here before first checking if there's an unindented placeable on the next line. The following snippets are valid Fluent 0.7:

```properties
key1 =
{ placeable}

key2 =
{ $selector ->
*[one] One
}
```

The reference parser has a number of fixtures for testing this in [select_indent.ftl](https://github.com/projectfluent/fluent/blob/master/test/fixtures/select_indent.ftl) and [variants_indent](https://github.com/projectfluent/fluent/blob/master/test/fixtures/variants_indent.ftl).

stasm: This might require a different logic than the one provided by `isPeekNextLineValue`. If so, my other comment about parametrizing `{skip: true}` might not hold anymore. I'd still rename `isPeekSomething` to `isSomething`, though.

fluent-syntax/src/parser.js
```diff
@@ -416,34 +415,37 @@ export default class FluentParser {
   getValue(ps) {
     if (ps.currentIs("{")) {
       ps.peek();
-      ps.peekInlineWS();
-      if (ps.isPeekNextLineVariantStart()) {
+      ps.peekBlank();
+      if (ps.currentPeek() === "*" || ps.currentPeek() === "[") {
```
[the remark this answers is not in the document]
zbraniecki: So, this is a bit more tricky. The situation is unusual in this place because we want to try to get a variant list, but if that doesn't work, we bail *back to the `{`* and go for `getPattern`.

That means that if we transition to use `isPeekNextLineVariantStart` I need to complicate it to not always skip.

fluent-syntax/src/ftlstream.js
zbraniecki: The previous in turn means that now not all of them skip, most of them do, but one of them doesn't, which makes it inconsistent as well :(

zbraniecki:
Updated to review, and copied reference tests. :) I was even able to remove a couple from the blacklist!

PR #286, opened 2018-09-26
Export FluentResource from fluent

stasm:
We added `FluentResource` in #244 but we kept it hidden. Let's expose it as an export of the `fluent` module.

### DIFF
--- a/fluent/src/index.js
+++ b/fluent/src/index.js
@@ -7,8 +7,6 @@
  *
  */
 
-export { default as _parse } from "./parser";
-
 export { default as FluentBundle } from "./context";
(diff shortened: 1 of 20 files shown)


fluent/src/index.js
```diff
@@ -7,9 +7,8 @@
  *
  */
 
-export { default as _parse } from "./parser";
-
-export { FluentBundle } from "./context";
+export { default as FluentBundle } from "./context";
```
stasm: I'll rename `context.js` to `bundle.js` in a follow-up commit (without a separate PR).

PR #289, opened 2018-10-10
The Optimistic Runtime Parser

stasm:
I re-wrote the runtime parser. It supports Syntax 0.7, runs against the reference fixtures, has half the lines of code, aaand it's a bit slower on SpiderMonkey (but also a bit faster on V8, so I assume the perf will improve over time).

### Goals

1. Support 100% of [Fluent Syntax 0.7](https://github.com/projectfluent/fluent/projects/3). This includes the indentation relaxation, dropping tabs and CR as syntax whitespace, normalizing new lines to LF, and only allowing numbers and identifiers as variant keys.
2. Maintain good performance. The parser is used in performance-critical code paths. Back in the days of Firefox OS it had to be both fast _and_ produce tightly packed results so that translations don't take up too much space on the device. I think the storage requirements can be relaxed these days.
3. Write code which will be easy to maintain in the future. The parser was first written even before Fluent branched off from L20n. It's seen many changes and additions over the last two years. As new features accrued it became hard to maintain it and also to keep track of all known bugs. My goal for the re-write was not only to clean it up but also to define the conformance story for the future and to improve the testing infrastructure.

### Design

The parser focuses on minimizing the number of false negatives at the expense of increasing the risk of false positives. In other words, it aims at parsing _valid_ Fluent messages with a success rate of 100%, but it may also parse some invalid messages which the reference parser would reject. The parser doesn't perform any validation and may produce entries  which wouldn't make sense in the real world. For best results users are advised to validate translations with the fluent-syntax parser pre-runtime.

The main parser loop iterates over the beginnings of messages and terms. This is to efficiently skip over comments (which have no use on runtime), and to recover from errors. When a fatal error is encountered, the parser instantly bails out of the currently-parsed message and moves on to the next one. Errors are discarded and are not visible to the users of `FluentResource`. The do carry a minimal description of what went wrong which may be useful when reading the code and for debugging, though.

The parser makes an extensive use of sticky regexes which can be anchored to any offset of the source string without slicing it. In some places, it's easier to just check the character currently at the cursor, so it does a fair share of that, too.

### Conformance

My original plan was to base the parser on the EBNF and only parse well-formed syntax. In this PR, I went for something a bit wider than that: a superset of well-formed syntax. The main deviation from the EBNF is related to parsing `VariantExpressions` and `CallExpressions`. The EBNF verifies that the they are called on `Terms` and `Functions` respectively. The optimistic parser doesn't differentiate between `Messages`, `Terms` and `Functions`. I decided to implement it this way because this code might soon change anyways (see https://github.com/projectfluent/fluent/issues/176).

Another deviation is that the parser treats commas in argument lists  as whitespace, similar to how Clojure treats them in sequence lists. I might suggest we upstream this in the spec, too, because it makes the implementation of args lists _much_ simpler.

I based this PR on top of the `zeroseven` branch. The `fluent-syntax` parser already supports Syntax 0.7 and passes the [reference fixtures](https://github.com/projectfluent/fluent/tree/master/test/fixtures). This made it possible to also turn on the reference testing in the runtime parser, too. `make fixtures` creates the parsed results for all reference fixtures; for now they must be verified manually before they're committed. `make test` can be used in development to assert that the output of the runtime parser still matches the committed one.

### Performance

I'm seeing a slight slowdown when running `make perf` with SpiderMonkey (`make perf-jsshell`). Interestingly, I'm also seeing a slight speed up in V8. 

```
SpiderMonkey 63 (in milliseconds, relative to master):
   mean:   5.99 (+26%)
   stdev:  0.22
   sample: 30

node 9.11 (in milliseconds, relative to master):
    mean:   3.14 (-15%)
    stdev:  0.12
    sample: 30
```

I asked @zbraniecki to run a Talos test using this branch. We'll see if the slow down is visible during Firefox startup and we'll then decide what do about it. I'm also seeing an opportunity to create a test case for the SpiderMonkey team. Maybe we're using features of the engine which haven't been optimized too much, yet? (This also applies to the `fluent-syntax` parser which is 4 times slower in SM than it is in V8).


stasm:
### How To Review

I'm not sure what the best way to review this PR is. @zbraniecki, @Pike, please chime in if you have thoughts on the best course of action. If you have time, please skim through the source of `fluent/src/resource.js` and flag the parts that are not clear and would benefit from better wording, either in code or in comments, or both.

It might be helpful to review some of the fixtures and the result of parsing by the optimistic parser. I recommend the following:

- [`multiline_values.ftl`](https://github.com/projectfluent/fluent/blob/master/test/fixtures/multiline_values.ftl) has a number of messages with complex multiline values, including values with placeables in tricky positions. Cf. the [parsed output](https://github.com/projectfluent/fluent.js/blob/de0796cd246ac7a8bc674c05363e0c33dfcf0b76/fluent/test/fixtures_reference/multiline_values.json).
- [`escaped_characters.ftl`](https://github.com/projectfluent/fluent/blob/master/test/fixtures/escaped_characters.ftl) tests known escape sequences in text elements and string literals. The runtime parser unescapes them during parsing. Cf. the [parsed output](https://github.com/projectfluent/fluent.js/blob/de0796cd246ac7a8bc674c05363e0c33dfcf0b76/fluent/test/fixtures_reference/escaped_characters.json).
- [`leading_dots.ftl`](https://github.com/projectfluent/fluent/blob/master/test/fixtures/leading_dots.ftl) sports a number of edge cases related to parsing the indent in multiline patterns. Cf. the [parsed output](https://github.com/projectfluent/fluent.js/blob/de0796cd246ac7a8bc674c05363e0c33dfcf0b76/fluent/test/fixtures_reference/leading_dots.json).
- [`select_expressions.ftl`](https://github.com/projectfluent/fluent/blob/master/test/fixtures/select_expressions.ftl) has a few examples of invalid selectors which parse nonetheless. `invalid-selector` parses because the runtime parser doesn't validate the expression in the selector. `nested-variant-list` parses becuase the parser doesn't perform the extra logic to forbid variant lists as descendants of patterns. Cf. the [parsed output](https://github.com/projectfluent/fluent.js/blob/de0796cd246ac7a8bc674c05363e0c33dfcf0b76/fluent/test/fixtures_reference/select_expressions.json).
- [`call_expressions.ftl`](https://github.com/projectfluent/fluent/blob/master/test/fixtures/call_expressions.ftl) demonstrates how invalid (but well-formed) arguments are still parsed by the runtime parser. Cf. `shuffled-args` and `duplicate-named-args` in the [parsed result](https://github.com/projectfluent/fluent.js/blob/de0796cd246ac7a8bc674c05363e0c33dfcf0b76/fluent/test/fixtures_reference/call_expressions.json).

stasm:
I may have found the cause of the slowdown which I observed yesterday. I documented the regexes used by the parser this morning and in the process I realized that the `RE_ATTRIBUTE_START` and the `RE_VARIANT_START` regexes could be used more idiomatically. Previously, I would use them inside of the corresponding `while` loop to break out of it. I changed the code to use the regex as the while loop's end condition instead. That's not what made a difference performance-wise, however. What did make a difference is that I also removed an unneeded `skipBlank` from the while loop. `Pattern` already skips all its trailing blank space; doing it again before the next attribute/variant wasn't useful.

```
With the extra skipBlank (milliseconds, compared to master):
  mean:   6.02 (+27%)
  stdev:  0.38
  sample: 30

Without it (milliseconds, compared to master):
  mean:   4.82 (+1%)
  stdev:  0.32
  sample: 30
```

fluent/src/resource.js
```diff
@@ -1,18 +1,467 @@
...
+// Common tokens.
+const TOKEN_BRACE_OPEN = /{\s*/y;
+const TOKEN_BRACE_CLOSE = /\s*}/y;
+const TOKEN_BRACKET_OPEN = /\[\s*/y;
+const TOKEN_BRACKET_CLOSE = /\s*]/y;
+const TOKEN_PAREN_OPEN = /\(\s*/y;
+const TOKEN_ARROW = /\s*->\s*/y;
+const TOKEN_COLON = /\s*:\s*/y;
+// As a deviation from the well-formed Fluent grammar, accept argument lists
+// without commas between arguments.
```
zbraniecki: I don't think I understand this comment. How's it different from other normalizations?

fluent/src/resource.js
```diff
@@ -1,18 +1,467 @@
...
+      return hasAttributes ? attrs : null;
+    }
+
+    function Pattern() {
+      // First try to parse any simple text on the same line as the id.
+      if (test(RE_TEXT_RUN)) {
+        var first = Match(RE_TEXT_RUN);
+      }
+
+      // If there's an backslash escape or a placeable on the first line, fall
```
zbraniecki: "there's a backslash"

fluent/src/resource.js
stasm: Can you help me re-word the comment? What do you mean by "other normalizations"? The regex under this comment has an optional comma (`.?`) which essentially means that commas in argument lists are treated as whitespace.

stasm:
Thanks, @zbraniecki! I've wondered about the naming (of course I have :)) and I like using the nouns over verbs specifically for writing parsers. I've used this same convention in the reference parser, so I might be biased by the code I already wrote before. To me, nouns represent will the concepts of tokens and grammar productions. I understand that not everyone may think that way.

I used the capitalization consistently for functions which return a parsed result, as opposed to functions which are operators or helpers. In a way, the parsers uses its own DSL to compose smaller productions into bigger ones. Again, this is something that I took away from writing the reference parser.

```js
let selector = InlineExpression();
if (token(TOKEN_BRACE_CLOSE)) {
    return selector;
}
```

Do you think using verbs would make the code easier to read for you and for newcomers? Example:

```js
let selector = parseInlineExpression();
if (skipToken(TOKEN_BRACE_CLOSE)) {
    return selector;
}
```

stasm:
Related: what's your take on using 4 spaces for indentation? I know Mozilla guidelines say 2 spaces for JS, but having more breathing room makes the code easier to read for me. I wonder if it's the same for other people. It also occurs to me that 2 spaces were perhaps a good compromise in the days of callbacks. A compromise which may have outlived its usefulness in the current days of async functions :) Can we turn this lint off in mozilla-central for Fluent files? This re-write would be a good opportunity to change the indentation.

fluent/src/resource.js
stasm: How does the following sound?

```js
// Note the optional comma. As a deviation from the Fluent EBNF, the parser
// accepts lists of call arguments without commas between them.
const TOKEN_COMMA = /\s*,?\s*/y;
```

zbraniecki:
> Do you think using verbs would make the code easier to read for you and for newcomers?

I don't think it matters that much. My only concern with `attrs.push(Attribute())` is that it looks like a class, rather than a function.

I looked at rust parser [0] and they do use `parse_foo`, `get_foo`, `expect_foo`, `eat_foo` etc.

> what's your take on using 4 spaces for indentation?

No strong opinion. I'm comfortable with 2 spaces in JS, C++ and with 4 in Rust, so I don't expect a big difference.

One thing is that I would prefer to migrate all at once and would prefer not to touch it while we're updating the code. Already I'd say that renaming variables/properties/classes/methods while updating the logic is the single greatest burden on reviewing the code in Fluent I encountered. I'd love not to add white space convention to that. :)

[0] https://github.com/rust-lang/rust/blob/master/src/libsyntax/parse/parser.rs

stasm:
OK, let's keep 2 spaces in this PR and discuss changing all code to 4 in a separate issue.

fluent/src/resource.js
zbraniecki: Ah, I see. I somehow confused the message as "accept argument lists with commas between arguments". Maybe - `As a deviation from the well-formed Fluent grammar, this parser does not enforce commas between arguments.`

stasm:
I decided to go with the _verb+noun_ naming scheme: `consumeToken`, `parseInlineExpression` etc. It's explicit, predictable, can't be confused with class construction, and hopefully won't catch anyone off-guard.

This is now ready to merge.

PR #291, opened 2018-10-12
Consistently require the skip option in isNext* methods

stasm:
While working on the Python port I noticed that there still was a slight inconsistency in method naming: sometimes it was `isNext.…` and sometimes `isPeekNext…`. In this PR I unify the naming and use the `skip` option in all invocations, as an explicit measure of documenting whether a method skips or just peeks.


stasm:
@zbraniecki Since you touched this code last, can you take a look? This doesn't change any logic, just method names and signatures.

PR #296, opened 2018-10-18
Properly parse CRLF

stasm:
In order to preserve spans, we need to actually parse CRLF rather than just replace it with LF before parsing.


stasm:
Phew, that was fast 🚀. Thanks for the review!

PR #298, opened 2018-10-19
End junk on blank lines to produce smaller Junk entries

### LINKED ISSUE #248: Multiple consecutive Junk entries should be allowed
According to the spec, `Junk` should only be parsed up until the next valid entry start _or_ a blank line. This effectively allows multiple consecutive `Junk` entries separated by blank lines.

stasm:
Fix #248.

This actually parses `Junk` better than the spec says. I filed https://github.com/projectfluent/fluent/issues/184 and https://github.com/projectfluent/fluent/issues/185 to fix the reference parser.

I'd like to land this as part of the Syntax 0.7 implementation.


Pike:
Before we go ahead, I'd prefer to think and test how adjacent Junk entries with their line endings touch cross channel. I don't like the idea of holding back zeroseven for that.

stasm:
OK, let's keep this on hold for now. I'll port this to Python when `zeroseven` merges into master, and I'd like to ask you for your help when it comes to testing this change with xchannel.

Pike:
Can you explain this once more? In https://github.com/projectfluent/fluent/pull/211, we moved the spec to what the tooling parsers do. Is this still open?

stasm:
Yes, this can be closed now, thanks.

PR #309, opened 2018-11-22
Syntax 0.8, part 2: Dedent multiline text to preserve content indentation

stasm:
Implement https://github.com/projectfluent/fluent/commit/d2229f2d7f5decd8fceb15b6276eaa90154a7241.


stasm:
@zbraniecki Here's another part of the Syntax 0.8 implementation. It turned out to be quite big, and I tried to split into a series of patches to make the review easier. However, I recommend reading the combined patch first, as it features a few cleanups and more comments than the intermediate commits.

This change has impact on Patterns spans, too. Previously, block patterns, i.e. patterns which start on a new line, would have their span start at the first non-blank character of the pattern:

```properties
# BEFORE
message =
    Value
    ^---- TextElement.span.start
    ^---- Pattern.span.start
```

I had to change this to be able to take the first line's indentation into account when dedenting the whole pattern. Here's where the spans start right now:

```properties
# AFTER
message =
    Value
    ^---- TextElement.span.start
^-------- Pattern.span.start
```

Let me know if you have any questions about this PR, I'll be happy to walk you though it!

fluent-syntax/src/parser.js
```diff
@@ -411,65 +409,170 @@ export default class FluentParser {
...
+    if (ps.isValueContinuation()) {
+      ps.skipToPeek();
+      return this.getValueType(ps, {isBlock: true, allowVariantList});
+    }
+
+    return null;
+  }
+
+  // Parse a VariantList (if allowed) or a Pattern.
+  getValueType(ps, {isBlock, allowVariantList}) {
```
zbraniecki: nit: since the `getValue` returns `Option<Value>` and `getValueType` returns `Value`, maybe we could rename it to `maybeGetValue` and `getValue`? I'm not super convinced that `maybe*` is the best naming scheme, but I don't like `getValueType` as it doesn't really convey the right difference from what `getValue` does.
Maybe you'll have a better idea. If not, it's ok to keep it imperfect :)

fluent-syntax/src/parser.js
```diff
@@ -411,65 +409,170 @@ export default class FluentParser {
...
+    ps.peekBlankInline();
+    if (allowVariantList && ps.currentPeek === "{") {
+      const start = ps.peekOffset;
       ps.peek();
       ps.peekBlankInline();
-      if (ps.isNextLineVariantStart({skip: false})) {
-        return this.getVariantList(ps);
+      if (ps.currentPeek === EOL) {
```
zbraniecki: how does this handle `\r\n`?

fluent-syntax/src/parser.js
```diff
@@ -411,65 +409,170 @@ export default class FluentParser {
...
+        const sum = new AST.TextElement(prev.value + element.value);
+        if (this.withSpans) {
+          sum.addSpan(prev.span.start, element.span.end);
+        }
+        trimmed[trimmed.length - 1] = sum;
+        continue;
       }
+
+      if (element.type === "Indent") {
+        // Is the indent hasn't been merged into a preceding TextElement,
```
zbraniecki: nit: "if"

fluent-syntax/src/parser.js
stasm: `ParserStream` normalizes `CRLF` to `LF`:

https://github.com/projectfluent/fluent.js/blob/e5cf0c0bbe69d7e97812f1241f18caef103a653c/fluent-syntax/src/stream.js#L13-L24

fluent-syntax/src/parser.js
stasm: I've been struggling with the naming of these methods and your comment suggests to me that my choices were not good enough :) Here's how the roles of these two methods differ:

- `getValue` distinguishes between inline and block values,
- `getValueType` distinguishes between `Patterns` and `VariantLists` (which are both value types).

Looking at this code today, I agree that this is confusing. Here's how I'm going to rename these methods before landing:

- `getValue` → `maybeGetValue`, to highlight the fact the it can return a `null`. It's the only `get*` method that does so. Thanks for the suggestion!
- `getValueType` → `getPatternOrVariantList`, to make its purpose clearer. _Value type_ was not immediately obvious.

Also note that when https://github.com/projectfluent/fluent/pull/220 lands, we'll be able to refactor this a bit and possibly remove `getPatternOrVariantList` because `VariantLists` will only be legal as values of `Terms`.

stasm:
Thanks for the review.

> I'd like to try to move away from touching the strings so much in the future and more to operate on indexes when dedenting. My suggested architecture would be to first collect start/end pointers for lines (with indents), and the leading indents as numbers. Then calculate the min of the indents and adjust the start pointer of each line by it and only then slice the TextElements out of it.

Yeah, I think this approach is worth exploring, perhaps even for parsing other productions, not only `Patterns`. I didn't want to make too big architectural changes in this PR. I don't feel comfortable making large-scale changes in this codebase and I was afraid of falling into a refactor-rabbit hole.

Issue #303, opened 2018-11-15
Implement Syntax 0.8


stasm:
Let's plan the implementation work in this issue. We'll submit PRs against a release branch, `zeroeight`, which I'll create later today.

Some 0.8 features don't need any work in `fluent.js`: https://github.com/projectfluent/fluent/pull/179 and https://github.com/projectfluent/fluent/pull/207 are already supported and https://github.com/projectfluent/fluent/pull/211 made the reference parser behave like the tooling one.

For features that need implementation, I'd like to follow the same order in which they landed in the spec. This will make it much easier to keep the reference tests green with each change. Here's the full list:

#### Part 1 (#307)

- [x] 1. Support astral Unicode characters in TextElements and StringLiterals
    https://github.com/projectfluent/fluent/pull/179, tests only
- [x] 2. Don't store the term sigil in the identifier name
    https://github.com/projectfluent/fluent/pull/183, tooling parser
- [x] 3. Treat backslash as normal char in TextElements
    https://github.com/projectfluent/fluent/pull/181, tooling parser, runtime parser
- [x] 4. Don't normalize CRLF to LF in Junk
    https://github.com/projectfluent/fluent/pull/193, tests only
- [x] 5. Forbid the closing brace in text
    https://github.com/projectfluent/fluent/pull/198, tooling parser, runtime parser

#### Part 2 (#309)

- [x] 6. Dedent multiline text to preserve content indentation
    https://github.com/projectfluent/fluent/pull/197, tooling parser, runtime parser

#### Part 3 (#310)

- [x] 7. Add FunctionReference
    https://github.com/projectfluent/fluent/pull/210, tooling parser
- [x] 8. Store unescaped content in StringLiteral.value and raw content in StringLiteral.raw
    https://github.com/projectfluent/fluent/pull/203, tooling parser

#### Part 4 (#311)

- [x] 9. Add Parameterized Terms
    https://github.com/projectfluent/fluent/pull/205, tooling parser, runtime parser, resolver

#### Part 5 (#312)

- [x] 10. Stop parsing Junk on lines which look like Entries
    https://github.com/projectfluent/fluent/pull/211, tests only
- [x] 11. Allow all Unicode characters
    https://github.com/projectfluent/fluent/pull/207, tests only

#### Part 6 (#313)

- [x] 12. Add Parameterized Attributes
    https://github.com/projectfluent/fluent/pull/212, tooling parser, runtime parser, resolver

#### Part 7 (#314)

[...]

PR #314, opened 2018-11-28
Syntax 0.8, part 7: \UHHHHHH and other changes

stasm:
Implement part 7 of #303.

- [x] 13.  Recognize \UHHHHHH as an escape sequence
    https://github.com/projectfluent/fluent/pull/201, tooling parser, runtime parser
- [x] 14. Change line endings in comments to line_end
    https://github.com/projectfluent/fluent/pull/219, tooling parser
- [x] 15. Forbid nested VariantLists
    https://github.com/projectfluent/fluent/pull/220, tooling parser

### DIFF
--- a/fluent-syntax/src/parser.js
+++ b/fluent-syntax/src/parser.js
@@ -6,10 +6,6 @@ import { ParseError } from "./errors";
 
 
 const trailingWSRe = /[ \t\n\r]+$/;
-// The Fluent Syntax spec uses /.*/ to parse comment lines. It matches all
-// characters except the following ones, which are considered line endings by
-// the regex engine.
-const COMMENT_EOL = ["\n", "\r", "\u2028", "\u2029"];
 
 
 function withSpan(fn) {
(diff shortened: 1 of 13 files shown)


stasm:
This PR depends on #313 which hasn't been merged yet. @Pike, if you want to review this first, the [compare view](https://github.com/stasm/fluent.js/compare/zeroeight-part6-macro-attrs...zeroeight-part7) might be helpful. Thanks.

fluent-syntax/src/parser.js
```diff
@@ -599,37 +590,44 @@ export default class FluentParser {
...
+  }
 
-    if (next === "u") {
-      let sequence = "";
-      ps.next();
-
-      for (let i = 0; i < 4; i++) {
-        const ch = ps.takeHexDigit();
+  getUnicodeEscapeSequence(ps, u, digits) {
+    ps.expectChar(u);
```
Pike: We already checked this in the switch statement, I'd just ps.next() here, and drop the `u` argument.

stasm: I'd prefer to keep `expectChar`. In fact, I'm planning a clean up after 0.8 where we get rid of most of `ps.next()`. I always seem to forget what it's supposed to consume, something `expectChar` is very explicit about.

stasm: Oh, and we need the `u` arg for the error message. That's the real reason why it's there.

Issue #208, opened 2018-05-23
FluentBundle.format() should accept a string identifier of a message or a path to an attribute


stasm:
If #205 lands, we could change `MessageContext.format()` to accept an identifier of a message and format it directly to string. It would avoid the intermediate step of having to first `getMessage` the message from the context, which is a bit clunky and is an artifact of how the bindings retrieve translated attributes.

```js
// BEFORE
let message = ctx.getMessage("message-key");
let translation = ctx.format(msg, args);
```

```js
// AFTER
let translation = ctx.format("message-key", args);
```

For attributes, we could introduce the following syntax:

```js
// AFTER
let translation = ctx.format("message-key.attr-name", args);
```

zbraniecki:
I'd like to suggest a transition period.

First, we would update `format` to take `id` or `id.attr` just like you did in Rust.

We would still preserve the `get_message` as semi-private way used by FluentDOM only for getting the whole shape all at once.

Then we can experiment with using the new `format` the way you suggested and analyze performance impact. I'm pretty skeptical about it because it just runs more code, but irrelevant, getting `format` to act the way we expect would be a big step forward and if we keep the `get_message` or not after that is less significant.

stasm:
@zbraniecki Would that require moving the following logic into the `fluent` package?

https://github.com/projectfluent/fluent.js/blob/873b409385d73c9e8c181243226f5096241d8dbd/fluent-dom/src/localization.js#L208-L227

Or do you have something different in mind?

zbraniecki:
yes, I believe so. It would return a formatted shape `Message {value, attributes}` in runtime.

stasm:
I'd prefer to keep this code in `fluent-dom` and in `fluent-react`. As you know, I'm not sure it represents the right approach to formatting compound messages. I'd like to keep `fluent` agnostic and unopinionated. The bindings modules are better suited to host this code given that they are higher up in the ladder of abstraction and can be more opinionated.

Which leads me to conclude that during the transition period, `fluent` should have three methods:

  1. the current `getMessage(id: str) -> Message`
  1. the current `format(resolvable: Message|Attribute) -> string?`
  1. a new `formatString(id: str) -> string?`

[...]

PR #322, opened 2018-12-05
Format

unclenachoduh:
Fix format for issue #208


fluent/src/bundle.js
```diff
@@ -191,23 +191,88 @@ export default class FluentBundle {
...
+      return resolve(this, args, message, errors);
+    }
+
+    if (parts.length === 2) {
+      if (message.attrs === null) {
+        errors.push(`Message has no attributes: "${id}"`);
+        return null;
+      }
+
+      let attr = message.attrs[parts[1]];
```
stasm: To keep it consistent with the `parts.length === 1` branch above, perhaps you could add an `if (typeof attr === "string")` fast path? I'm not sure they're needed and I'll do some perf testing when I get a chance. For now it would be best to keep both branches as similar to each other as possible.

fluent/src/bundle.js
```diff
@@ -191,23 +191,88 @@ export default class FluentBundle {
...
-  format(message, args, errors) {
-    // optimize entities which are simple strings with no attributes
-    if (typeof message === "string") {
-      return this._transform(message);
+
+  format(path, args, errors) {
+
+    var parts = path.split(".");
+    var id = parts[0];
+    var message = this._messages.get(id);
```
stasm: Please use `let` for all of these.

fluent/src/bundle.js
```diff
@@ -191,23 +191,88 @@ export default class FluentBundle {
...
+    };
+
+    if (message.attrs === undefined) {
+      return compoundShape;
+    }
+
+    for (let attr of Object.keys(message.attrs)) {
+      compoundShape.attributes.set(
+        attr,
+        message.attrs[attr]
```
stasm: You'll need to call `resolve` here to format the attribute to a string.

stasm:
@unclenachoduh Do you know if you'll be available in the coming weeks to finish this PR? I'd like to land it in early February.

unclenachoduh:
Hey @stasm I can get on this weekend to resolve the changes you've asked for. I'm a little busy right now, so I don't think I'll be able to write the new tests for `compound`.

stasm:
I'll be happy to help with the tests. I'd like to land this next week; let me know if you're still available to apply my feedback. I understand that you're busy so no worries if not.

fluent/src/bundle.js
unclenachoduh: I'm not sure where to place this fast path and which return it correlates to. Should this go in line 229?

fluent/src/bundle.js
unclenachoduh: I think I did this right.

tools/kekoa.js
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
unclenachoduh: You sure you don't want to keep it?

unclenachoduh:
I think I still have one of these changes pending. Take a look at the changes.

fluent/src/bundle.js
```diff
@@ -191,23 +178,90 @@ export default class FluentBundle {
...
-  format(message, args, errors) {
-    // optimize entities which are simple strings with no attributes
-    if (typeof message === "string") {
-      return this._transform(message);
+
+  format(path, args, errors) {
+
+    let var parts = path.split(".");
+    let var id = parts[0];
+    let var message = this._messages.get(id);
```
eemeli: These should presumably be `let parts ...` rather than `let var parts ...`?

fluent/test/functions_builtin_test.js
```diff
@@ -21,42 +21,33 @@ suite('Built-in functions', function() {
     test('missing argument', function() {
       let msg;
```
eemeli: This variable is not used. Also applies to `msg` in further tests in this file.

fluent/test/primitives_test.js
```diff
@@ -67,55 +65,43 @@ suite('Primitives', function() {
...
+      const val = bundle.format('foo', args, errs);
       assert.equal(val, 'Foo');
       assert.equal(errs.length, 0);
     });
 
-    test('is detected to be non-complex', function(){
-      const msg = bundle.getMessage('foo');
-      assert.equal(typeof msg, 'string');
-    });
-
```
eemeli: It would probably be better to mark this tests as skipped with `test.skip(...)` rather than just removing it.

fluent/test/primitives_test.js
```diff
@@ -143,41 +129,31 @@ suite('Primitives', function() {
...
+      const val = bundle.format('bar', args, errs);
       assert.equal(val, 'FooBar');
       assert.equal(errs.length, 0);
     });
 
-    test('is detected to be complex', function(){
-      const msg = bundle.getMessage('bar');
-      assert(Array.isArray(msg));
-    });
-
```
eemeli: It would probably be better to mark this test as skipped with `test.skip(...)` rather than just removing it.

eemeli:
I rather like this API, but I think the `getMessage()` deprecation ought to be split into a separate follow-on change, because really at the same time the behaviour of `hasMessage()` and `get messages` should also be reconsidered.

More specifically, I think `hasMessage()` should also support the `message.attr` path format, and `get messages` (or something like it) should iterate on message keys rather than `[key, message]` entries. Discussion on these ought to be considered separately from this change, though they are related.

stasm:
I'd like to finish @unclenachoduh's work and land this change soon. To clean things up a bit, I'll squash all commits from this PR and I'll open a new PR keeping attribution of the code written so far.

stasm:
> I rather like this API, but I think the getMessage() deprecation ought to be split into a separate follow-on change, because really at the same time the behaviour of hasMessage() and get messages should also be reconsidered.

With `format` taking an id (or `id.attr`), there aren't any reasons for keeping `getMessages`. It returns an internal representation of the message which no other method will be able to work with after the change to `format`. Do you have other use-cases for it in mind?

> More specifically, I think hasMessage() should also support the message.attr path format, and get messages (or something like it) should iterate on message keys rather than [key, message] entries. Discussion on these ought to be considered separately from this change, though they are related.

I'll file new issues about these, as you suggested. Thanks!

eemeli:
I have no other use cases in mind for `getMessages()`. Just wanted to be sure the `format()` & `compile()` refactoring would not get stuck waiting for the other related issues.

stasm:
I opened a new PR to continue working on this: #360. I updated and rebased the code and fixed a few issues. I'm going to close this PR now.

PR #326, opened 2018-12-19
Use external dom overlays

zbraniecki:
This is the set of code changes for using external dom-overlays.

I modified some of the tests that get different output now. Some of them I think are as expected, others are possibly a good indication of further changes we have to make, but should give you a good outlook at what changes.

### DIFF
--- a/fluent-dom/src/overlay.js
+++ b/fluent-dom/src/overlay.js
@@ -1,3 +1,15 @@
+const { localizeElement } = require("../../../fluent-domoverlays-js/src/index");
+const { JSDOM } = require("jsdom");
+
+function parseDOM(s) {
+  return JSDOM.fragment(s);
+}
+
+global.Node = {
+  ELEMENT_NODE: 1,
+  TEXT_NODE: 3,
+}
+
 /* eslint no-console: ["error", {allow: ["warn"]}] */
 /* global console */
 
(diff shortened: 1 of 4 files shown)


stasm:
Thanks, I'll take a look at your repo. What is your expectation wrt. to this PR? Do you want to land it, or is this just a demo?

zbraniecki:
I expect to eventually land it once we finalize the external crate. For now, I'm using it as a "here's what the new approach does to our tests".

fluent-dom/src/overlay.js
```diff
@@ -1,3 +1,10 @@
+import { translateNode }  from "../../../fluent-domoverlays-js/src/index";
+const { JSDOM } = require("jsdom");
+
+function parseDOM(s) {
+  return JSDOM.fragment(s);
```
stasm: We'll need to replace this with something built into the browser environment.

zbraniecki:
> Can you explain in more detail why you think external overlays are needed?

I start seeing DOM Overlays as a separate component that has different characteristics from the rest of `fluent-dom` - it's synchronous, independent of any other packages (i.e. it doesn't really care where you get the DOM fragments from) and we'll likely want to port it to compiled languages 1-1 which means that having a test suite that can be easily used from C++ or Rust will make porting easy.

I can see non-fluent projects eventually being interested in DOM Overlays logic - for example the mini-cache that Mossop is working on for Firefox is not really a fluent package and it doesn't depend on any other fluent component, but would use DOM Overlays directly.

Having it as a standalone component, much like we did for `cached-iterable` would make it easier to develop the particular functionality, ensure it remains stable and provide auxiliary features such as benchmarks, without adding more to `fluent-dom`.

zbraniecki:
I'm going to open a new one with a new draft PR

PR #332, opened 2019-01-29
Added option to FluentBundle to allow overriding messages and terms with a new value

marcselman:
By allowing overrides I can dynamically load new messages when needed.

### DIFF
--- a/fluent/src/bundle.js
+++ b/fluent/src/bundle.js
@@ -47,7 +47,7 @@ export default class FluentBundle {
    *   - `allowOverrides` - boolean specifying whether it's allowed to override
    *                      an existing message or term with a new value.
    *                      Default: false
-   * 
+   *
    *   - `transform` - a function used to transform string parts of patterns.
    *
    * @param   {string|Array<string>} locales - Locale or locales of the bundle
@@ -151,13 +151,13 @@ export default class FluentBundle {
       if (id.startsWith("-")) {
         // Identifiers starting with a dash (-) define terms. Terms are private
         // and cannot be retrieved from FluentBundle.
-        if (this._allowOverrides == false && this._terms.has(id)) {
+        if (this._allowOverrides === false && this._terms.has(id)) {
           errors.push(`Attempt to override an existing term: "${id}"`);
           continue;
         }
         this._terms.set(id, value);
       } else {
-        if (this._allowOverrides == false && this._messages.has(id)) {
+        if (this._allowOverrides === false && this._messages.has(id)) {
           errors.push(`Attempt to override an existing message: "${id}"`);
           continue;
         }
--- a/fluent/src/bundle.js
+++ b/fluent/src/bundle.js
@@ -109,8 +109,8 @@ export default class FluentBundle {
    *
    *     // Returns a raw representation of the 'foo' message.
    *
-   *     bundle.addMessages('bar = Bar', { allowOverrides: true });
-   *     bundle.addMessages('bar = Newbar');
+   *     bundle.addMessages('bar = Bar');
+   *     bundle.addMessages('bar = Newbar', { allowOverrides: true });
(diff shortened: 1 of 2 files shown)


stasm:
Thank you for the PR, @marcselman! Can you describe in a bit more detail in which situations this could be useful? When updated translations are available, I usually recommend creating a new bundle. This is similar in effect to the scenario when the language preference changes dynamically.

marcselman:
I have an app that I deploy to different domains (projects). Most of the strings are the same for each domain but on some I want to change a piece of text.

I have setup my .ftl files something like this:
```
fluent\en-US.ftl
fluent\nl-NL.ftl
fluent\project-group1\en-US.ftl
fluent\project-group1\nl-NL.ftl
fluent\project-group1\project1\en-US.ftl
fluent\project-group1\project1\nl-NL.ftl
fluent\project-group1\project2\en-US.ftl
fluent\project-group1\project2\nl-NL.ftl
```

The files in `fluent` are the base texts that are shared by all the projects
The project-group folders contain strings that are specific for that specific group of projects
The project folders contain specific strings that are only needed for that project

If for example I have 10 projects that share the same string I would put that in `fluent\en-US.ftl`.
But I want a different string for one specific project I would put that in `fluent\project-group1\project1\en-US.ftl`. Right now that gives the console error and does not override the string.
The only way to work around it is to move the string to all the separate project files so than I have 9 duplicate copies and one different. That makes it a lot of work to maintain.

marcselman:
@stasm Have you had some time to look at the code? It's a minimal change.

stasm:
Hi @marcselman, thanks again for the PR and for explaining your use-case. I've talked about this with @Pike and he suggested that the `addResource` and `addMessages` methods might be better places to pass this option, rather than setting it bundle-wide once. This would allow greater granularity in controlling when to allow overrides and when to forbid them. This could be used to discover duplicate message definitions across resources which are not meant as overrides, but rather base resources to be overridden. 

What do you think? Would you mind amending the PR to reflect this feedback? Ideally, we'd also add a few tests before landing it. Let me know what you think. Thanks!

marcselman:
Hi @stasm, I've updated the PR with the suggested changes. I hope it can be merged into the library so I can use it in my project.

stasm:
Thanks, @marcselman, nice work. Would also mind adding a simple test for this change?

I haven't run it, but something like this should do the trick. I would put it in [`fluent/test/context_test.js`](https://github.com/projectfluent/fluent.js/blob/master/fluent/test/context_test.js) around line 90. Or let me know and I'll be happy to do it before merging this PR.

```js
  suite('allowOverrides', function(){
    suiteSetup(function() {
      bundle = new FluentBundle('en-US', { useIsolating: false });
      let resource1 = FluentResource.fromString('key = Foo');
      let resource2 = FluentResource.fromString('key = Bar');
      bundle.addResource(resource1);
    });

    test('allowOverrides is false', function() {
      let errors = bundle.addResource(resource2);
      assert.equal(errors.length, 1);
      let msg = bundle.getMessage('key');
      assert.equal(bundle.format(msg), 'Foo');
    });

    test('allowOverrides is true', function() {
      let errors = bundle.addResource(resource2, {allowOverrides: true});
      assert.equal(errors.length, 0);
      let msg = bundle.getMessage('key');
      assert.equal(bundle.format(msg), 'Bar');
    });
});
```

marcselman:
@stasm Done. I've also added an extra `allowOverrides` test to the `addMessages` suite.

stasm:
Nice, thanks again! I'll publish a new version of `fluent` later today, or tomorrow.

marcselman:
🙌🏻 awesome!

stasm:
I published `fluent` 0.11.0 on npm today. Thank you, @marcselman, for the contribution!

PR #333, opened 2019-02-04
Travis: drop sudo:false directive

olleolleolle:
This PR removes the `sudo: false` directive, which is now unused.


stasm:
Thank you for the PR, @olleolleolle!

Issue #334, opened 2019-02-06
Language matching should add likely subtags again after clearing the region


stasm:
_Originally reported in https://github.com/mozilla-services/screenshots/issues/5347._

After digging into `fluent-langneg` code and comparing it to `LocaleService.cpp`, I think I can confirm that this is a bug in `fluent-langneg`. It comprises three parts.

1. The list of likely subtags in `fluent-langneg` is limited. It currently doesn't include data fro many locales, including `zh-HK`.
2. The matching algorithm in `fluent-langneg` is missing an extra step. After the region of the requested language is set to `*`, `LocaleService.cpp` tries to add likely subtags _again_. In combination with the step in which the requested region is cleared, this allows the algorithm to correctly select `zh-HK` based on the script used.
3. The list of available locales claims to be _unordered_, but in reality it uses its inherent ordering in some of the steps of the algorithm.

In other words, what currently happens is:

1. Request `zh-HK`.
1. Clear the region → `zh-*`.
1. Match against `zh-*-CN-*` and `zh-*-TW-*`.
1. Return `zh-CN` followed by `zh-TW`.

After adding the missing likely subtags, the bug persists:

1. Request `zh-HK`.
1. Add the likely subtag → `zh-Hant-HK`.
1. Clear the region → `zh-Hant-*`.
1. Match against `zh-*-CN-*` and `zh-*-TW-*`.
1. Return `zh-CN` followed by `zh-TW`.

Here's what should happen instead when we add the missing step to the algorithm:

1. Request `zh-HK`.
2. Add the likely subtag → `zh-Hant-HK`.
3. Clear the region → `zh-Hant-*`.
4. Add the likely subtag again → `zh-Hant-TW`.
5. Match against  `zh-*-TW-*`.
6. Return `zh-TW`.

zbraniecki:
taking. Thank you stas!

stasm:
This was fixed two weeks ago in #335. Thanks, @zbraniecki!

PR #335, opened 2019-02-06
Update the negotiation and Locale logic to match latest algo

zbraniecki:
This patch brings the JS implementation very close to the C++ and Rust implementations.

### DIFF
--- a/fluent-langneg/src/locale.js
+++ b/fluent-langneg/src/locale.js
@@ -34,53 +34,63 @@ export default class Locale {
    * It also allows skipping the script section of the id, so `en-US` is
    * properly parsed as `en-*-US-*`.
    */
-  constructor(locale, range = false) {
+  constructor(locale) {
     const result = localeRe.exec(locale.replace(/_/g, "-"));
     if (!result) {
+      this.isWellFormed = false;
       return;
     }
 
-    const missing = range ? "*" : undefined;
+    const language = result[1] || undefined;
+    const script = result[2] || undefined;
+    const region = result[3] || undefined;
+    const variant = result[4] || undefined;
 
-    const language = result[1] || missing;
-    const script = result[2] || missing;
-    const region = result[3] || missing;
-    const variant = result[4] || missing;
-
-    this.language = language;
-    this.script = script;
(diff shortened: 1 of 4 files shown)


zbraniecki:
it removes the `*` as an option for language tag, because we really never need it. Instead it uses ranges on `match` method which is what we do in C++ and Rust.

zbraniecki:
Fixes #334

fluent-langneg/src/locale.js
```diff
@@ -34,53 +34,63 @@ export default class Locale {
...
-  matches(locale) {
+  matches(locale, thisRange = false, otherRange = false) {
     return localeParts.every(part => {
-      return this[part] === "*" || locale[part] === "*" ||
-        (this[part] === undefined && locale[part] === undefined) ||
-        (this[part] !== undefined && locale[part] !== undefined &&
-        this[part].toLowerCase() === locale[part].toLowerCase());
+      return ((thisRange && this[part] === undefined) ||
+              (otherRange && locale[part] === undefined) ||
+              this[part] === locale[part]);
```
stasm: Understanding this method is important for understanding the whole algorithm and I'm not fond of adding two boolean arguments to it. It's hard to follow what they mean when the method is called. In fact, I think I largely preferred the old code which used an explicit `*` to represent ranges. 

A few suggestions on how to improve this code. I'm not saying all of them need to be applied at the same time, but I'd like to see this method improved _somehow_.

- Remove the `thisRange` param as it is always `true` across all callsites, or start using it for exact matching.
- Use an `options` dictionary: `locale.matches(other, {thisRange: true, otherRange: true})`.
- Separate this code into multiple methods: `exactMatches`, `matchesThisRange`, `matchesRanges`.
- Add a flag to `Locale` called `undefinedAsRange` and a method called `toRange()` which returns a new `Locale` instance based on the one the method was called on. `match` would inspect the value of the flag for both `this` and `other` and act accordingly.

Now, I understand that this code is based on the existing C++/Rust code. With that in mind, I think this can be moved to a follow-up. Please file it if you agree.

PR #342, opened 2019-02-19
Implement Syntax 0.9

stasm:
https://github.com/projectfluent/fluent/projects/6


fluent/test/values_ref_test.js
```diff
@@ -57,7 +57,8 @@ suite('Referencing values', function(){
     const msg = bundle.getMessage('ref2');
     const val = bundle.format(msg, args, errs);
     assert.equal(val, 'B2');
-    assert.equal(errs.length, 0);
+    // XXX Not an error; selectors may be missing in parameterized Terms.
```
stasm: I'll file a spec issue on this.

fluent-syntax/src/parser.js
```diff
@@ -383,22 +380,24 @@ export default class FluentParser {
...
-
     if (ps.currentChar === ".") {
-      num += ".";
       ps.next();
-      num = `${num}${this.getDigits(ps)}`;
+      fraction = this.getDigits(ps);
     }
 
-    return new AST.NumberLiteral(num);
+    let value = parseFloat(`${integer}.${fraction}`);
```
zbraniecki: I like to think about AST as a structured, semantic view of the source string.

From that perspective, what if we stored a number as a slice `[start;end]` and parsed it in the resolver instead?

My concern here is that storing numbers and parsing them differs between different programming languages Fluent may work in. Number is a low level primitive, but it still can differ in how it stores fractions, unsigned vs signed etc. By keeping AST free of that logic, we ensure that the AST is portable between the implementations irrelevant of how JS, Python or Rust treat numerical values.

Not a blocker, just a suggestion.

fluent-syntax/src/parser.js
```diff
@@ -766,14 +668,44 @@ export default class FluentParser {
...
+      return new AST.TermReference(id, attr, args);
     }
 
     if (ps.isIdentifierStart()) {
       const id = this.getIdentifier(ps);
-      return new AST.MessageReference(id);
+
+      if (ps.currentChar === "(") {
+        // It's a Function. Ensure it's all upper-case.
+        if (!/^[A-Z][A-Z_?-]*$/.test(id.name)) {
```
zbraniecki: this looks weird... I'd expect that function identifier be a subset of EBNF identifier which is `[a-zA-Z] [a-zA-Z0-9_-]*` and I'd expect the function to be `[A-Z] [A-Z0-9_-]*`

Instead you cut out numbers and added `?` which I think is a leak of your old idea to use `?` in function names. Can you clarify what's the spec here?

stasm: This wasn't intentional. I probably kept copying this code for the few last releases from one place to another. The spec has it as `const VALID_FUNCTION_NAME = /^[A-Z][A-Z0-9_?-]*$/;` which also shouldn't include the `?`. I'll fix it.

zbraniecki: oh, cool! thanks. It would be also nice to have those things tested in fixtures (`_`, `-` and `0-9`)

fluent-syntax/src/parser.js
stasm: To close the loop: this was fixed in the spec in https://github.com/projectfluent/fluent/issues/239. I've updated the PR to reflect it.

fluent-syntax/src/parser.js
stasm: I fixed this in the spec in https://github.com/projectfluent/fluent/issues/244 and I've since updated this PR.

PR #346, opened 2019-03-11
Update perf tests and unify them with fluent-rs benches

zbraniecki:
This is a refresh of our benchmarking suite. Changes:

* unifies the code across three runtimes into `benchmark.common`
* adds simple.ftl, menubar.ftl and preferences.ftl to match fluent-rs
* makes the test throw if errors are reported from parsing or resolving
* updates node impl to use `process.hrtime.bigint`

The last change, I recognize, is questionable one since it requires Node 10 to run, but tbh I doubt anyone uses our benchmark suite in node pre-10 and bigint gives better resolution with lower variance in my experience.

I kept the `workload-low` in the patch because it allows us to compare the pre-post patch numbers directly, but tbh menubar.ftl is very similar to workload-low, and menubar is the file I use in https://bugzilla.mozilla.org/show_bug.cgi?id=1501886 so I believe it's more realistic to use it here.

preferences.ftl are pulled from the current mozilla-central and reflect the biggest FTL file we use.

I like keeping both menubar and preferences because one is really big, while the other is directly on the startup path.

Hope you'll like the changes :)


tools/perf/benchmark.common.js
```diff
@@ -0,0 +1,79 @@
+
+function runTests(env) {
+  const testData = JSON.parse(env.readFile("data.json"));
+
+  const results = {};
+
+  for (let testName in testData) {
+    let test = testData[testName];
```
stasm: Would it make sense to split `data.json` into multiple JSONs, one for each benchmark? `simple.ftl` would go with `simple.json` etc. I think it would make it clearer what `data.json` actually is, and also easier to find out which variables are available in each benchmark.

zbraniecki: Maybe something like I suggested in https://github.com/projectfluent/fluent/issues/249 would work with that well? I'd prefer to make such changes as part of the reference fixtures, if that's ok with you.

stasm: I think that's a larger task, really. Splitting the jsons sounds simple enough that I'd be in favor of doing it in this PR.

zbraniecki: Adding this here is also not trivial. I need to find per-engine way of reading the content of a directory to collect the list of tests available. I'll try to see how much work it is

stasm: Ah, I see what you mean. This is both the storage of variable values _and_ the list of tests to run. I missed that before. Maybe renaming this file to `tests.json` would already be an improvement?

stasm: Or, `benchmarks.json`, or `testsuites.json`.. you get the idea. `data` is a bit cryptic :)

stasm:
Hey @zbraniecki, I'd like to report two follow-ups to this.

1. `make perf` doesn't work in node 8 nor 9 right now because it doesn't load the `PluralRules` polyfill. Please add `intl_pluralrules` to `benchmark.node.js`. You might need to set `ClobberIntlPluralRules = true` to make it also work for node 10 (and not print to stdout).

2. The reported times are largely dependent on the ordering of the benchmarks defined in `fixtures/benchmarks.json`. Here's what I currently see on my machine when `workload-low` is the last benchmark to run:
```
parse-syntax/"workload-low":
  mean:   4.13
  stdev:  2.11
  sample: 30
parse-runtime/"workload-low":
  mean:   0.95
  stdev:  0.43
  sample: 30
resolve-runtime/"workload-low":
  mean:   0.19
  stdev:  0.14
  sample: 30
```
And here are the results for it when it runs first:
```
parse-syntax/"workload-low":
  mean:   17.51
  stdev:  2.65
  sample: 30
parse-runtime/"workload-low":
  mean:   4.66
  stdev:  0.91
  sample: 30
resolve-runtime/"workload-low":
  mean:   3.28
  stdev:  0.71
  sample: 30
```

Do you know what might be causing this?

zbraniecki:
Oh, that makes total sense. Thank you! Will file a follow up PR.

PR #348, opened 2019-03-13
Ignore missing variables in term references 

stasm:


stasm:
> One concern I have is whether the migration for const {errors} = scope.errors to scope.errors.push has perf impact. Now, I know we have more questions about that, but I want to make sure you had an eye on that.

Thanks for asking. I didn't see any impact when I tested it earlier today. I used the existing perf testing code with the existing `workload-low` file and also with a manually copied `preferences.ftl` from #346.

Pike:
Cool, my r+ stands 🚢

PR #349, opened 2019-03-13
Fix stas feedback on perf tests

zbraniecki:
Stas identified two problems with my last change:

1) lack of `intl-pluralrules` broke testing on node v8/v9
2) all samples were run in a single instance which caused JIT to kick in

I reorganized test.js a bit to separate out each test sample to be called separately, and verified that reordering samples (or tests) does not impact the result.

### DIFF
--- a/tools/perf/benchmark.common.js
+++ b/tools/perf/benchmark.common.js
@@ -1,8 +1,8 @@
 function runTest(env) {
   const results = {};
   const testData = JSON.parse(env.readFile("./fixtures/benchmarks.json"));
-  const {args, functions} = testData[env.sampleName];
-  const ftlCode = env.readFile(`./fixtures/${env.sampleName}.ftl`);
+  const {args, functions} = testData[env.benchmarkName];
+  const ftlCode = env.readFile(`./fixtures/${env.benchmarkName}.ftl`);
 
   {
     const testName = "parse-syntax";
@@ -14,7 +14,7 @@ function runTest(env) {
       throw Error("Junk in syntax parser result!");
     }
 
-    results[`${testName}/"${env.sampleName}"`] = env.ms(end) - env.ms(start);
+    results[`${testName}/${env.benchmarkName}`] = env.ms(end) - env.ms(start);
   }
 
   let resource;
@@ -28,7 +28,7 @@ function runTest(env) {
     // we'll rely on the syntax parser to verify that
     // the sample test syntax is correct.
 
-    results[`${testName}/"${env.sampleName}"`] = env.ms(end) - env.ms(start);
+    results[`${testName}/${env.benchmarkName}`] = env.ms(end) - env.ms(start);
   }
 
   {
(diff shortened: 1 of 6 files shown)


tools/perf/benchmark.common.js
```diff
@@ -1,19 +1,8 @@
...
-    let test = testData[testName];
-    runTest(env, testName, test.args, test.functions, results);
-  }
-
-  return results;
-}
-
-function runTest(env, name, args, fncs, results) {
-  const ftlCode = env.readFile(`./fixtures/${name}.ftl`);
+  const testData = JSON.parse(env.readFile("./fixtures/benchmarks.json"));
```
stasm: Maybe an idea for a future improvement, not blocking this PR: could we read this file once and pass it to `runTest`?

tools/perf/benchmark.common.js
```diff
@@ -25,7 +14,7 @@ function runTest(env, name, args, fncs, results) {
       throw Error("Junk in syntax parser result!");
     }
 
-    results[`${testName}/"${name}"`] = env.ms(end) - env.ms(start);
+    results[`${testName}/"${env.sampleName}"`] = env.ms(end) - env.ms(start);
```
stasm: Since we're touching this code again, can we remove the quotes around the benchmark name?

tools/perf/test.js
```diff
@@ -14,7 +14,7 @@ program
   .usage('[options] [command]')
   .option('-e, --engine <string>', 'Engine to test: node, jsshell, d8 [node]',
           'node')
-  .option('-s, --sample <int>', 'Sample size [30]', 30)
+  .option('-s, --sampleSize <int>', 'Sample size [30]', 30)
```
stasm: Why change this? If needed, please use `--sample-size`.

tools/perf/test.js
```diff
@@ -47,47 +47,63 @@ function color(str, col) {
...
-        callback(results);
-      }
-    });
+runAll(parseInt(program.sampleSize)).then(printResults);
+
+async function runAll(sampleSize) {
+  const results = {};
+
+  const testData = JSON.parse(fs.readFileSync(`${__dirname}/fixtures/benchmarks.json`).toString());
+  for (let sampleName in testData) {
```
stasm: Would `benchmarkName` be a better name? We use `sample` in the statistical meaning in this file, too.

tools/perf/test.js
```diff
@@ -47,47 +47,63 @@ function color(str, col) {
...
+    if (!program.raw && program.progress) {
+      process.stdout.write(color(`\n${sampleName}: `, INDICATOR));
+    }
+    await runSample(sampleName, sampleSize, results);
```
stasm: Same here, maybe rename this to `runBenchmark?

tools/perf/benchmark.common.js
zbraniecki: I would prefer not to pass complex data such as args/callbacks via command line tbh.

tools/perf/test.js
zbraniecki: to separate sample name from sample size.

zbraniecki: reverted and renamed sampleName to benchmarkName

PR #360, opened 2019-04-26
FluentBundle.compound

stasm:
This is a continuation of the work started by @unclenachoduh in #322.

### DIFF
--- a/fluent/src/bundle.js
+++ b/fluent/src/bundle.js
@@ -181,34 +181,40 @@ export default class FluentBundle {
   /**
    * Format a message to a string or null.
    *
-   * Format a raw `message` from the bundle into a string (or a null if it has
-   * a null value).  `args` will be used to resolve references to variables
-   * passed as arguments to the translation.
+   * Find a message or an attribute by `path` in the bundle and format it into
+   * a string (or a null if it is a message and has a null value). `path` may
+   * be a simple message identifier (`foo`) or a path to an attribute using
+   * a dot as the separator (`foo.bar`). `args` will be used to resolve
+   * references to variables passed as arguments to the translation.
    *
    * In case of errors `format` will try to salvage as much of the translation
    * as possible and will still return a string.  For performance reasons, the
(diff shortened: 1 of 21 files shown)


fluent/src/bundle.js
```diff
@@ -64,15 +64,6 @@ export default class FluentBundle {
...
 
-  /*
-   * Return an iterator over public `[id, message]` pairs.
-   *
-   * @returns {Iterator}
-   */
-  get messages() {
-    return this._messages[Symbol.iterator]();
-  }
-
```
eemeli: It would be good to retain some way of iterating over a bundle's messages. Atm. e.g. `print()` in `tools/format.js` ends up needing to access `bundle._messages`.

How about something like this?

```js
get messageKeys() {
  return this._messages.keys();
}
```

stasm: I don't think we need to add this right now. I don't know of good runtime use-cases for iterating over all messages in a bundle. I usually prefer to keep the API surface as small as possible, and extend it only when needed.

The code in `tools/format.js` should use the result of the full-AST parsing from `fluent-syntax` to build an iterable of message keys for the purpose of printing all messages. I intend to fix it in a follow-up but I can also do it here.

fluent/src/bundle.js
```diff
@@ -194,50 +172,141 @@ export default class FluentBundle {
...
+   * @param   {?Object} args
+   * @param   {?Array} errors
    * @returns {?string}
    */
-  format(message, args, errors) {
-    // optimize entities which are simple strings with no attributes
-    if (typeof message === "string") {
-      return this._transform(message);
+  format(path, args, errors) {
+    let parts = path.split(".");
```
zbraniecki: you could just use `path.split(".", 1);` to separate one `.` and not bother with anything past that. The rest is an attribute name.

stasm:
We're going to the go with the `formatPattern` proposal in #380.

PR #366, opened 2019-04-29
Add FluentMessage

### LINKED ISSUE #365: Store messages as FluentMessage in the runtime
This is a proposal for a new API for formatting messages from bundles. It's an alternative to #208. It solves the same problems that #208 does, but in a way which doesn't regress the ability to retrieve a message representation without formatting it and cache it for subsequent formatting operations. See also #358.

### The current API (`fluent` 0.12)

```js
bundle.addMessages(`
foo = Foo
    .bar = Bar
`);
let message = bundle.getMessage("foo");
bundle.format(message, args, errs);
if (message.attrs && message.attrs["bar"]) {
    bundle.format(message.attrs["bar"], args, errs]);
}
```

This API exposes the internal representation of th e message to the consumer. The `message` might be a primitive string, an array, or an object of `{value,attrs}`. In the last case, `value` maybe `undefined`, a string, or an array; `attrs` may be `undefined` or an object of attributes whose values may again be strings or arrays.

This complexity comes from the days of Firefox OS where it was imperative to minimize the size of the JSON representing translations because the storage on low-end devices was limited.

### The `compound` API proposed in #208

```js
bundle.addMessages(`
foo = Foo
    .bar = Bar
`);

// The "pull" API.
bundle.format("foo", args, errs);
bundle.format("foo.bar", args, errs);

// The "push" API; returns a `{value, attributes}` shape.
bundle.compound("foo", args, errs);
```

In this API design, there are two entry points. The `format` method allows to specify the attribute to format after a dot. Each attribute formatted in this manner would cause the message to be re-retrieved by the bundle internally, which isn't optimal performance-wise.

The `compound` entry-point returns an object of formatted values and attributes. I'm calling it "push" because all attributes would be formatted at all times (which is probably what we want anyways for the most part). However, if a need to implement performant "pull" API arises, this method could evolve to return a lazy `attributes` field which would format attributes only when queried. This idea was the gist of my comment in https://github.com/projectfluent/fluent.js/issues/208#issuecomment-431294807.

This design hides the details of how messages are stored internally by the bundle from consumers, but it doesn't allow to retrieve a message once for later re-use when formatting the same UI widget multiple times.

### The `FluentMessage` API (this proposal)

In this proposal, I'd like to formalize the interface of `FluentMessage`, returned by `getMessage`. `FluentMessages` would provide methods for resolving their value and attributes, and a iterable over their attributes' names. Obviously, in current JS we can't really hide the fields used for storing the parsed value and attributes, but we can hint at their being private with the `_` prefix, and later on switch to proper private fields when they're standardized.

```js
bundle.addMessages(`
foo = Foo
    .bar = Bar
`);

// An atomic retrieve-and-format API for when attributes are not defined on the message.
// Good for quickly getting started with the bundle and for imperative callsites.
bundle.format("foo", args, errs);

// `message` can be cached for re-use. Attributes can 
let message = bundle.getMessage("foo");
bundle.formatValue(message, args, errs);

// The "pull" API. `name` is known up front.
bundle.formatAttribute(message, name, args, errs);

// The "push" API. `name` comes from the message itself.
for (let name of message.attributeNames) {
    bundle.formatAttribute(message, name, args, errs);
}
```

This API provides better encapsulation of the runtime AST inside of the `FluentMessage` class. It supports both the "pull" and the "push" API, and allows caching of the message for multiple formatting operations.

stasm:
Fix #365. Obsoletes #322 and #360 (on which it's based).


fluent/src/bundle.js
```diff
@@ -62,15 +62,7 @@ export default class FluentBundle {
...
-  }
-
-  /*
-   * Return an iterator over public `[id, message]` pairs.
-   *
-   * @returns {Iterator}
-   */
-  get messages() {
-    return this._messages[Symbol.iterator]();
+    this._dirty = new WeakSet();
```
zbraniecki: what's that? It doesn't seem to be used anywhere

fluent/src/bundle.js
```diff
@@ -170,74 +153,124 @@ export default class FluentBundle {
...
    * @returns {?string}
    */
-  format(message, args, errors) {
-    // optimize entities which are simple strings with no attributes
-    if (typeof message === "string") {
-      return this._transform(message);
+  format(id, args, errors = []) {
+    if (!this._messages.has(id)) {
+      errors.push(`Message not found: "${id}"`);
+      return undefined;
```
zbraniecki: Why `undefined` rather than `null`? Is the user expected to always check if the message exists before trying to format it? I thought `format("non-existing") -> null` is a reasonable API use and not an error?

fluent/src/bundle.js
```diff
@@ -170,74 +153,124 @@ export default class FluentBundle {
...
+  }
 
-    return resolve(this, args, message, errors);
+  _createScope(args, errors = []) {
+    return {
+      args, errors,
+      bundle: this,
+      dirty: this._dirty,
+      // TermReferences are resolved in a new scope.
+      insideTermReference: false,
```
zbraniecki: I struggled with that in Rust because of lifetimes. While the exact problem is very rust specific and doesn't warrant design changes, I kinda like what I ended up with - `locale_args` in the scope - https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/src/resolve.rs#L244

Maybe worth considering here? Not a blocker of course.

fluent/src/resource.js
```diff
@@ -55,7 +56,7 @@ const MAX_PLACEABLES = 100;
 /**
  * Fluent Resource is a structure storing a map of parsed localization entries.
  */
-export default class FluentResource extends Map {
+export default class FluentResource extends Set {
```
zbraniecki: Beyond updating the comment just a line above, can we consider Array vs. Set here? We won't likely be looking for duplicate-free set, so why not Array?

fluent/src/bundle.js
stasm: It's used in `createScope`, and then in `Pattern` to detect cyclic references.

fluent/src/bundle.js
stasm: I decided to used `undefined` here and when formatting missing attributes in order to tell missing values from null values apart. Consider:

```properties
foo =
    .attr = Attr
```

- `format("foo")` → `null`,
- `format("bar")` → `undefined`.

Do you think they should both return `null` instead?

fluent/src/bundle.js
stasm: Nice! So `local_args` is both a store of arguments passed to the term _and_ a flag meaning _we're resolving a term now_. I like it!

fluent/src/resource.js
stasm: Good idea, let me test it with `make perf`.

fluent/src/bundle.js
zbraniecki: That's an interesting question. I'm torn. For me the question boils down to - should attempting to format a message that doesn't existing in the bundle be an error (which means, you should always call `hasMessage` before you call format)?

fluent/src/resource.js
stasm: Sweet, extending `Array` gives a slight speed up in `addResource`. I guess iterating over the elements of an array is faster than of a set. Thanks for the suggestion!

zbraniecki: well, Set does more - it has to hash and look for duplicates on insertion.

stasm:
Here's the output of `make perf` compared to `master`:

```
parse-runtime/simple:
  mean:   2.46 (+3%)
  stdev:  0.49
  sample: 30
create-runtime/simple:
  mean:   0.44 (-54%)
  stdev:  0.12
  sample: 30
resolve-runtime/simple:
  mean:   1.15 (+35%)
  stdev:  0.53
  sample: 30
parse-runtime/preferences:
  mean:   8.36 (-1%)
  stdev:  1.33
  sample: 30
create-runtime/preferences:
  mean:   0.74 (-43%)
  stdev:  0.15
  sample: 30
resolve-runtime/preferences:
  mean:   9.1 (+11%)
  stdev:  1.57
  sample: 30
parse-runtime/menubar:
  mean:   4.95 (+13%)
  stdev:  0.66
  sample: 30
create-runtime/menubar:
  mean:   0.53 (-46%)
  stdev:  0.17
  sample: 30
resolve-runtime/menubar:
  mean:   2.36 (+15%)
  stdev:  0.51
  sample: 30
parse-runtime/workload-low:
  mean:   4.99 (-5%)
  stdev:  0.78
  sample: 30
create-runtime/workload-low:
  mean:   0.46 (-57%)
  stdev:  0.06
  sample: 30
resolve-runtime/workload-low:
  mean:   1.82 (+10%)
  stdev:  0.34
  sample: 30
```

I think `create-runtime` (a benchmark which I added when working on this PR; it tests the bundle creation and `addResource`) is faster mostly because `FluentResources` aren't subclasses of `Map` anymore. I was expecting parsing to go up a bit because it now creates `FluentMessage` instances inside of the `FluentResource`, but there aren't any significant changes.

We're paying more for resolving and I think it's the price of the encapsulation: `bundle.formatValue` calls `createScope`, then `message.resolveValue`, which calls `Type`. Only then do we check if the value is a simple string and return it. It happens later than on `master` which—because no encapsulation is enforced—checks the type of the message right away inside of `bundle.format`. However, it's still not bad (up to +35% on `simple.ftl`) and combined with the speed-ups in `addResource`, a one-time create-parse-format cycle is generally faster, or at least the same, in this PR and on `master`. For the aforementioned `simple.ftl`, it takes 4.18 ms on my laptop on `master`, and `4.05` with this PR.

zbraniecki:
It would be nice to have a `create-parse-format` cycle perf end-to-end test :)

fluent/src/bundle.js
stasm: I think the bindings should always check, yes. They should also use `formatValue` and `formatAttribute`, rather than `format`. `format` is intended for more imperative approaches. The fact that it returns `undefined` on missing messages was intended to be consistent with `formatAttribute` returning `undefined` on missing attributes.

It also works well semantically, I think. The difference between `null` and `undefined` maps nicely to Fluent's notions of value-less messages vs. missing messages. Attempting to format a missing message should result in an error being reported. As a developer, there's little incentive for me to intentionally format missing messages :) So it's likely an error scenario, which I'd like to be able to react to.

fluent/src/resource.js
stasm: I know, but interestingly, there's no noticeable difference on the benchmark which inserts. Instead, it's the one which iterates over the elements which sped up :)

stasm:
We're going to the go with the `formatPattern` proposal in #380.

PR #368, opened 2019-04-30
Print braces around missing references

### LINKED ISSUE #367: Print expression syntax when references fail
Right now, when a reference fails to resolve, the resolver falls back to the name of the broken variable or message:

    ⁨$username⁩ ⁨added ⁨$photocount⁩ new photos⁩ to ⁨their stream⁩.

I'd like to add braces to this output to make it clearer that it comes from a broken syntax:
```
{$username⁩} ⁨added ⁨{$photocount⁩} new photos⁩ to ⁨their stream⁩.
```

stasm:
Fix #367.

This also fixes `NUMBER` and `DATETIME` to handle other types of arguments than numbers and dates, respectively. The fallback could be even better in those cases, but this PR is already an improvement, and I'd like to land it soon rather than wait for the spec to be ready.

### DIFF
--- a/fluent/src/builtins.js
+++ b/fluent/src/builtins.js
@@ -11,14 +11,7 @@
  * `FluentType`.  Functions must return `FluentType` objects as well.
  */
 
-import { FluentNumber, FluentDateTime } from "./types.js";
-
-export default {
-  "NUMBER": ([arg], opts) =>
-    new FluentNumber(arg.valueOf(), merge(arg.opts, opts)),
-  "DATETIME": ([arg], opts) =>
-    new FluentDateTime(arg.valueOf(), merge(arg.opts, opts)),
-};
+import { FluentNone, FluentNumber, FluentDateTime } from "./types.js";
 
 function merge(argopts, opts) {
   return Object.assign({}, argopts, values(opts));
@@ -30,4 +23,26 @@ function values(opts) {
     unwrapped[name] = opt.valueOf();
   }
   return unwrapped;
+}
+
+export
+function NUMBER([arg], opts) {
+  if (arg instanceof FluentNone) {
+    return arg;
+  }
+  if (arg instanceof FluentNumber) {
+    return new FluentNumber(arg.valueOf(), merge(arg.opts, opts));
+  }
+  return new FluentNone("NUMBER()");
+}
+
+export
+function DATETIME([arg], opts) {
+  if (arg instanceof FluentNone) {
+    return arg;
+  }
+  if (arg instanceof FluentDateTime) {
+    return new FluentDateTime(arg.valueOf(), merge(arg.opts, opts));
+  }
+  return new FluentNone("DATETIME()");
 }
--- a/fluent/src/types.js
+++ b/fluent/src/types.js
@@ -45,8 +45,12 @@ export class FluentType {
 }
 
 export class FluentNone extends FluentType {
+  valueOf() {
+    return null;
+  }
+
   toString() {
-    return this.value || "???";
+    return `{${this.value || "???"}}`;
   }
 }
 
--- a/fluent/src/resolver.js
+++ b/fluent/src/resolver.js
@@ -37,7 +37,7 @@
 
 import { FluentType, FluentNone, FluentNumber, FluentDateTime }
   from "./types.js";
-import builtins from "./builtins.js";
+import * as builtins from "./builtins.js";
 
 // Prevent expansion of too long placeables.
 const MAX_PLACEABLE_LENGTH = 2500;
--- a/fluent/test/arguments_test.js
+++ b/fluent/test/arguments_test.js
@@ -101,49 +101,49 @@ suite('Variables', function() {
     test('falls back to argument\'s name if it\'s missing', function() {
       const msg = bundle.getMessage('foo');
       const val = bundle.format(msg, {}, errs);
-      assert.equal(val, '$arg');
+      assert.equal(val, '{$arg}');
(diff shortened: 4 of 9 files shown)


fluent/src/builtins.js
```diff
@@ -31,3 +24,25 @@ function values(opts) {
...
+
+export
+function NUMBER([arg], opts) {
+  if (arg instanceof FluentNone) {
+    return arg;
+  }
+  if (arg instanceof FluentNumber) {
+    return new FluentNumber(arg.valueOf(), merge(arg.opts, opts));
+  }
+  return new FluentNone("NUMBER()");
```
zbraniecki: What's the intention for returning `NUMBER()` as the fallback string? I'm concerned that it locks down how we display errors based on something slightly misleading - if I have `key = Foo { NUMBER($emailCount, minimumFractionDigits: 4) }` then displaying `Foo [NUMBER()]` may be misleading.

On top of that, from the design perspective, in `fluent-rs` I leaned on the side of keeping display formatting located in one place, rather than each function returning its own representation.

So, to translate it to JS, I'd expect this to return sth like `FluentError({type: "function", name: "NUMBER"})` and optionally `args` so that some `toString` of `FluentError` can construct the visual representation, for example `NUMBER()` like here, or something else.

The added benefit is that since you're passing semantic information rather than stringified representation, you can differentiate how you show the `FluentError` between what you send to console, and what you display in the formatter partial string.

It's not a blocker, just a suggestion.

fluent/src/types.js
```diff
@@ -45,8 +45,12 @@ export class FluentType {
...
 export class FluentNone extends FluentType {
+  valueOf() {
+    return null;
+  }
+
   toString() {
-    return this.value || "???";
+    return `{${this.value || "???"}}`;
   }
 }
```
zbraniecki: For what I mentioned earlier, I ended up with https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/src/types.rs#L23-L70 and in result I'm down to one case where I actually have something that will produce `???` and it's the scenario where the default variant is missing - https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/src/types.rs#L149-L157 and https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/src/resolve.rs#L206

I hope to move away from that as well and remove `FluentValue::None` completely.

I think it's a cleaner design with easier maintenance and ability to further improve UX of handling error scenarios.

zbraniecki: It also improved the experience when the error is `key0 = Foo { key1.attr }` to avoid formatting it into `Foo { key1 }` in the error scenario - https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/tests/resolve_attribute_expression.rs#L44-L58

stasm: > I hope to move away from that as well and remove FluentValue::None completely.

But you'll still have `FluentValue::Error`, right? There needs to be something for storing a failed resolution. In `fluent.js` it's `FluentNone` because error reporting is not very advanced right now. In this PR, I only want to improve the output of the fallback. Error reporting is a much larger task in scope and I'd prefer to first design the spec for it.

stasm: > It also improved the experience when the error is key0 = Foo { key1.attr } to avoid formatting it into Foo { key1 } in the error scenario - https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/tests/resolve_attribute_expression.rs#L44-L58

I have a branch which improves this, too :) I'll open a PR later today.

fluent/src/builtins.js
stasm: > What's the intention for returning NUMBER() as the fallback string?

This happens when the arg resolves cleanly, but it's not the right type, e.g. `NUMBER($userName)`. Right now this prints `NaN` which I'd almost consider a bug. In this PR, I'm changing it to `NUMBER()`. I think ideally the `$userName` reference would store both the resolved value of the variable _and_ its name, so that when the `NUMBER` functions fails, the output can use the name for fallback: `$userName` or `NUMBER($userName)`.

I think this again would benefit from the resolver spec, so in this PR, I'd just like to fix the `NaN` thing.

stasm: I think that in the long run, we'll add more information to `FluentType` instances, including the detailed error information in `FluentNone`. The `fluent-rs` examples look like something we could base the design on.

stasm:
> I like the direction. I left some comments about the implementation based on how I structured the error handling in Rust, but if you think it's not worth refactoring it any further right now, or that my design decisions are not good, I'm happy to revisit it to align with JS implementation.

The `fluent-rs` code is well designed. I just don't want to change it all at once here. The goal of this PR is to make an incremental improvement to how unresolvable expressions are printed when interpolated.

> Would You have {$unreadEmails} unread emails look better or You have [$unreadEmails] unread emails? I'm tempted to say that the latter is more clearly a message with errors, rather than FTL string, but I'm torn on whether it's a good or a bad thing.

I'm open to discussing this more. My initial reaction is that the `[...]` syntax also has a well-defined meaning in Fluent, and ` You have [$unreadEmails] unread emails?` would also constitute re-use. Sticking to curly braces seems like the safest bet right now.

zbraniecki:
> My initial reaction is that the [...] syntax also has a well-defined meaning in Fluent, and You have [$unreadEmails] unread emails? would also constitute re-use. Sticking to curly braces seems like the safest bet right now.

Convinced.

fluent/src/types.js
stasm: See https://github.com/projectfluent/fluent.js/pull/370.

PR #369, opened 2019-04-30
ftl dedenter: Don't remove blank lines

### LINKED ISSUE #362: ftl template literal tag removes blank lines inside multiline text
Setup bundle like:
```
const test = new FluentBundle("de-DE");
test.addMessages(ftl`
    blank-lines =

        The blank line above this line is ignored.
        This is a second line of the value.

        The blank line above this line is preserved.
`);
```
Calling `test.getMessage("blank-lines")` will return:
```
[
  "The blank line above this line is ignored.",
  "↵",
  "This is a second line of the value.",
  "↵",
  "The blank line above this line is preserved."
]
```
Without `ftl` and dedenting manually it will return:
```
[
  "The blank line above this line is ignored.",
  "↵",
  "This is a second line of the value.",
  "↵↵",
  "The blank line above this line is preserved."
]
```
Notice that the second time it returns two newlines at index 3: `"↵↵"`

stasm:
Fix #362.

### DIFF
--- a/fluent/src/util.js
+++ b/fluent/src/util.js
@@ -1,27 +1,22 @@
-function nonBlank(line) {
-  return !/^\s*$/.test(line);
-}
-
-function countIndent(line) {
-  const [indent] = line.match(/^\s*/);
-  return indent.length;
-}
+const RE_NOT_BLANK = /\S/;
+const RE_INDENTION = /^\s*/;
 
 /**
  * Template literal tag for dedenting FTL code.
  *
- * Strip the common indent of non-blank lines. Remove blank lines.
+ * Calculate the common indent of non-blank lines. Then, trim it from the front
+ * of all lines. Blank lines with whitespace characters in them are trimmed
+ * too. If they're shorter than the common indent, only the line break is kept.
  *
  * @param {Array<string>} strings
+ * @returns string
  */
-export function ftl(strings) {
-  const [code] = strings;
-  const lines = code.split("\n").filter(nonBlank);
-  const indents = lines.map(countIndent);
-  const common = Math.min(...indents);
-  const indent = new RegExp(`^\\s{${common}}`);
-
-  return lines.map(
-    line => line.replace(indent, "")
-  ).join("\n");
+export function ftl([code]) {
+  let indents = code
+    .split("\n")
+    .filter(line => RE_NOT_BLANK.test(line))
+    .map(line => RE_INDENTION.exec(line)[0].length);
+  let commonLength = Math.min(...indents);
+  let commonIndent = new RegExp(`^[ \\t]{0,${commonLength}}`, "gm");
+  return code.replace(commonIndent, "");
 }
--- a/fluent/test/util_test.js
+++ b/fluent/test/util_test.js
@@ -16,35 +16,26 @@ suite('ftl dedent helper', function() {
     test('start with EOL, no indent, end without EOL', function() {
       assert.equal(
         ftl`
-  foo = Foo`,
-        'foo = Foo'
+foo = Foo`,
+        '\nfoo = Foo'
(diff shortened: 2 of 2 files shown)


fluent/src/util.js
```diff
@@ -1,27 +1,22 @@
-function nonBlank(line) {
-  return !/^\s*$/.test(line);
-}
-
-function countIndent(line) {
-  const [indent] = line.match(/^\s*/);
-  return indent.length;
-}
+const RE_NOT_BLANK = /\S/;
+const RE_INDENTION = /^\s*/;
```
Pike: I think both of these regexes should be clear that indention is just ' ', and not the rest of whitespace, `[ \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]` (says MDN)

fluent/src/util.js
```diff
@@ -1,27 +1,22 @@
...
-  return lines.map(
-    line => line.replace(indent, "")
-  ).join("\n");
+export function ftl([code]) {
+  let indents = code
+    .split("\n")
+    .filter(line => RE_NOT_BLANK.test(line))
+    .map(line => RE_INDENTION.exec(line)[0].length);
+  let commonLength = Math.min(...indents);
+  let commonIndent = new RegExp(`^[ \\t]{0,${commonLength}}`, "gm");
```
Pike: I don't think tabs should indent here.

stasm:
> We only indent on ' ', right? 

We do (i.e. Fluent) but JavaScript doesn't :) If someone uses tabs to indent JS, they would expect `ftl` to work in their code, too, I think. We could reduce the scope of the helper, and explicitly un-support tabs, too.

stasm:
> but I don't recall what the outcome or rationale was

I'm no sure if I understand what the question is :) Are you asking about the `\` escaping the new line?

fluent/src/util.js
stasm: Hmm, I'm not sure how this should work exactly. The non-blank test is intended to filter out empty lines which would otherwise set the `commonLength` to `0`. Perhaps it would be OK to only support fully empty blank lines in this scenario? In other words, if a blank line has a few spaces in it, the `ftl` dedenter would not filter it out and take its indent into account when calculating the common indent?

stasm:
Thank you for the feedback, @Pike, great points. The dedentation logic is surprisingly tricky :) I wonder if the solution to the points you raised is a "product" decision, rather than code fixes.

I see a number of paths forward. Let's leave this PR open until I'm back and decide then.

- Remove the helper entirely. Fluent examples wouldn't look as nice in the source code, but all ambiguity would be removed.
- Support only spaces in source code's indentation.
- Filter out only empty blank lines, rather than any blank lines, including lines with inline whitespace in them, as it's done right now.
- Change the API of the helper to something more Swift-like, where the indent of the last line before the closing tick defines what will be stripped from the beginning of each line.

stasm:
I came back to this after vacation, and I realized that the less magic there is in the `ftl` helper, the easier it will be to explain and understand what it does. I considered different options for changing the behavior of the helper to make its output more predictable. At first, I liked the _Filter out only empty blank lines_ solution from above. I quickly realized that in most cases it was equivalent to Swift's approach; the last line is usually indented, i.e. it's not empty, and it contributes to the calculation of the common indent. However, there were edge-cases where this wasn't the case. For instance, putting the closing delimiter as the first character of a new line would make the last line empty, and thus filtered out, which means it isn't taken into account in this particular scenario. The more I thought about it, the more straight-forward Swift'a approach appeared to be, and in the end I decided to implement it in full in #374 as a separate `@fluent/dedent` package.

PR #374, opened 2019-05-22
Add the fluent-dedent package

### LINKED ISSUE #362: ftl template literal tag removes blank lines inside multiline text

stasm:
Fix #362. Supersedes #369.

This PR adds the `@fluent/dedent` package which only provides the `ftl` template literal tag. Once it's merged and published, I'll remove the `ftl` helper from the `fluent` package. By moving the helper to its own package the refinements to it and changes to the behavior will not require version bumps of the main `fluent` package.

I rewrote the whole thing. The behavior is now based on the [multiline strings spec in Swift](https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html), where the indent of the closing delimiter defines the amount of indent removed from other lines. The first and the last lines are removed from the output, too. If they're not all-whitespace, or if the indentation is insufficient in any of the content lines, a `RangeError` is thrown.

```javascript
import ftl from "@fluent/dedent";
let messages = ftl`
    hello = Hello, world!
    welcome = Welcome, {$userName}!
    `;

assert(messages === "hello = Hello, world!\nwelcome = Welcome, {$userName}!");
```

I considered other options, but this one appealed to me the most with its simplicity and predictability. It's easy to explain in a single sentence, and visually too.

### DIFF
--- a/fluent-dedent/src/index.js
+++ b/fluent-dedent/src/index.js
@@ -35,7 +35,7 @@ export default function ftl(strings, ...values) {
     }
     if (lineIndent !== commonIndent) {
       // The indentation of the line must match commonIndent exacty.
-      throw new RangeError(`Insufficient indentation in line ${idx}.`);
+      throw new RangeError(`Insufficient indentation in line ${idx + 1}.`);
     }
     // Strip commonIndent.
     return line.slice(commonIndent.length);
--- a/fluent-dedent/src/index.js
+++ b/fluent-dedent/src/index.js
@@ -1,0 +1,36 @@
+// A blank line may contain spaces and tabs.
+const RE_BLANK = /^[ \t]*$/;
+
+/**
+ * Template literal tag for dedenting Fluent code.
+ *
+ * Strip the indent of the last line from each line of the input. Remove the
+ * first and the last line from the output. The snippet must start on a new
+ * line and it must end on a line of its own, with the closing delimiter on a
+ * next line.
+ *
+ * @param {Array<string>} strings
+ * @returns string
+ */
+export default function ftl(strings) {
+  let lines = strings[0].split("\n");
+  let [first, last] = [lines.shift(), lines.pop()];
+
+  if (!RE_BLANK.test(first)) {
+    throw new RangeError("Content must start on a new line.");
+  }
+  if (!RE_BLANK.test(last)) {
+    throw new RangeError("Closing delimiter must appear on a new line.");
+  }
+
+  let commonIndent = last.length;
+  function dedent(line, idx) {
+    let indent = line.slice(0, commonIndent);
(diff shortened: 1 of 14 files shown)


fluent-dedent/src/index.js
```diff
@@ -0,0 +1,36 @@
...
+ * Strip the indent of the last line from each line of the input. Remove the
+ * first and the last line from the output. The snippet must start on a new
+ * line and it must end on a line of its own, with the closing delimiter on a
+ * next line.
+ *
+ * @param {Array<string>} strings
+ * @returns string
+ */
+export default function ftl(strings) {
+  let lines = strings[0].split("\n");
```
Pike: Should this also test for extra arguments?

stasm: Yes, it looks like I forgot to push the second commit which adds this :) I've updated the PR just now.

stasm:
Thanks, @Pike.

> There's the ambiguity around ' ' and '\t', that is, a space is a tab and vice versa in indent.

Do you mean that we should strip exactly the same sequence of whitespace characters fro each line of content, rather than a given _amount_ of them? That's a good point, I'll change it.

fluent-dedent/src/index.js
```diff
@@ -0,0 +1,45 @@
...
+    let lineIndent = line.slice(0, commonIndent.length);
+    if (lineIndent.length === 0) {
+      // Empty blank lines are preserved even if technically they are not
+      // indented at all. This also short-circuits the dedentation logic when
+      // commonIndent.length is 0, i.e. when all indents should be kept.
+      return line;
+    }
+    if (lineIndent !== commonIndent) {
+      // The indentation of the line must match commonIndent exacty.
+      throw new RangeError(`Insufficient indentation in line ${idx}.`);
```
Pike: This is just personal taste, I prefer human-readable positions to be 1-based, this is zero-based.

stasm: Sounds good to me, thanks!

PR #378, opened 2019-06-20
Default the child element of Localized to null

### LINKED ISSUE #377: <Localized /> fallback case throws react error if no children are provided
If you use `<Localized />` and it has to revert to the fallback (such as when the `id` was not found in the bundles), `undefined` is returned from the component. This causes a somewhat unhelpful error message:

> Uncaught Invariant Violation: Localized(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.

flut1:
If no child element is provided and `<Localized />` tries to return the children as a fallback, `undefined` is returned. This causes a React error that is difficult to track down. Instead, default the child element to `null`, which is a valid return value for a React component.

Fixes #377

### DIFF
--- a/fluent-react/src/localized.js
+++ b/fluent-react/src/localized.js
@@ -79,7 +79,7 @@ export default class Localized extends Component {
 
   render() {
     const { l10n, parseMarkup } = this.context;
-    const { id, attrs, children: elem } = this.props;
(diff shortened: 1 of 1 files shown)


stasm:
Thanks, @flut1. I'll publish a new version of `fluent-react` this week.

Issue #376, opened 2019-05-28
Migrate to TypeScript?


stasm:
Over the last month I've used TypeScript in a side-project, and it's been a delight. It's quick to pick up, doesn't get into the way, has excellent editor support in VS Code, and compiles to readable JS. I'd like to kick off a discussion about migrating the `fluent.js` family of packages to TypeScript,. We could start with the more user-facing ones: `fluent`, `fluent-react`, and `fluent-langneg`, which also have the benefit of being small, well tested, and of already having [user-contributed typings](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32235) thanks to @huy-nguyen.

See also: https://github.com/projectfluent/fluent.js/issues/329, https://github.com/projectfluent/fluent.js/pull/231.

#### Update (January 2020)

The migration is currently ongoing. I'll keep the following list up to date as more packages are migrated.

- [x] `@fluent/bundle`: #436, v0.15.0
- [x] `@fluent/dedent`: #451 v0.2.0
- [x] `@fluent/dom`: #474 
- [x] `@fluent/langneg`: #462 v0.4.0
- [x] `@fluent/react`: #458 v0.12.0
- [x] `@fluent/sequence`: #450 v0.5.0
- [x] `@fluent/syntax`: #457 v0.15.0

huy-nguyen:
I'm willing to help out if need be.

Pike:
FWIW, VS Code might be the only acceptable editor for typescript. I tried something small in Atom over the weekend, and it was very Rust, with the recommended add-on. No help, bustages on compile, unhelpful error messages. In VS Code, the experience was pretty different. But even then, there were warnings arount tslint that couldn't be fixed following the instructions. 

I'd assume that typescript comes with quite a vendor lock in.

Gregoor:
IntelliJ also works great for TS. Afaik lots of TS IDE functionality comes from its language server, so it should behave pretty similar between editors.
Though I'd agree that the error messages could be better, they can be pretty verbose.

Seally:
I'd be happy if this project moves to TypeScript as well. I'm thankful the TypeScript definitions exist (and it is usable), but I feel it's not quite "there" yet and there might not be enough eyes looking at it to keep up with all the changes and edge cases.

As for editors, Atom seems to have the `atom-typescript` plugin, but I don't know how the experience there compares to VS Code and Intellij/Webstorm.

birtles:
[...]

Issue #358, opened 2019-04-22
Document/allow attribute access from a FluentBundle


eemeli:
See also #208, #282

At the moment, the internal structure of a Message is not public:
https://github.com/projectfluent/fluent.js/blob/326b6ee0667fb877ae19f134485c1716de97ec69/fluent/src/bundle.js#L86-L97

Despite this, it's the interface that at least `fluent-dom` and `fluent-react` use to access message attributes. This makes things difficult for external developers, as `fluent` currently does not provide any other means of accessing them. Is the plan to solve this to wait for #322 to land?

stasm:
Yeah, we should definitely stop doing this in the bindings packages. And you're right: #322 intends to rectify the situation. I didn't have time to work on it before the Syntax 1.0 release, but it's now a top priority for me. Do you think it's the right fix?

eemeli:
Yeah, I at least like the new `compound()` and `format()`, and dropping `getMessage()`. I'll add a comment or two to the PR with a couple of more specific concerns.

stasm:
@eemeli Would you consider this fixed by the `formatPattern` API?

eemeli:
Yes! 🎉

PR #380, opened 2019-06-26
FluentBundle.formatPattern

stasm:
Rather than hide the runtime representation of messages, let's document it and commit to only changing it in major version releases.

- `FluentResource` now derives from `Array` rather than a `Set`. I've noticed a perf win related to this change.
- Messages and terms are now always stored as a `{value, attributes:{}}` object. The value maybe null or may be a pattern (more on that below). `attributes` is an `Object.create(null)` whose keys are attribute names and values are patterns.
- Patterns represented either as primitive strings (when they don't have any placeable) or as arrays of primitive strings or expressions.
- Expressions are stored as `{type, ...}` objects. The exact fields depend on the expression.
- The mindset of formatting shifts from _formatting a message_ or _formatting an attribute_ to simply: _formatting a pattern_. To make it explicit, I renamed `FluentBundle.format` to `.formatPattern`.
- To format a message with attributes, something like the following should be used:

    ```js
    let message = bundle.getMessage(id);
    if (message.value) {
        bundle.formatPattern(message.value, args, errors);
    }
    for (let [name, value] of Object.entries(message.attributes)) {
        bundle.formatPattern(value, args, errors);
    }
    ```

    This pattern supports all possible methods of iterating over the keys of an object, as well as formatting only a set of known attributes instead of all of them.
- I still need to figure out how to document or even formally describe the now-public data representation of messages. Perhaps this could be left as a follow up and taken care of together with or after #376, if we decide to do it.


stasm:
This PR could be viewed as a fix to #358.

stasm:
The performance looks good, with a significant win related to `FluentResource` now extending `Array`. Here's `make perf-compare-jsshell` against `master`:

```
parse-runtime/simple:
  mean:   4.82 (+0%)
  stdev:  0.47
  sample: 30
create-runtime/simple:
  mean:   0.74 (-60%)
  stdev:  0.07
  sample: 30
resolve-runtime/simple:
  mean:   1.55 (-14%)
  stdev:  0.21
  sample: 30

parse-runtime/preferences:
  mean:   18.3 (-2%)
  stdev:  3.17
  sample: 30
create-runtime/preferences:
  mean:   1.19 (-63%)
  stdev:  0.28
  sample: 30
resolve-runtime/preferences:
  mean:   18.95 (+2%)
  stdev:  1.95
  sample: 30


parse-runtime/menubar:
  mean:   9.59 (-7%)
  stdev:  1.04
  sample: 30
create-runtime/menubar:
  mean:   0.85 (-56%)
  stdev:  0.17
  sample: 30
resolve-runtime/menubar:
  mean:   3.7 (-13%)
  stdev:  0.67
  sample: 30


parse-runtime/workload-low:
  mean:   10.09 (0%)
  stdev:  1.21
  sample: 30
create-runtime/workload-low:
  mean:   0.87 (-56%)
  stdev:  0.07
  sample: 30
resolve-runtime/workload-low:
  mean:   3.25 (+2%)
  stdev:  0.32
  sample: 30

```

eemeli:
With this design, how deep does the definition go? Specifically, could/should a pattern be considered an opaque blob, or is there some benefit in defining e.g. expressions as well?

stasm:
I'd like to document all of the runtime AST, including expressions. I would not, however, expose additional methods like `formatExpression`. I think treating patterns as the only entry points to translations should be sufficient.

eemeli:
I would prefer patterns being left opaque, or at least documented as internal to the implementation. The output of `fluent-compiler` is a bit different from that level down, and needing to match the runtime API for the patterns and potentially expressions as well would be rather tedious, especially if there aren't any real use cases for them outside of `formatPattern`.

stasm:
Would you say that the goal of `fluent-compiler` is to be interoperable wrt. the runtime AST with `fluent` (soon to be called `@fluent/bundle`, so I'll use that as the name)? I think it would be nice if it was a drop-in _replacement_ for `@fluent/bundle`, implementing the same API methods. At the same time, I wouldn't expect a `message` retrieved from `fluent-compiler` to work in `@fluent/bundle` and vice versa. Is this something that would be helpful?

(I can see how the `format`/`compound` API sidesteps this problem, BTW.)

When I say _documented and public_, I don't mean _part of the Fluent specification_. I only propose that the shape of the runtime AST be documented as part of `@fluent/bundle`'s implementation, and any changes to it would require major version bumps of `@fluent/bundle`.

eemeli:
My intent is to get the compiler to a state where tools that are built on top of `@fluent/bundle` could use the output of the compiler as a drop-in replacement, so that the `Resource` and `Bundle` of both implementations would provide the same APIs. I don't mean that you'd be able to take a `Resource` from one and plug it into a `Bundle` of another.  Hence the desire to limit the depth into patterns/expressions that might be included in the published API, as that limits what the compiler would need to do as well.

Maybe another way of expressing this would be to ask whether anything else except for `Bundle#formatPattern()` would be expected to care about the shape or internals of a pattern?

zbraniecki:
> Maybe another way of expressing this would be to ask whether anything else except for Bundle#formatPattern() would be expected to care about the shape or internals of a pattern?

I second this. While I'm not strictly opposed to Stas' approach, I think that specifying the AST as public and stable is a major effort for very symbolic gain.
I'd rather keep it as private.

stasm:
> (@eemeli) My intent is to get the compiler to a state where tools that are built on top of `@fluent/bundle` could use the output of the compiler as a drop-in replacement, so that the `Resource` and `Bundle` of both implementations would provide the same APIs. I don't mean that you'd be able to take a `Resource` from one and plug it into a `Bundle` of another.

This sounds great, thank you for clarifying.

> Hence the desire to limit the depth into patterns/expressions that might be included in the published API, as that limits what the compiler would need to do as well.

I see what you mean. If all of the runtime AST it is public, than the criteria for creating drop-in replacements widen.

> Maybe another way of expressing this would be to ask whether anything else except for `Bundle#formatPattern()` would be expected to care about the shape or internals of a pattern?

Thanks for rephrasing. In the `formatPattern` proposal it has been my intent that only `formatPattern` cares and works with the internals of a pattern.

> (@zbraniecki) I second this. While I'm not strictly opposed to Stas' approach, I think that specifying the AST as public and stable is a major effort for very symbolic gain. I'd rather keep it as private.

I'd like for it to be private too, but the reality and past experience is that there's no true _private_ for data which is meant to be serializable to JSON. We've seen people use it for use-caes which should really use `fluant-syntax` ([example](https://searchfox.org/mozilla-central/rev/11712bd3ce7454923e5931fa92eaf9c01ef35a0a/browser/base/content/test/static/browser_misused_characters_in_strings.js#208)).

Would it be enough to slightly obfuscate the internals of patterns to hint to consumers that they're not supposed to access them? Something like `{_type: "var", _name: "user"}`, `{@@type: "var", @@name: "user"}`, `{_type: NodeKind.VariableReference, _name: "user"}` (if we migrate to TypeScript), or even a tuple: `["var", "user"]`?

stasm:
I think the question about what's public and what's private can be broken down into two questions:

1. How should changes to the `Pattern` type impact the version number of `@fluent/bundle`?

    Originally I proposed that any changes are signaled by a major version bump. This would effectively make the entire runtime AST public, which has consequences for Question 2.

    Would it be OK to allow for changes inside of the `Pattern` type in minor releases?

1. What level of compatibility is expected from drop-in replacements?

    My original thinking was that replacements must implement the same methods as `@fluent/bundle`'s `FluentBundle` (i.e. `addResource`, `hasMessage`, `getMessage`, `formatPattern`) and the object returned by `getMessage` must have exactly the same shape as returned by `FluentBundle.getMessage`.

    After @eemeli's comments, I'm now partial to requiring that `getMessage` return `{value: Pattern, attributes: {[name: string]: Pattern}}` shapes, where `Pattern` is an implementation-specific type.

zbraniecki:
I'm not sure how this will work. At some place we need to expose a stable public API, and while working on Rust and Gecko implementations I have two strong cases where I have to explicitly decide on what I expose.

In Rust, it's a bit more flexible, I can create a type called `Pattern` and expose it. But in WebIDL it's a bit more tricky. I can't imagine specifying IDL for a complex structure that I think you want to make `Pattern` to be.

What I have right now is:

`Message`:
 - Rust - https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/src/bundle.rs#L21
 - WebIDL - https://searchfox.org/mozilla-central/rev/040aa667f419932adf425d92c7438f03230ad96b/dom/chrome-webidl/Localization.webidl#32-40

`L10nKey`:
 - WebIDL - https://searchfox.org/mozilla-central/rev/040aa667f419932adf425d92c7438f03230ad96b/dom/chrome-webidl/Localization.webidl#16-21
 - (I don't have it yet in Rust but wanted to introduce)

Those two are relatively simple to encode and carry around, and even can be easily serialized if needed.
`L10nKey` serves as input, and `Message` serves as output.

I know `FluentBundle` is meant to be lower level and maybe it won't public IDL-like APIs, but I'm concerned about writing it in JS as "Object" and then trying to reverse-engineer into more specific structures later.

Additionally, if you make the `FluentBundle` output implementation-specific opaque types you basically make the FluentBundle and its higher-level wrapper bound to each other.
It may sound like a theoretical issue, but I'm right this moment experimenting with replacing JS `FluentBundle` and `FluentResource` with Rust `FluentBundle` and `FluentResource` in Gecko without touching L10nRegistry and/or Localization APIs. It seems very promising and is only possible because the API of `FluentBundle` takes "simple" structs and returns "simple" structs.
It seems to me like your approach will make us lose it.

p.s. The number of places where this conversation is being held is a bit overwhelming for me.

eemeli:
> (@stasm) I think the question about what's public and what's private can be broken down into two questions:
> 
> 1. How should changes to the `Pattern` type impact the version number of `@fluent/bundle`?
>    Originally I proposed that any changes are signaled by a major version bump. This would effectively make the entire runtime AST public, which has consequences for Question 2.
>    Would it be OK to allow for changes inside of the `Pattern` type in minor releases?

I would say that `Pattern` changes should be allowed in minor releases, as/if its internals are not a part of the public API. It's also good to note here that it's much easier to make private things public, rather than the reverse.

> 2. What level of compatibility is expected from drop-in replacements?
>    My original thinking was that replacements must implement the same methods as `@fluent/bundle`'s `FluentBundle` (i.e. `addResource`, `hasMessage`, `getMessage`, `formatPattern`) and the object returned by `getMessage` must have exactly the same shape as returned by `FluentBundle.getMessage`.
>
>    After @eemeli's comments, I'm now partial to requiring that `getMessage` return `{value: Pattern, attributes: {[name: string]: Pattern}}` shapes, where `Pattern` is an implementation-specific type.

This would be perfect for me. `fluent-compiler` for instance encodes a `Pattern` as a function of the message arguments, returning an array of parts that are then stringified and joined by `formatPattern`.

> (@zbraniecki) Additionally, if you make the `FluentBundle` output implementation-specific opaque types you basically make the FluentBundle and its higher-level wrapper bound to each other.
> It may sound like a theoretical issue, but I'm right this moment experimenting with replacing JS `FluentBundle` and `FluentResource` with Rust `FluentBundle` and `FluentResource` in Gecko without touching L10nRegistry and/or Localization APIs. It seems very promising and is only possible because the API of `FluentBundle` takes "simple" structs and returns "simple" structs.
> It seems to me like your approach will make us lose it.

As I see it, this binding only applies if the opaque output of the bundle is meant to be consumed by anything other than the bundle itself. I'm facing pretty much the same issue with `fluent-compiler`, and I don't see a problem here.

stasm:
> `Message`:
> 
>     * Rust - https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/src/bundle.rs#L21
>     * WebIDL - https://searchfox.org/mozilla-central/rev/040aa667f419932adf425d92c7438f03230ad96b/dom/chrome-webidl/Localization.webidl#32-40

Both of these describe a _formatted_ message. The `formatPattern` proposal also requires that there be a _raw_ message shape, returned by `FluentBundle.getMessage`. This raw message also has the `value` and `attributes` keys, whose values are a `Pattern` and a map of attribute names to `Patterns`, respectively. Those patterns can then be passed to `FluentBundle.formatPattern` to get a formatted string.

> Additionally, if you make the `FluentBundle` output implementation-specific opaque types you basically make the FluentBundle and its higher-level wrapper bound to each other.

I don't think this is true. Higher-level wrappers will continue to use `FluentBundle.getMessage` and `FluentBundle.formatPattern` in unison. As long as both of these methods come from the same implementation, things should work fine.

> It seems very promising and is only possible because the API of FluentBundle takes "simple" structs and returns "simple" structs. It seems to me like your approach will make us lose it.

The `formatPattern` proposal specifically doesn't change the fact that `FluentBundle` takes "simple" structs and returns "simple" structs_, unless I misunderstand what you mean by that. In your experiments, how do you describe the struct passed to the current `FluentBundle.format` method? The `Localization.webidl` you pointed to describes a _formatted_ message, but `FluentBundle.format` takes a runtime AST structure.

stasm:
Ah, I see it now. Your experiment uses the `compound` API (which hasn't landed), not the current `format` API ([1](https://phabricator.services.mozilla.com/D36286#change-EOGjyxVpa3BG), [2](https://phabricator.services.mozilla.com/D36286#change-aWcZhKJBoBYl)). So you're only working with _formatted_ messages there.

@zbraniecki Do you have thoughts on how this experiment would need to be adjusted to allow any of the API redesign proposal which keep the `FluentBundle.getMessage` method? Would it make sense to expose the `Pattern` AST node via the WebIDL?

stasm:
I talked to @zbraniecki last night and we clarified the questions we both had. For the Gecko scenario, it will be possible for `getMessage` to return a `{value: Pattern, attributes: Record<string, Pattern>}` shape, where `Pattern` is an opaque type.

The `{value: Pattern, attributes: Record<string, Pattern>}` shape will be part of the public API of `@fluent/bundle`. The `Pattern` type will remain "private", and other implementation will be free to use any structure they see fit for the purpose. This should enable much more experimentation and creativity wrt. the resolver's implementation.

Thank you, everyone, for your comments and patience. I'll rebase this PR and update it slightly, and I'll have it ready for review by the end of this week.

fluent-dom/src/localization.js
```diff
@@ -164,26 +164,26 @@ export default class Localization {
...
+ * Format the value of a message into a string or `null`.
  *
  * This function is passed as a method to `keysFromBundle` and resolve
  * a value of a single L10n Entity using provided `FluentBundle`.
  *
- * If the function fails to retrieve the entity, it will return an ID of it.
- * If formatting fails, it will return a partially resolved entity.
- *
- * In both cases, an error is being added to the errors array.
+ * If the message doesn't have a value, return `null`.
```
stasm: This is how it worked before, too. The comment was wrong.

fluent/test/context_test.js
```diff
@@ -180,13 +180,17 @@ suite('Bundle', function() {
...
     });
 
-    test('returns null for terms and missing messages', function() {
-      assert.equal(bundle.getMessage('-bar'), null);
-      assert.equal(bundle.getMessage('baz'), null);
-      assert.equal(bundle.getMessage('-baz'), null);
+    test('returns undefined for terms and missing messages', function() {
+      assert.strictEqual(bundle.getMessage('-bar'), undefined);
+      assert.strictEqual(bundle.getMessage('baz'), undefined);
+      assert.strictEqual(bundle.getMessage('-baz'), undefined);
```
stasm: I filed #389 to switch to using `strictEqual` in all tests.

fluent/src/bundle.js
```diff
@@ -170,74 +161,80 @@ export default class FluentBundle {
...
+      let scope = this._createScope(args, errors);
+      let value = resolveComplexPattern(scope, pattern);
+      return value.toString(scope);
     }
 
-    // optimize simple-string entities with attributes
-    if (typeof message.value === "string") {
-      return this._transform(message.value);
-    }
+    throw new TypeError("Invalid Pattern type");
```
stasm: I'm not really sure what's the right thing to do here. Should we guard against unexpected types of `patterns`? This could be helpful when someone forgets to check if `message.value` is not `null` before calling `formatPattern`. If I don't throw here, `resolveComplexPattern` likely will, anyways, but with a more cryptic error.

stasm:
@Pike This is now ready for your review.

The PR changes the `FluentResource` to be a subclass of `Array`; all runtime parser fixtures had to be regenerated. I also had to update almost all tests, because of the `format` → `formatPattern` change.

I left a few comments above. I have a few ideas for direct follow-ups, but I didn't want to make this PR even bigger. I filed new issues for each of the planned follow-ups.

My biggest question is: should `formatPattern` throw on invalid `pattern` types? I'd like to avoid forcing consumers of the API to wrap each call to `formatPattern` in a `try...catch` block, mostly for performance reasons. If you only pass `value` and `attributes` that you got from `getMessage`, the `pattern` type is guaranteed to be correct. So as long as you use the API right, no throwing will happen. However, I also see value in throwing early on invalid types. Otherwise, the error is way more cryptic since it happens deep inside of the resolver.

stasm:
@zbraniecki Would you mind taking a look at the changes to `fluent-dom`, please?

fluent/src/bundle.js
```diff
@@ -170,74 +161,80 @@ export default class FluentBundle {
...
+   *         // Returns 'Hello, Jane!' and `errors` is empty.
+   *
+   *         bundle.formatPattern(hello.value, undefined, errors);
+   *         // Returns 'Hello, {$name}!' and `errors` is now:
+   *         // [<ReferenceError: Unknown variable: name>]
+   *     }
+   *
+   * @param   {string|Array} pattern
+   * @param   {?Object} args
+   * @param   {?Array} errors
```
Pike: I'm a bit concerned about `errors` being optional. I understand that it's nice for tests, but it's really not how the API should be used.

Interestingly, `args` is actually optional in the code.

I'd feel better about this if we created an empty array in `scope` if no errors are given.

Maybe this is good for a follow-up? I.e., declaring them optional or not, and how to deal with errors when they're not passed in? (Sneak peek, I wonder if we should throw with the actual error instead of with `undefined has no method push`.

fluent-react/src/localized.js
```diff
@@ -101,27 +101,36 @@ export default class Localized extends Component {
...
+    if (msg.value === null) {
+      return cloneElement(elem, localizedProps);
+    }
+
+    const messageValue = bundle.formatPattern(msg.value, args);
+
+    // Check if the fallback is a valid element -- if not then it's not
+    // markup (e.g. nothing or a fallback string) so just use the
+    // formatted message value
+    if (!isValidElement(elem)) {
```
Pike: The check to `isValidElement` moved to a weird spot, aka, you call `cloneElement` before this now?

fluent/src/bundle.js
eemeli: Could leaving `errors` undefined mean that the error gets thrown instead?

stasm: > I'd feel better about this if we created an empty array in `scope` if no errors are given.

That's how it works right now. It happens in `_createScope(args, errors = []) {..}`.

fluent/src/bundle.js
stasm: My rationale for throwing here is that this isn't a formatting errors. It's not caused by a bug in the translation which resulted in an incompletely formatted output. As such, I don't think it fits into `errors` either. This is an error in how the API has been used by the consumer. If the consumer takes care to always only pass `Patterns` to `formatPattern`, this error will never happen.

fluent-react/src/localized.js
stasm: Thanks, I'll investigate.

fluent/src/bundle.js
Pike: oops. Filed #390 on making errors actually fail verbosely if they're hidden.

fluent/src/bundle.js
Pike: I guess your last comment also addresses your initial concern if developers should guard their calls with `try catch` ?

Also, we're not really protecting this API by checking if the input is an `Array` against being fed garbage, `[{foopy:true}]` is still bad ;-)

If `Pattern` was a subclass of `Array`, we might have a slightly stronger assertion?

fluent/src/bundle.js
stasm: Ah, I see what you mean. Thanks for filing #390.

fluent/src/bundle.js
stasm: > I guess your last comment also addresses your initial concern if developers should guard their calls with try catch ?

Oh! Yes, I think so! :)

Let me try to rephrase this to make sure I understand your point: If I'm fine expecting developers to use a `try catch`, I should also be fine expecting them to check `if (message.value)` before calling `formatPattern`. Is this what you meant?

Pike: I'm not sure what you're trying to achieve. If this is just about `message.value` being null unexpectedly, I'd consider `null` to be actually an OK return value.

I read this to be about creating a defined error scenario when people pass in garbage like `Math.random` or so. I wouldn't expect good users of the API to `try-catch` to proof them for API abuse.

stasm: OK, thanks for clarifying. I'd like to consider adding `null` as an OK return value in a follow-up. I'd prefer if this method only returned strings, so that it's not necessary to check for non-null output after every `formatPattern`.

For now, I'll keep the throwing on non-string, non-array pattern in order to help developers find buggy callsites faster.

fluent-react/src/localized.js
stasm: Nice catch, @Pike. I'm adding two unit tests to cover scenarios where this would have bitten us. Thanks again.

stasm:
@Pike I've updated the PR to address your feedback. I also added two tests to `fluent-react`, as mentioned earlier. Thanks again for catching the `isValidElement` problem.

I kept `throw` in `formatPattern` when the argument is not a string nor an array, because I'd like to avoid adding `null` as a valid return value. At least for now. My rationale is that doing so would effectively force every call to `formatPattern` to be accompanied by a `!== null` check. Perhaps we can find a better way in #390.

zbraniecki:
> Thanks, lgtm. Still need @zbraniecki to look at the changes in fluent-dom, aka Fluent in gecko in spe.

I think that `Fluent in Gecko` currently becomes more of a `does it work in fluent-rs` than `fluent-dom`.

I'm tackling that in https://github.com/projectfluent/fluent-rs/issues/119 and trying to find a way to massage this API into Rust and Gecko.
I'm not sure if this is the best approach - I can also just review this patch as-is, but then I don't think there's another place to say "From Gecko perspective, it's not enough that the API proposal works in JS, and we need to make it work in Gecko" and I believe it should block this.

stasm:
I created a release branch called https://github.com/projectfluent/fluent.js/tree/release-fluent-zero-thirteen and re-targeted this PR to it. I hope to be able to land this PR there soon and move on to working on the follow-up for the rest of the week.

fluent-dom/src/localization.js
```diff
@@ -164,26 +164,26 @@ export default class Localization {
...
  *
  * @param   {FluentBundle} bundle
- * @param   {Array<Error>}   errors
- * @param   {string}         id
- * @param   {Object}         args
+ * @param   {Array<Error>} errors
+ * @param   {Object} message
+ * @param   {Object} args
  * @returns {string}
```
zbraniecki: Can you update this comment to indicate that it can be null?

fluent-dom/src/localization.js
```diff
@@ -195,34 +195,28 @@ function valueFromBundle(bundle, errors, id, args) {
...
     attributes: null,
   };
 
-  if (msg.attrs) {
+  if (message.value) {
+    formatted.value = bundle.formatPattern(message.value, args, errors);
+  }
+
+  if (message.attributes) {
     formatted.attributes = [];
```
zbraniecki: use `formatted.attributes = new Array(message.attributes.length());` to avoid reallocation.

stasm: Oh, this actually made me realize that there's a bug here. `message.attributes` is a non-nullable object now, so the `if (message.attributes)` check always passes.

stasm: Aaand fixed. Thanks!

Issue #379, opened 2019-06-20
Unit test for fluent-react withLocalization fails inconsistently


flut1:
I noticed that my pull request #378 was failing in Travis on the following test:

> getString with access to the l10n context, with message changes

Upon further investigation, it seems like this also fails on the master branch, but not consistently. If I run all tests for fluent-react, they seem to _sometimes_ succeed, and if I run this test in isolation it seems to _always_ succeed. 

I can think of 2 possible reasons:
 - The tests are somehow affecting each other. The test is using the [full rendering API](https://airbnb.io/enzyme/docs/api/mount.html) from Enzyme. Their docs say that tests _may_ affect each other. 
 - The test contains a race condition. The line `l10n.setBundles([newBundle]);` causes a force-update in `WithLocalization`, but maybe React doesn't flush these changes synchronously.

flut1:
My suggestion would be to add `wrapper.update()` before the assertion here:

https://github.com/projectfluent/fluent.js/blob/ab37e1807116f777eb485aa2798ceae4cc17e9a0/fluent-react/test/with_localization_test.js#L112

This seems to fix it. Is it ok to force-update the component or should the test also assert that `WithLocalization` updates by itself? 

A similar test in `localized_change_test.js` seems to already have this `wrapper.update()` statement before the last assertion.

PR #382, opened 2019-06-29
Fix #381: change fluent-langnegs variant detection pattern to detect valid variants according to UTS35

### LINKED ISSUE #381: fluent-langneg incorrectly only allows 3-letter-long variants
This surfaced as https://github.com/mozilla/voice-web/issues/2076.

The code which builds the `localeRe` only allows 3-letter-long variants in locale tags:
https://github.com/projectfluent/fluent.js/blob/ab37e1807116f777eb485aa2798ceae4cc17e9a0/fluent-langneg/src/locale.js#L8

As per [BCP47](https://tools.ietf.org/html/bcp47#section-2.2.5), alphanumeric variant subtags should be 5 to 8 characters long, instead.

```
variant = 5*8alphanum         ; registered variants
        / (DIGIT 3alphanum)
```
```
In order to distinguish variants from other types of subtags,
registrations MUST meet the following length and content
restrictions:

1.  Variant subtags that begin with a letter (a-z, A-Z) MUST be
    at least five characters long.

2.  Variant subtags that begin with a digit (0-9) MUST be at
    least four characters long.
```

gion-andri:
According to UTS35 (Unicode Language Identifier standard, https://unicode.org/reports/tr35/), variants of locale tags can have the following form: 
```
  | = (alphanum{5,8} | digit alphanum{3}) ;
```

This change updates the implementation of fluent-langneg accordingly.

### DIFF
--- a/fluent-langneg/src/locale.js
+++ b/fluent-langneg/src/locale.js
@@ -5,17 +5,17 @@ import { getLikelySubtagsMin } from "./subtags";
 const languageCodeRe = "([a-z]{2,3}|\\*)";
 const scriptCodeRe = "(?:-([a-z]{4}|\\*))";
 const regionCodeRe = "(?:-([a-z]{2}|\\*))";
-const variantCodeRe = "(?:-([a-z]{3}|\\*))";
+const variantCodeRe = "(?:-(([0-9][a-z0-9]{3}|[a-z0-9]{5,8})|\\*))";
 
 /**
  * Regular expression splitting locale id into four pieces:
  *
- * Example: `en-Latn-US-mac`
+ * Example: `en-Latn-US-macos`
  *
  * language: en
  * script:   Latn
  * region:   US
- * variant:  mac
+ * variant:  macos
  *
  * It can also accept a range `*` character on any position.
  */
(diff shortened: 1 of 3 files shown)


gion-andri:
I don't know why this react-test failed. Is this related to #379?

zbraniecki:
Looks like it! @stasm - can we land it without the fix for #379?

stasm:
Yes, sounds good to me. I'll debug this rogue test this week. No need to block this PR on it.

stasm:
@zbraniecki I flagged you for review since you are the author and the maintainer of this code. Let me know if you'd like me to continue doing so, and whether you'd like to review this particular PR. It looks good to me.

zbraniecki:
I'm happy to defer reviews to you in whatever capacity you feel comfortable with. Now that we have Unicode Language Identifiers (finally!) we don't have to carve the subset of BCP47 manually.

I'll maintain MozLocale and unic-locale which both will follow the same ULI pattern and hopefully unic-locale will soon replace MozLocale in Gecko and maybe will become an option for the replacement of this package as well :)

stasm:
Thank for the PR, @gion-andri! I published `fluent-langneg` 0.2.0 with your fix.

Issue #390, opened 2019-07-12
Throw on bundle.formatPattern if errors would be swallowed


Pike:
`bundle.format` and soon, `bundle.formatPattern` swallow errors if you don't pass in an errors array.

We should throw on errors instead, I think.

stasm:
Do you think we should:

1. throw to complain that `errors` were not passed as an argument, or…
2. change the error reporting mode and throw a special kind of error containing the array of collected errors?

Pike:
... or ...

3. throw the first error
4. throw the last error

I'd not do 1, I didn't think about 2.

eemeli:
I'd vote for option 3. At that point, no matter what the specific error, showing some fallback string like the message key in the UI is probably at least pretty close to the least wrong thing to do, while printing the error in the console. If the library user wants to do something more complicated than that, that's exactly what the `errors` argument is for.

Issue #364, opened 2019-04-29
Emit errors on FluentBundle instances


stasm:
Rather than passing an `errors` array to `format`, we could emit errors on the `FluentBundle` instance.

Pike:
That makes FluentBundle not reentrant, though? Probably not much of a problem in JavaScript, but as a general runtime API design, I don't think that's a good idea.

stasm:
This is only intended for the JS implementation, yes :)

Pike:
TBH, as long as we only have "one example implementation" to show how to do runtime, I'd recommend to not create APIs that are bad ideas in other languages.

stasm:
We have three, and we'll have a reference one soon, too. I strongly feel that we should design the JS API to be the best in class for JS apps.

stasm:
Can you expand a bit on what you mean by reentrant here? Do you use it as a proxy for _thread-safe_?

Pike:
Not a proxy for, but a condition for thread-safety. https://en.wikipedia.org/wiki/Reentrancy_(computing)

eemeli:
@stasm, could you clarify what you mean by "emit errors on the `FluentBundle` instance"? Would that mean throwing errors from e.g. `bundle.format()`, or having something like a `bundle.onError()` method to register a callback for error reporting?

stasm:
Sorry, I should have been more specific. I meant the latter: `bundle.onerror` or `bundle.addEventListener("error")`.

Pike:
I just realized that this ties in to a pet peeve of mine, and that's runtime error fallback.

I think that not doing fallback on runtime errors is a bug. It's an OK shortcut to get gecko off the ground, but I don't think it's the right design choice.

Moving errors to an API "someone can know" makes it much more cumbersome to implement runtime-error-fallback based on `FluentBundle`.

stasm:
While I don't agree with the premise that we should fall back on runtime errors, I see how switching to the emitter pattern would make that harder. I see `FluentBundle` as a low-level API which should allow many different use-cases and fallback strategies. Let's keep a way to easily query the errors on a per-`format()` basis. Currently this is done by passing the `errors` array as the last argument to `format`. I wonder what the performance impact of returning an array of errors in a `{value, errors}` tuple would be. Do you think that's something worth testing?

[...]

PR #395, opened 2019-07-17
Bail out of Patterns with too long placeables

### LINKED ISSUE #388: Unskip the Billion Laughs test
Let's fix it.

stasm:
Fix #388.

### DIFF
--- a/fluent/src/resolver.js
+++ b/fluent/src/resolver.js
@@ -289,10 +289,11 @@ export function resolveComplexPattern(scope, ptn) {
           `(${part.length}, max allowed is ${MAX_PLACEABLE_LENGTH})`
         )
       );
-      result.push(part.slice(MAX_PLACEABLE_LENGTH));
-    } else {
-      result.push(part);
+      scope.dirty.delete(ptn);
+      return new FluentNone();
     }
+
+    result.push(part);
 
     if (useIsolating) {
       result.push(PDI);
--- a/fluent/src/resolver.js
+++ b/fluent/src/resolver.js
@@ -283,14 +283,15 @@ export function resolveComplexPattern(scope, ptn) {
     }
 
     if (part.length > MAX_PLACEABLE_LENGTH) {
-      scope.errors.push(
-        new RangeError(
-          "Too many characters in placeable " +
-          `(${part.length}, max allowed is ${MAX_PLACEABLE_LENGTH})`
-        )
-      );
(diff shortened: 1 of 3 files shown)


fluent/src/resolver.js
```diff
@@ -289,11 +289,12 @@ export function resolveComplexPattern(scope, ptn) {
           `(${part.length}, max allowed is ${MAX_PLACEABLE_LENGTH})`
         )
       );
-      result.push(part.slice(MAX_PLACEABLE_LENGTH));
-    } else {
-      result.push(part);
+      scope.dirty.delete(ptn);
+      return new FluentNone();
```
stasm: This isn't perfect, but does provide some level of protection. The problem this approach still suffers from is that `FluentNone` will then `toString`-ed to `{???}` many many times over, depending on how many patterns fail in references up the nesting tree. This also means that the number of accumulated errors may be huge.

I filed https://github.com/projectfluent/fluent/issues/277 to continue the discussion in the spec.

Pike:
I wonder if we should `throw` to bail out of the attack. But that puts us back into the square of having to `try-catch` every call? Well, not every, we could do that in `formatPattern`, and that would only do it for complex patterns.

You had concerns around `try-catch` and perf?

I'm having second thoughts on this, in particular around question like how this would be implemented with `formatToParts` where we don't know the length. Or in Rust, where we can't just throw an exception and clean things up at the top-level.

Is this specific implementation practically better than what we have? Maybe it's not worth to tweak the status quo right now, and modify the behavior once we have something spec'ed?

stasm:
> You had concerns around try-catch and perf?

Yes, but I'll need to create a benchmark to verify this. I'll report in #364 or #390, or both.

> Is this specific implementation practically better than what we have? Maybe it's not worth to tweak the status quo right now, and modify the behavior once we have something spec'ed?

Yes, my rationale for opening this PR was along the same lines. This isn't perfect, but it is better that what we currently do on master.

stasm:
I changed the approach to treating the error as a _fatal_ one, which causes the resolver to instantly bail out of the affected pattern. This provides a much better CPU protection in deeply nested scenarios. I think we could even consider increasing `MAX_PLACEABLE_LENGTH`, but I left it for another time.

stasm:
Using `try...catch` has no impact on our benchmarks.

stasm:
Thanks for the review and the suggestion to bail out in this particular case!

PR #396, opened 2019-07-18
Omitting errors in formatPattern() makes it throw

### LINKED ISSUE #390: Throw on bundle.formatPattern if errors would be swallowed

stasm:
This PR introduces a new opt-in error reporting strategy in `formatPattern`.

- If the `errors` argument is specified, the behavior is unchanged: `formatPattern` returns a string which is a best-effort attempt to resolve the given pattern. Any errors produced during the resolution are collected in the `errors` array.

- If the `errors` argument is omitted, the behavior changes to throwing the first encountered error, which also stops the resolving immediately. No fallback value is produced in this case. It's entirely up the user of the API to decide what the fallback should be, if any.

This fixes #390. Also see #364.

### DIFF
--- a/fluent/src/bundle.js
+++ b/fluent/src/bundle.js
@@ -207,30 +207,35 @@ export default class FluentBundle {
    *         // [<ReferenceError: Unknown variable: name>]
    *     }
    *
+   * If `errors` is omitted, the first encountered error will be thrown.
+   *
    * @param   {string|Array} pattern
    * @param   {?Object} args
    * @param   {?Array} errors
    * @returns {string}
    */
   formatPattern(pattern, args, errors) {
-    // Resolve a simple pattern without creating a scope.
+    // Resolve a simple pattern without creating a scope. No error handling is
+    // required; by definition simple patterns don't have placeables.
     if (typeof pattern === "string") {
       return this._transform(pattern);
     }
 
     // Resolve a complex pattern.
-    if (Array.isArray(pattern)) {
-      let scope = this._createScope(args, errors);
-      try {
-        let value = resolveComplexPattern(scope, pattern);
-        return value.toString(scope);
-      } catch (err) {
-        scope.errors.push(err);
-        return new FluentNone().toString(scope);
-      }
+    let scope = this._createScope(args, errors);
+    let value;
+    try {
+      value = resolveComplexPattern(scope, pattern).toString(scope);
+    } catch (err) {
+      scope.errors.push(err);
+      value = new FluentNone().toString(scope);
     }
 
-    throw new TypeError("Invalid Pattern type");
+    if (!errors && scope.errors.length > 0) {
+      throw scope.errors[0];
+    }
+
+    return value;
   }
 
   _createScope(args, errors = []) {
--- a/fluent/src/bundle.js
+++ b/fluent/src/bundle.js
@@ -1,6 +1,7 @@
 import {resolveComplexPattern} from "./resolver.js";
 import FluentResource from "./resource.js";
 import { FluentNone } from "./types.js";
+import Scope from "./scope.js";
 
 /**
  * Message bundles are single-language stores of translations.  They are
@@ -222,7 +223,7 @@ export default class FluentBundle {
     }
 
     // Resolve a complex pattern.
-    let scope = this._createScope(args, errors);
+    let scope = new Scope(this, errors, args);
     try {
       let value = resolveComplexPattern(scope, pattern);
       return value.toString(scope);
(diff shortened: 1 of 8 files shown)


fluent/src/bundle.js
```diff
@@ -207,35 +207,38 @@ export default class FluentBundle {
...
-      try {
-        let value = resolveComplexPattern(scope, pattern);
-        return value.toString(scope);
-      } catch (err) {
+    let scope = this._createScope(args, errors);
+    try {
+      let value = resolveComplexPattern(scope, pattern);
+      return value.toString(scope);
+    } catch (err) {
+      if (scope.errors) {
```
stasm: Here, `scope.error` controls whether _fatal_ errors should be thrown or should produce `{???}`.

fluent/src/resolver.js
```diff
@@ -47,6 +47,16 @@ const FSI = "\u2068";
 const PDI = "\u2069";
 
 
+// Helper: Report an error.
+function report(scope, error) {
+  if (scope.errors) {
```
stasm: Here, `scope.error` controls whether all error are treated as _fatal_.

stasm:
@eemeli I think this PR is very much in line with your suggestion from https://github.com/projectfluent/fluent.js/pull/380#discussion_r302963170. Would you like to take a look? Thanks!

fluent/test/functions_builtin_test.js
```diff
@@ -22,70 +26,93 @@ suite('Built-in functions', function() {
...
+      assert.strictEqual(bundle.formatPattern(msg.value, args, errors), '100%');
+      assert.strictEqual(errors.length, 0);
 
       msg = bundle.getMessage('num-bad-opt');
-      assert.strictEqual(bundle.formatPattern(msg.value, args), '1');
+      assert.strictEqual(bundle.formatPattern(msg.value, args, errors), '1');
+      assert.strictEqual(errors.length, 0);
     });
 
+    // XXX Functions should report errors.
```
Pike: Can I haz issue for `XXX` ?

fluent/test/errors_test.js
```diff
@@ -0,0 +1,96 @@
+"use strict";
+
+import assert from "assert";
+import ftl from "@fluent/dedent";
+
+import FluentBundle from "../src/bundle";
+
+suite("Errors", function() {
```
Pike: I know that you like your tests verbose, but this test file scared me.

I think one suite with two tests would provide more value to readers, and give the same test coverage. "Two calls" w/ `errs`, and "First error is thrown" w/out.

stasm: You're right, I'll make this shorter.

stasm:
> I wonder, should we put report on scope itself? report(scope,..) looks like it should.

I don't mind `report` being just a function, but I like the idea of defining scope more formally as a class. Thanks for the suggestion, I'll do it.

fluent/src/scope.js
```diff
@@ -0,0 +1,29 @@
+export default class Scope {
+  constructor(
+    bundle,
+    errors,
+    args,
+    parameterized = false,
+    dirty = new WeakSet()
+  ) {
+    this.bundle = bundle;
```
stasm: It would be nice in the future to replace `Scope.bundle` with all the private members of `FluentBundle` which are needed by the resolver. I.e. add `Scope.messages`, `Scope.terms` etc. This would reduce the API surface of the bundle. I'm leaving this as a follow-up, though, to keep this PR on topic.

fluent/src/resolver.js
```diff
@@ -191,9 +191,9 @@ function TermReference(scope, {name, attr, args}) {
     return new FluentNone(id);
   }
 
-  // Every TermReference has its own args.
-  const [, keyargs] = getArguments(scope, args);
-  const local = scope.clone(keyargs);
+  // Every TermReference has its own variables.
+  const [, params] = getArguments(scope, args);
```
Pike: FWIW, I think using `params` as variable name here is good.

fluent/src/scope.js
```diff
@@ -1,16 +1,22 @@
 export default class Scope {
-  constructor(bundle, errors, args,
-    insideTermReference = false, dirty = new WeakSet()
+  constructor(
+    bundle,
+    errors,
+    args,
+    parameterized = false,
```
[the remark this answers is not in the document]
stasm: I think I'd prefer to go back to `insideTermReference`.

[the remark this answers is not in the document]
Pike: WFM.

stasm:
Thank you for the reviews, @eemeli and @Pike!

PR #398, opened 2019-07-19
Report errors from Functions and Intl objects

### LINKED ISSUE #106: Handle errors thrown by Intl formatters
In https://github.com/projectfluent/fluent.js/commit/cc32a1ec5608b2409132f2f0c0551106100d8cb5 we started catching errors thrown by Intl formatters inside of `FluentType`. Should we report them somehow?

### LINKED ISSUE #107: Handle errors thrown by Call Expressions
In https://github.com/projectfluent/fluent.js/commit/cc32a1ec5608b2409132f2f0c0551106100d8cb5 we started catching errors thrown by Fluent functions (like built-ins) in Call Expressions. Should we report them somehow?

stasm:
I had a look at our error reporting and I realized that there really isn't anything that would make it hard to report errors thrown by Functions and from `memoizeIntlObject`. Yay.

Fix #106. Fix #107.

### DIFF
--- a/fluent/src/builtins.js
+++ b/fluent/src/builtins.js
@@ -28,7 +28,7 @@ function values(opts) {
 export
 function NUMBER([arg], opts) {
   if (arg instanceof FluentNone) {
-    return new FluentNone("NUMBER()");
+    return new FluentNone(`NUMBER(${arg.valueOf()})`);
   }
 
   if (arg instanceof FluentNumber) {
@@ -41,7 +41,7 @@ function NUMBER([arg], opts) {
 export
 function DATETIME([arg], opts) {
   if (arg instanceof FluentNone) {
-    return new FluentNone("DATETIME()");
+    return new FluentNone(`DATETIME(${arg.valueOf()})`);
   }
 
   if (arg instanceof FluentDateTime) {
--- a/fluent/src/builtins.js
+++ b/fluent/src/builtins.js
@@ -11,7 +11,7 @@
  * `FluentType`.  Functions must return `FluentType` objects as well.
  */
 
-import { FluentNumber, FluentDateTime } from "./types.js";
+import { FluentNone, FluentNumber, FluentDateTime } from "./types.js";
 
 function merge(argopts, opts) {
   return Object.assign({}, argopts, values(opts));
@@ -27,6 +27,10 @@ function values(opts) {
 
 export
 function NUMBER([arg], opts) {
+  if (arg instanceof FluentNone) {
+    return new FluentNone("NUMBER()");
+  }
+
   if (arg instanceof FluentNumber) {
     return new FluentNumber(arg.valueOf(), merge(arg.opts, opts));
   }
@@ -36,6 +40,10 @@ function NUMBER([arg], opts) {
 
 export
 function DATETIME([arg], opts) {
+  if (arg instanceof FluentNone) {
+    return new FluentNone("DATETIME()");
+  }
+
   if (arg instanceof FluentDateTime) {
     return new FluentDateTime(arg.valueOf(), merge(arg.opts, opts));
   }
--- a/fluent/src/builtins.js
+++ b/fluent/src/builtins.js
@@ -11,7 +11,7 @@
  * `FluentType`.  Functions must return `FluentType` objects as well.
  */
 
-import { FluentNone, FluentNumber, FluentDateTime } from "./types.js";
+import { FluentNumber, FluentDateTime } from "./types.js";
 
 function merge(argopts, opts) {
   return Object.assign({}, argopts, values(opts));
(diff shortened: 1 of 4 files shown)


fluent/src/builtins.js
```diff
@@ -27,22 +27,18 @@ function values(opts) {
 
 export
 function NUMBER([arg], opts) {
-  if (arg instanceof FluentNone) {
-    return arg;
-  }
   if (arg instanceof FluentNumber) {
     return new FluentNumber(arg.valueOf(), merge(arg.opts, opts));
   }
-  return new FluentNone("NUMBER()");
```
stasm: This was introduced in April in https://github.com/projectfluent/fluent.js/commit/24355cf9aa01ef19b72613c0d7423c978bf51490 to provide better fallback values. Today we can simply throw :)

fluent/src/builtins.js
```diff
@@ -27,22 +27,18 @@ function values(opts) {
 
 export
 function NUMBER([arg], opts) {
-  if (arg instanceof FluentNone) {
-    return arg;
-  }
```
Pike: I think we should keep this ...

fluent/src/builtins.js
```diff
@@ -27,22 +27,18 @@ function values(opts) {
...
-  return new FluentNone("NUMBER()");
+
+  throw new TypeError("Invalid argument type to NUMBER");
 }
 
 export
 function DATETIME([arg], opts) {
-  if (arg instanceof FluentNone) {
-    return arg;
-  }
```
Pike: ... and this ...

fluent/test/functions_builtin_test.js
```diff
@@ -25,20 +25,32 @@ suite('Built-in functions', function() {
...
-      assert.strictEqual(bundle.formatPattern(msg.value, {}, errors), '{$arg}');
-      assert.strictEqual(errors.length, 1);
-      assert.ok(errors.pop() instanceof ReferenceError);
-
+      assert.strictEqual(bundle.formatPattern(msg.value, {}, errors), '{NUMBER()}');
+      assert.strictEqual(errors.length, 2);
+      assert.ok(errors[0] instanceof ReferenceError);
+      assert.strictEqual(errors[0].message, "Unknown variable: $arg");
+      assert.ok(errors[1] instanceof TypeError);
+      assert.strictEqual(errors[1].message, "Invalid argument type to NUMBER");
```
Pike: ... to avoid double reporting a single error.

stasm:
Thanks, @Pike.

> I think we should ensure that a single error has a single message.

I think I'd like to disagree. It's a slippery slope. There might be scenarios where it does make sense to have more than one message per error. We don't know yet, and I wouldn't make exclude it from the design. I wouldn't make it an assumption of the general design.

In my mind, reporting these errors is a feature. They describe what happened. There was a missing variable reference, and then you tried to pass it to a function.

It's an interesting bit of design because this _technically_ isn't the resolver. It's how we implement the built-in functions, or any function for that matter. And I'm not sure if I like the idea of having to check every argument to a function in case it's `FluentNone`. Conceptually, throwing is a simpler, more atomic approach. At least in JS.

Would you be willing to revisit this when we discuss the spec? We might want to consider how this impacts other implementations, too. Also, let's keep in mind that we're talking about an edge-case of an edge-case :)

Pike:
In a way, you're making an argument for stack traces.

I think we created quite a problem if a stack trace is something a downstream consumer like a Firefox dev or a localizer need to figure out how to fix their Fluent string.

I also think that mixing errors and stack traces is making things harder to understand. Every human consumer of Fluent errors needs to understand the fluent.js code structure to be able to tell apart errors and their stacks.

stasm:
Yes, in a way they're stack traces, but flat. Neither solution is satisfying to me, really. One produces too many errors in some (very rare) cases; the other requires boilerplate in function implementations, and might not even hold for custom functions, depending on their implementations.

If I wanted to implement a custom function `SUM(one, two)`, how should the code look like? Would I need to check each arg for `FluentNone` and, if true, return one of them or a new `FluentNone`? Then check if both args are `FluentNumbers` and compute the sum. And then throw for all other cases?

In the future, we might add the possibility of using variables as keyword arguments, too (https://github.com/projectfluent/fluent/issues/230), which would mean that function need to check them against `FluentNone` as well.

What bothers me about this approach is that it's entirely up to the function implementation to adhere to it. Or not. I'd rather find a bullet-proof solution inside the resolver that guarantees the desired behavior. Building on top of the stack track analogy, perhaps we should in fact create a new trace in each `FunctionReference`? We could use a new `Scope` with an empty `errors` array for this purpose, and then figure out how to put those errors back in the original top-level `errors`. That's of course an idea for a follow-up.

For right now, my instinct is to go with the approach which employs less magic. By that, I mean fewer conditions and checks, and instead relying on the code that's already there. Yes, it means that in some (again, very rare) cases two errors will be produced instead of one, but crucially, the important one (about the missing variable) will still be there.

If you strongly feel that's not the right call, I'll add those `ifs` back.

Pike:
Luke and I had a lengthy conversation around functions and errors in https://github.com/projectfluent/python-fluent/pull/92.

I think we need to establish that Fluent functions are kernel code. They might be user land, but they're still kernel. We can't take away that Fluent functions need to validate their arguments. Number and date formatters are fuzzed in the implementations we refer to, so that's less of a risk. But that's an exception.

To take this out of abstract constructs, in this concrete case we're talking about:

```
Unknown variable: $arg
Invalid argument type to NUMBER
```

and I find that confusing. I can reverse engineer myself out of that, but it's still confusing to me.

I'd rather not have that problem right now.

PR #403, opened 2019-07-23
Update FluentBundle docstrings

stasm:
The docstrings needed a few updates to reflect the recent changes to the FluentBundle API.

### DIFF
--- a/fluent/src/bundle.js
+++ b/fluent/src/bundle.js
@@ -1,35 +1,26 @@
+import {FluentNone} from "./types.js";
 import {resolveComplexPattern} from "./resolver.js";
 import FluentResource from "./resource.js";
-import { FluentNone } from "./types.js";
 import Scope from "./scope.js";
 
 /**
- * Message bundles are single-language stores of translations.  They are
- * responsible for parsing translation resources in the Fluent syntax and can
- * format translation units (entities) to strings.
- *
- * Always use `FluentBundle.formatPattern` to retrieve translation units from a
- * bundle. Translations can contain references to other entities or variables,
- * conditional logic in form of select expressions, traits which describe their
- * grammatical features, and can use Fluent builtins which make use of the
- * `Intl` formatters to format numbers, dates, lists and more into the
- * bundle's language. See the documentation of the Fluent syntax for more
- * information.
+ * Message bundles are single-language stores of translation resources. They are
+ * responsible for formatting message values and attributes to strings.
  */
 export default class FluentBundle {
   /**
    * Create an instance of `FluentBundle`.
    *
    * The `locales` argument is used to instantiate `Intl` formatters used by
-   * translations.  The `options` object can be used to configure the bundle.
+   * translations. The `options` object can be used to configure the bundle.
    *
    * Examples:
    *
-   *     const bundle = new FluentBundle(locales);
+   *     let bundle = new FluentBundle(["en-US", "en"]);
    *
-   *     const bundle = new FluentBundle(locales, { useIsolating: false });
+   *     let bundle = new FluentBundle(locales, {useIsolating: false});
    *
-   *     const bundle = new FluentBundle(locales, {
+   *     let bundle = new FluentBundle(locales, {
    *       useIsolating: true,
    *       functions: {
    *         NODE_ENV: () => process.env.NODE_ENV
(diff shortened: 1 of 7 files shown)


fluent/src/resolver.js
```diff
@@ -14,24 +15,14 @@
...
- *      bundle for which the given resolution is happening
- *  * {Object} args
- *      list of developer provided arguments that can be used
- *  * {Array} errors
- *      list of errors collected while resolving
- *  * {WeakSet} dirty
- *      Set of patterns already encountered during this resolution.
- *      This is used to prevent cyclic resolutions.
+ * Functions in this file pass around an instance of the `Scope` class, which
+ * stores the data required for successful resolution and error recovery.
```
Pike: Side question, is this rendered in the esdocs somewhere? I couldn't find it, and I wonder if that's a bug or a feature.

Also, can't comment on that inline, we don't do list formats or more right now?

Should we also tweak `bundle's language` to `languages`, as it can have more than one?

fluent/src/scope.js
```diff
@@ -6,13 +6,17 @@ export default class Scope {
     insideTermReference = false,
     dirty = new WeakSet()
   ) {
+    // The bundle for which the given resolution is happening.
```
Pike: Should these be exposed through esdocs?

One thing about `Scope` on esdocs is that it's the only class there that's not `FluentFoo`. Is that good or bad?

fluent/src/types.js
```diff
@@ -36,29 +36,45 @@ export class FluentType {
...
+  /**
+   * @param   {string} value - The fallback value of this FluentNone
+   * @returns {FluentType}
+   */
   constructor(value = "???") {
     super(value);
   }
 
+  /**
+   * @returns {string}
```
Pike: esdoc generates a confusing doc string for this one, at least locally for me. This might need a full doc string and not just a return type?

fluent/src/types.js
```diff
@@ -36,29 +36,45 @@ export class FluentType {
...
+   */
   toString() {
     return `{${this.value}}`;
   }
 }
 
 export class FluentNumber extends FluentType {
+  /**
+   * @param   {(number|string)} value
+   * @param   {Object} opts
```
Pike: This would benefit from actually documenting the `opts`.

fluent/src/types.js
```diff
@@ -71,10 +87,19 @@ export class FluentNumber extends FluentType {
 }
 
 export class FluentDateTime extends FluentType {
+  /**
+   * @param   {(Date|number|string)} value
+   * @param   {Object} opts
```
Pike: Same here.

fluent/src/resolver.js
stasm: > Side question, is this rendered in the esdocs somewhere? I couldn't find it, and I wonder if that's a bug or a feature.

I think it's a missing featre. The `@overview` tag is [supported in JSDoc](https://jsdoc.app/tags-file.html). I don't see it in the list of [tags supported by ESDoc](https://esdoc.org/manual/tags.html), though.

> Also, can't comment on that inline, we don't do list formats or more right now?

Correct. Fixed.

> Should we also tweak bundle's language to languages, as it can have more than one?

Fixed.

fluent/src/scope.js
stasm: > Should these be exposed through esdocs?

Fixed.

> One thing about Scope on esdocs is that it's the only class there that's not FluentFoo. Is that good or bad?

It's also the only class that is not part of the public API and it's not exposed through `index.js`. I think it's OK to keep the current name.

fluent/src/types.js
stasm: I'd like to avoid listing all posibile options. I'll mention `Intl.NumberFormat` and `Intl.DateTimeFormat`.

fluent/src/resolver.js
Pike: Found https://github.com/esdoc/esdoc/issues/66, it's wontfix.

fluent/src/bundle.js
```diff
@@ -76,13 +66,19 @@ export default class FluentBundle {
   }
 
   /*
```
Pike: Make that `/**` so that esdoc picks this up? Same for `hasMessage`.

stasm: Good catch!

PR #404, opened 2019-07-23
Remove FluentBundle.addMessages

### LINKED ISSUE #328: FluentResource entries allocation in FluentBundle is inconsistent
While working on fluent-rs I realized that we currently have a rather undefined around memory allocation in FluentBundle.

`add_resource` method is intended to take a reference to a `FluentResource` stored in some `ResourceManager` (like L10nRegistry), and in result the signature of the FluentBundle might look like this:

```rust
struct FluentBundle<'bundle> {
  resources: Vec<&'bundle FluentResource<'bundle>>
}
```

Unfortunately, at the same time, we have `add_messages` which accepts a `&str` and creates a `FluentResource` out of it and then stores its reference.
The issue here is that we don't have a place where to allocate and retain the `FluentResource` itself in that scenario.

In JS/python this is messy, but hidden. Rust exposes that inconsistency.

In my current model I'm going for sth like that:

```rust
struct ResourceManager {
  strings: HashMap<String, String>,
  resources: HashMap<String, FluentResource>
}

struct FluentBundle<'bundle> {
  resources: Vec<&'bundle FluentResource<'bundle>>
}

```

so that the resources and their strings are stored on ResourceManager and if two bundles use the same `FluentResource` they both will get immutable reference to it.

What should we do about `add_messages`? Is that still useful? If so, would it make sense to require some manager to be available to the bundle such as:

```rust
let bundle = FluentBundle::new(resmgr);
bundle.add_messages("key = Value");
```

or maybe we could consider dropping that and requiring an external resource manager to be used and provide references to the resources?

@stasm - thoughts?

stasm:
Fix #328.

### DIFF
--- a/fluent/test/context_test.js
+++ b/fluent/test/context_test.js
@@ -7,19 +7,15 @@ import FluentBundle from '../src/bundle';
 import FluentResource from '../src/resource';
 
 suite('Bundle', function() {
-  let bundle, args, errs;
+  let bundle;
 
-  setup(function() {
-    errs = [];
-  });
-
-  suite('addMessages', function(){
+  suite('addResource', function(){
     suiteSetup(function() {
       bundle = new FluentBundle('en-US', { useIsolating: false });
-      bundle.addMessages(ftl`
+      bundle.addResource(new FluentResource(ftl`
         foo = Foo
-        -bar = Private Bar
-        `);
+        -bar = Bar
+        `));
     });
 
     test('adds messages', function() {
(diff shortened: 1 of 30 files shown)


fluent/test/context_test.js
```diff
@@ -7,19 +7,16 @@ import FluentBundle from '../src/bundle';
...
-  setup(function() {
-    errs = [];
-  });
-
-  suite('addMessages', function(){
+  suite('addResource', function(){
     suiteSetup(function() {
       bundle = new FluentBundle('en-US', { useIsolating: false });
-      bundle.addMessages(ftl`
+      let resource = new FluentResource(ftl`
```
Pike: Do we need this resource?

stasm: Are you asking whether this should be a separate variable?

Pike: Yeah.

stasm: I see. Yeah, I can inline this one as well. Thanks.

PR #405, opened 2019-07-24
fix SyntaxError on Edge

mathroc:
when using a MutationObserver with attributesFilter on edge `attributes: true` needs to be defined (see https://stackoverflow.com/questions/50593385/mutationobserver-syntax-error-on-ie-11)

I guess `attribute: true` (without the `s`) was a typo


zbraniecki:
Thank you!

PR #410, opened 2019-07-26
Allow NUMBER and DATETIME to take strings, numbers, dates

stasm:
This PR relaxes the type guards inthe `NUMBER` and `DATETIME` builtins. They now accept strings, `FluentNumbers`, and `FluentDateTimes`. However, if the native value of the argument results in a `NaN` during `parseFloat`, an error is thrown.

Furthermore, `FluentDateTime` now stores the date as a number, which works well with the `isNaN` check mentioned above. `DateTimeFormat.format()` converts its `Date`-typed argument to a number anyways.

### DIFF
--- a/fluent/src/builtins.js
+++ b/fluent/src/builtins.js
@@ -31,11 +31,12 @@ function NUMBER([arg], opts) {
     return new FluentNone(`NUMBER(${arg.valueOf()})`);
   }
 
-  if (arg instanceof FluentNumber) {
-    return new FluentNumber(arg.valueOf(), merge(arg.opts, opts));
+  let value = parseFloat(arg.valueOf());
+  if (Number.isNaN(value)) {
+    throw new TypeError("Invalid argument to NUMBER");
   }
 
-  throw new TypeError("Invalid argument type to NUMBER");
+  return new FluentNumber(value, merge(arg.opts, opts));
 }
 
 export
@@ -44,9 +45,10 @@ function DATETIME([arg], opts) {
     return new FluentNone(`DATETIME(${arg.valueOf()})`);
   }
 
-  if (arg instanceof FluentDateTime) {
-    return new FluentDateTime(arg.valueOf(), merge(arg.opts, opts));
+  let value = parseFloat(arg.valueOf());
+  if (Number.isNaN(value)) {
+    throw new TypeError("Invalid argument to DATETIME");
   }
 
-  throw new TypeError("Invalid argument type to DATETIME");
+  return new FluentDateTime(value, merge(arg.opts, opts));
 }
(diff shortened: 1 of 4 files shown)


fluent/test/functions_builtin_test.js
```diff
@@ -99,24 +99,19 @@ suite('Built-in functions', function() {
...
-      assert.strictEqual(errors.length, 1);
-      assert.ok(errors[0] instanceof TypeError);
-      assert.strictEqual(errors[0].message, "Invalid argument type to NUMBER");
+      assert.strictEqual(bundle.formatPattern(msg.value, args, errors), '147,510,720,000,000%');
+      assert.strictEqual(errors.length, 0);
 
       errors = [];
       msg = bundle.getMessage('num-bad-opt');
-      assert.strictEqual(bundle.formatPattern(msg.value, args, errors), '{NUMBER()}');
+      assert.strictEqual(bundle.formatPattern(msg.value, args, errors), '1475107200000');
```
stasm: This value comes from here: https://github.com/projectfluent/fluent.js/blob/d3022dec7a68020e0dc0a3d295c0261ec4b21e71/fluent/src/types.js#L97

The variable reference resolved fine, but the bad option caused `toString` to go into the error branch. The error is correctly reported, but I'm not sure if just `return this.value` is the right fallback. We need to return a string here, so `FluentNone` is not option.

stasm: Perhaps simulating the result of a `FluentNone` would be better? I.e. `return "{???}"`.

stasm: Or, we could format using some reasonable defaults.

stasm: In 81cbe84 I changed this to return safe defaults.

fluent/src/resolver.js
```diff
@@ -134,10 +134,10 @@ function VariableReference(scope, {name}) {
     case "string":
       return arg;
     case "number":
-      return new FluentNumber(arg);
+      return new FluentNumber(parseFloat(arg));
```
Pike: This one's not needed, we know we have a number.

fluent/src/types.js
```diff
@@ -107,12 +106,14 @@ export class FluentDateTime extends FluentType {
   /**
    * Create an instance of `FluentDateTime` with options to the
    * `Intl.DateTimeFormat` constructor.
-   * @param   {(Date|number|string)} value
+   * @param   {number} value
```
Pike: Make this a bit more precise? `timestamp in milliseconds` or so?

fluent/src/builtins.js
```diff
@@ -44,9 +45,10 @@ function DATETIME([arg], opts) {
     return new FluentNone(`DATETIME(${arg.valueOf()})`);
   }
 
-  if (arg instanceof FluentDateTime) {
-    return new FluentDateTime(arg.valueOf(), merge(arg.opts, opts));
+  let value = parseFloat(arg.valueOf());
```
Pike: Let's use `Number` here.

PR #422, opened 2019-08-08
Allow whitespace before CallArguments (fixes #421)

### LINKED ISSUE #421: Parser should allow white-space before call arguments
This is parser fixes to match the added tests for projectfluent/fluent#281.

I tested this locally, and it only affects fluent syntax, but not bundle.

Pike:
This is picking up the tests for https://github.com/projectfluent/fluent/issues/281,
and adjusts the parser to pass those tests.

I'm not happy with the peek, but this is what makes the tests pass.

### DIFF
--- a/fluent-syntax/src/parser.js
+++ b/fluent-syntax/src/parser.js
@@ -676,7 +676,9 @@ class FluentParser {
       }
 
       let args;
-      if (ps.currentChar === "(") {
+      ps.peekBlank();
+      if (ps.currentPeek === "(") {
+        ps.skipToPeek();
         args = this.getCallArguments(ps);
       }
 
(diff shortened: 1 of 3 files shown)


fluent-syntax/src/parser.js
```diff
@@ -685,13 +687,15 @@ class FluentParser {
 
     if (ps.isIdentifierStart()) {
       const id = this.getIdentifier(ps);
+      ps.peekBlank();
```
zbraniecki: alternatively, you could store the ptr, `skipBlank` and then only take `currentChar === "."` if `stored_ptr == ptr`.

Pike: That's like re-implementing peek, that doesn't smell right. If I'd want to change this, I'd look for ways to structure the code so that I could just eat the whitespace, because there's no grammar structure that doesn't allow it.

But then I didn't want to invest my brain in that, too.

PR #430, opened 2019-10-02
hoist static functions in withLocalization

blushingpenguin:
This hoists static functions in components decorated using withLocalization as per https://reactjs.org/docs/higher-order-components.html

### DIFF
--- a/fluent-react/package.json
+++ b/fluent-react/package.json
@@ -48,6 +48,7 @@
   "dependencies": {
     "cached-iterable": "^0.2.1",
     "@fluent/sequence": "0.4.0",
+    "hoist-non-react-statics": "^3.0.0",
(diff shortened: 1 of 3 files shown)


Pike:
How does this relate/interact to/with the work @Gregoor started in #406?

blushingpenguin:
> How does this relate/interact to/with the work @Gregoor started in #406?

They are tangential -- in that PR withLocalization has been reimplemented using react hooks instead of context -- the behaviour wrt to static functions is the same.  The two changes do conflict, but only in a minor way -- AFAICS the only difference needed would be:

```javascript
return hoistNonReactStatics(WithDisplay, Inner);
```
in fluent-react/src/with_localization.js

Issue #435, opened 2019-12-02
Revisit Compatibility


stasm:
(Previous discussions: #133, #269)

#### Summary

I'd like to start a long-term clean-up project which aims at the following:

- Simplify the build pipeline and reduce the number of dev dependencies.
- Focus on natively supporting ES modules in browsers and in Node.
- Increase the usefulness of CI testing.
- Don't break the current compatibility of `compat.js` builds.

#### The Status Quo and Its Problems

- We currently don't support ES modules in Node.js.
  - As of Node.js 13.2, only files ending in `.mjs` or files under a folder hierarchy with a `package.json` with `{"type": "module"}` can be `imported` rather than `required`.
  - We have neither.
- Our build pipeline is complex.
  - We use Rollup to bundle our files.
    - It's not really needed for using Fluent in Node.js and as a dependency for larger projects with their own build pipelines.
    - OTOH, a single-file package is easy to include Fluent in jsfiddle, codepen etc.
      - Then again, all of modern browsers support `<script type="module">`.
  - We use Babel to transpile the `compat.js` build.
    - We use `preset-env` to define what should be transpiled. The browser matrix was chosen not based on usage data, but based on the JS features we didn't want to transpile. That's kind of backwards of how `preset-env` should work.
  - In #376 we're talking about adding TypeScript to the mix, which would further complicate the pipeline, testing, debugging etc.
- We write tests using ES modules and make them work in Node by patching `require`.
  - This is done using `@babel/plugin-transform-es2015-modules-commonjs` or `esm`.
- We test code in different version of Node (good), but we do it by transpiling our unit tests with a Babel setup which is different from the one which we use for shipping (bad).

#### Ideas

I'd like the clean-up to be careful and incremental. In no particular order:

- [ ] Add integration tests.
  - Create a new test suite using our `compat.js` builds and run it in different versions of Node and older browsers without any further transpilation.
  - Switch our unit tests to only run in the currently active version of Node.js.
- [ ] Drop `preset-env` and transpile to `ES2017` which we effectively do right now anyways.
- [ ] Drop `esm` and `@babel/plugin-transform-es2015-modules-commonjs` when running tests.
  - This requires that we use a test runner which natively supports ES modules. Work is underway in both [mocha](https://github.com/mochajs/mocha/pull/4038) (which we currently use) and [jest](https://github.com/facebook/jest/issues/4842) to enable this.
[...]

PR #436, opened 2019-12-02
Migrate @fluent/bundle to TypeScript

stasm:
See #376 for the discussion. As an experiment, I ported `@fluent/bundle` to TypeScript. Migrating the files in `src/` was a breeze; I did run into a few troubles with tests and packaging, however.

#### Summary

- Files in `src/` are now written in TypeScript.
- `make compile` compiles them into corresponding JS files in `esm/`.
  - This also produces type definitions.
  - `tsc` is a project-based compiler which compiles many files at once. While it's possible to run it on a file-by-file basis, it's slower to do so and some features of TS won't work (e.g. `const enum`). I added an empty file at `esm/.compiled` which is touched every time `tsc` runs to indicate to `make` that the compiled files are up to date compared to the source in `src`.
- I kept all test files in JS.
  - They import from the `esm/` directory, which means that the project needs to be compiled before tests are run.
  - `make test` takes care of that automatically.
- The other option for running tests was to use a Babel plugin for stripping TypeScript syntax from files in `src` at the moment when the tests are run.
  - My own preference (at least for now) was to introduce a two-step process of first compiling the source and then running the tests. The compilation can fail early due to typing errors which shortens the feedback loop.
  - I also liked the idea of running tests on actual compiled files in the `esm` folder rather than using one tool to compile them for shipping (`tsc`) and another to run tests (`@babel/preset-typescript`).
- I used `typedoc` to produce API docs.
- I didn't touch other parts of the build pipeline.
  - I considered using `tsc`'s `target: "es2017"` to produce the `compat.js` build, but in the end decided that it would be best to limit the number of moving parts in the PR.
  - The `compat.js` build is still produced with Rollup and Babel, as it used to be.
  - I filed #435 to discuss a long term clean-up effort which could end up with us dropping Babel in favor of `tsc`'s `target: "es2017"`.

#### Work Remaining
- [x] I still need to configure the linter.

#### API Changes
- I renamed the `FluentType` class to `FluentBaseType`, and then I re-introduced `FluentType` as a type alias in TS: `type FluentType = string | FluentBaseType`. I'm not sure about the naming, however, and I'm open to alternatives. 
- (Added on Dec 9) I removed `FluentError`. It was only used in the runtime parser. In the resolver we use native errors: `RangeError`, `ReferenceError`, and `TypeError`, so I changed all uses of `FluentError` to the native `SyntaxError`.


fluent-bundle/src/bundle.ts
```diff
@@ -1,12 +1,33 @@
...
-export default class FluentBundle {
+export class FluentBundle {
+  public locales: Array<string>;
+
+  public _terms: Map<string, RuntimeTerm> = new Map();
+  public _messages: Map<string, RuntimeMessage> = new Map();
+  public _functions: Record<string, CustomFunction>;
+  public _useIsolating: boolean;
+  public _transform: CustomTransform;
+  public _intls: WeakMap<any, any> = new WeakMap();
```
stasm: What's a good way to type the `Intl` formatters here? A union of all existing ones sounds like something which eventually becomes out of date.

stasm:
@blushingpenguin, @huy-nguyen, @Seally -- would you be interested in taking a look at this PR and giving me hints about how to improve it? Thanks!

fluent-bundle/src/builtins.ts
```diff
@@ -0,0 +1,67 @@
...
+    throw new TypeError("DATETIME accepts exactly one argument");
+  }
+  let arg = args[0];
+
+  if (arg instanceof FluentNone) {
+    return new FluentNone(`DATETIME(${arg.valueOf()})`);
+  }
+
+  if (arg instanceof FluentNumber) {
+    return new FluentDateTime(arg.valueOf(), { ...arg.opts, ...values(opts) });
```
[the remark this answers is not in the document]
stasm: I'm having second thoughts about this.

In TypeScript, this works because [`Intl.DateTimeFormatOptions`](
https://github.com/microsoft/TypeScript/blob/2ab1b71fff3bf54d5150ed29ddc71217d0018ad5/lib/lib.es5.d.ts#L4273-L4287) happens to be assignable to [`Intl.NumberFormatOptions`](
https://github.com/microsoft/TypeScript/blob/2ab1b71fff3bf54d5150ed29ddc71217d0018ad5/lib/lib.es5.d.ts#L4236-L4247). I'm not sure if we should rely on this.

On a higher level, this is a spec question: what should be the outcome of `NUMBER($date)` and `DATETIME($number)`? I think it will be safer to not make assumptions about the answer right now and wait for the formatting spec to be finished.

I'll revert the `FluentDateTime extends FluentNumber` change.

fluent-bundle/src/builtins.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  if (arg instanceof FluentNone) {
+    return new FluentNone(`NUMBER(${arg.valueOf()})`);
+  }
+
+  if (arg instanceof FluentNumber || arg instanceof FluentDateTime) {
+    return new FluentNumber(arg.valueOf(), { ...arg.opts, ...values(opts) });
+  }
+
+  throw new TypeError("Invalid argument to NUMBER");
```
stasm: Unfortunately, the amount of changes in this file makes diff think it's a new file. Here's what `NUMBER` used to look like:

```js
export function NUMBER([arg], opts) {
  if (arg instanceof FluentNone) {
    return new FluentNone(`NUMBER(${arg.valueOf()})`);
  }

  let value = Number(arg.valueOf());
  if (Number.isNaN(value)) {
    throw new TypeError("Invalid argument to NUMBER");
  }

  return new FluentNumber(value, { ...arg.opts, ...values(opts) });
}
```

This code didn't compile in TS because `arg.opts` doesn't exist on `FluentBaseType`; it's only defined on `FluentNumber` and `FluentDateTime`. BTW. the JS code works because `{...undefined}` evaluates to `{}`.

In the TS rewrite, I introduced explicit `arg instanceof FluentNumber || arg instanceof FluentDateTime` guards and dropped the `Number()` parsing. This changes the current behavior of `@fluent/bundle`: calling `NUMBER` on a string-typed variable used to parse the string as a number and format it. After this change, the argument will be rejected as invalid.

fluent-bundle/src/scope.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  reportError(error: FluentError) {
+    if (!this.errors) {
+      throw error;
+    }
+    this.errors.push(error);
+  }
+
+  memoizeIntlObject<ObjectT, OptionsT>(
```
Seally: Assuming we keep `FluentBundle.prototype._intls` to `WeakMap<any, any>`, we may be able to drop `ObjectT` from the generics list and just use `any`.

fluent-bundle/src/bundle.ts
Seally: I'm afraid I got nothing. :(

I've tried a couple of ideas. The stricter ones got stuck being incompatible with some other types. Other, more relaxed types seem marginally better than `any` and might not be worth it.

I can tell that the type `NonNullable<Parameters<typeof Intl.DateTimeFormat>[1]>` will give `Intl.DateTimeFormatOptions` (works for `Intl.NumberFormat` too), but I couldn't figure out how to get TypeScript to infer `typeof value` from usage. And that's about the only potentially useful thing that came out of my fiddling with the typings here and in the `Scope` class.

stasm: Thanks. I went with `WeakMap<object, Record<string, object>>` for now which is only marginally better than `any`.

fluent-bundle/src/scope.ts
stasm: I like the fact that with generics here, `scope.memoizeIntlObject(Intl.PluralRules)` is correctly typed as returning an instance of `Intl.PluralRules`, so I think I'll leave it as it is right now.

stasm:
I got to a point where this PR is now ready for review. I'd like to flag both @Pike for a general review, and @zbraniecki, who has typing experience from working on `fluent-rs`.

I added explicit return types to all functions and methods and I like how it improves the reading experience when browsing the code, in particular outside the IDE. I have a few naming questions; I asked them in inline comments.

Pike:
Some ad-hoc comments from the first phase of my review:

I tried to map the syntax ebnf to ast.ts, and struggled for two reasons:

* the ordering of types and variants is different
* the `Runtime` prefix dominates my ability to discover concepts I know.

The latter might have been exaggerated by the fact that I was looking at expressions, which means everything is called `RuntimeFooExpression`, which doesn't blend well with how humans read.

Or at least I read that they read by "first few characters and last few characters and sum of characters inbetween".

fluent-bundle/src/resolver.ts
```diff
@@ -139,6 +172,7 @@ function VariableReference(scope, {name}) {
       if (arg instanceof Date) {
         return new FluentDateTime(arg.getTime());
       }
+    // eslint-disable-next-line no-fallthrough
```
Gregoor: Is this one really needed here. From my understanding it shouldn't because the `return` prevents fallthrough?

fluent-bundle/src/resolver.ts
```diff
@@ -198,13 +238,25 @@ function TermReference(scope, {name, attr, args}) {
...
+  { name, args }: RuntimeFunctionReference
+): FluentType {
   // Some functions are built-in. Others may be provided by the runtime via
   // the `FluentBundle` constructor.
-  const func = scope.bundle._functions[name] || builtins[name];
+  let func = scope.bundle._functions[name];
   if (!func) {
-    scope.reportError(new ReferenceError(`Unknown function: ${name}()`));
-    return new FluentNone(`${name}()`);
+    switch (name) {
```
Gregoor: I think I liked the old way better, but I guess the way TS does exports it's not as easy to achieve?

fluent-bundle/src/resolver.ts
stasm: The `return` statement is guarded by an `if`, and the fall-through is desired here: if it's an object but not a `Date`, fall through to `default` and report an error.

fluent-bundle/src/resolver.ts
stasm: Yeah, I was getting an error about `typeof import(...)` not having an index signature. A workaround which I found was to use `let func = scope.bundle._functions[name] || (<any>builtins)[name];`, which is shorter and much closer to the old way. Do you think it's a better choice here than the explicit switch?

fluent-bundle/src/builtins.ts
stasm: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1605062 to prepare for this change in `mozilla-central`.

stasm:
I updated this PR to merge cleanly into the current master.

fluent-bundle/src/resolver.ts
Pike: ... or add `NUMBER` and `DATE` to `_functions` in the bundle constructor?

fluent-bundle/src/resource.ts
```diff
@@ -238,14 +268,15 @@ export default class FluentResource {
...
-        elements[lastIndex] = trim(elements[lastIndex], RE_TRAILING_SPACES);
+      if (typeof lastElement === "string") {
+        elements[lastIndex] = trim(lastElement, RE_TRAILING_SPACES);
       }
 
-      let baked = [];
+      let baked: Array<PatternElement> = [];
       for (let element of elements) {
-        if (element.type === "indent") {
+        if (typeof element !== "string" && element.type === "indent") {
```
Pike: Could this be `instanceof Indent` ?

fluent-bundle/src/resolver.ts
Gregoor: > 
> 
> Yeah, I was getting an error about `typeof import(...)` not having an index signature. A workaround which I found was to use `let func = scope.bundle._functions[name] || (<any>builtins)[name];`, which is shorter and much closer to the old way. Do you think it's a better choice here than the explicit switch?

Yeah no I guess I agree that the switch is better than that, because we don't lose type info that way. Another option which imo beats switch would be sth like:
```typescript
let func = scope.bundle._functions[name] || (
    name == "DATETIME" || name == "NUMBER" ? builtins[name] : null
);
```
There should also be a more generic version to achieve the type refinement necessary to use `name` as an index, but I'm not coming up with it rn 🤔

fluent-bundle/src/ast.ts
```diff
@@ -0,0 +1,86 @@
...
+export type MessageReference = {
+  type: "mesg";
+  name: string;
+  attr: string | null;
+};
+
+export type FunctionReference = {
+  type: "func";
+  name: string;
+  args: Array<Expression>;
```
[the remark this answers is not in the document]
stasm: Yep, thanks.

fluent-bundle/src/ast.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export type ComplexPattern = Array<PatternElement>;
+
+export type PatternElement = string | Expression;
+
+export type Indent = {
+  type: "indent";
+  value: string;
+  length: number;
+};
```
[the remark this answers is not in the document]
stasm: I made `Indent` be a class and moved it into `resource.ts`.

fluent-bundle/src/resolver.ts
stasm: I like @Pike's suggestion to deal with this when the bundle is constructed rather than later on, when translations are resolved. I implemented it in https://github.com/projectfluent/fluent.js/pull/436/commits/4d48ced773d600d0ecbcf51961539ef254b88231.

fluent-bundle/src/bundle.ts
```diff
@@ -1,12 +1,13 @@
...
 import { FluentResource } from "./resource.js";
-import { FluentNone, FluentType, FluentFunction } from "./types.js";
+import { FluentValue, FluentNone, FluentFunction } from "./types.js";
 import { Message, Term, Pattern } from "./ast.js";
+import { NUMBER, DATETIME } from "./builtins.js";
 
 export type TextTransform = (text: string) => string;
 
-export type FluentArgument = string | number | Date | FluentType;
+export type FluentArgument = string | number | Date | FluentValue;
```
Pike: Do you want to keep doubling the `string` here? Or should we do a `FluentType<any>` here?

fluent-bundle/src/bundle.ts
```diff
@@ -54,7 +55,7 @@ export class FluentBundle {
   constructor(
     locales: string | Array<string>,
     {
-      functions = {},
+      functions,
```
Pike: This means that `functions` is now a required option? Or is this the `undefined` spread? I recall you mentioning that that doesn't work in typescript?

Also, does this have type information?

fluent-bundle/src/bundle.ts
stasm: It's redundant because of the implementation detail, but I kept it because semantically I wanted to say: an argument can either be:

  - an instance of `FluentValue`, or
  - a JS primitive or a `Date`, in which case we'll take care of converting it to a proper `FluentValue`.

Do you think it's worth keeping this distinction in code?

fluent-bundle/src/bundle.ts
stasm: That's a bit hard to read, but the types are defined in the lines below this one. It's due to how destructuring of parameters works in JS.

```ts
constructor(
    locales: string | Array<string>,
    {
      functions,
      useIsolating = true,
      transform = (v: string): string => v
    }: {
      functions?: Record<string, FluentFunction>;
      useIsolating?: boolean;
      transform?: TextTransform;
    } = {}
  ) { .. }
```

You can see that `function?` is possibly `undefined`. Both JS and TS have no problem with spreading it: `({...undefined})` evals to an empty object.

fluent-bundle/src/bundle.ts
Pike: The semantics of `FluentType<any>` could be:

A JS primitive or a `Date`, or a custom FluentType, or one with options.

I don't mind either way, it's just that it might say just slightly different things?

fluent-bundle/src/bundle.ts
Pike: Ah, thanks.

fluent-bundle/src/bundle.ts
stasm: What if I expressed it as follows? Does it communicate the semantics better?

```ts
type NativeArgument = string | number | Date;
export type FluentArgument = FluentValue | NativeArgument;
```

Pike: Yeah, that's good.

stasm: Thanks!

stasm:
I have verified that `@fluent/bundle` built from TypeScript works correctly in a React app (The Fluent Playground) and in Firefox (as `Fluent.jsm`).

PR #438, opened 2019-12-03
Exclude <title> from HTML sanitization

zbraniecki:
Fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1599864

### DIFF
--- a/fluent-dom/src/overlay.js
+++ b/fluent-dom/src/overlay.js
@@ -67,7 +67,11 @@ export default function translateElement(element, translation) {
   const {value} = translation;
 
   if (typeof value === "string") {
-    if (!reOverlay.test(value)) {
+    if (element.localName === "title" &&
+        element.namespaceURI === "http://www.w3.org/1999/xhtml") {
+      // A special case for the HTML title element whose content must be text.
+      element.textContent = value;
+    } else if (!reOverlay.test(value)) {
       // If the translation doesn't contain any markup skip the overlay logic.
       element.textContent = value;
     } else {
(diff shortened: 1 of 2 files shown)


Pike:
Is `title` really the only element that we should tackle here? Do we need to check for namespaces?

stasm:
I think we should check the namespace in the PR, and then open a new issue to remove the XUL support from `@fluent/dom`, since IIUC we use `L10nOverlays.cpp` in Gecko anyways. Which I also think means that this PR is not really relevant to the bug in question? Still, thanks for submitting the change upstream, @zbraniecki.

Other than `<title>`, I'm not sure if there are other HTML elements which must only contain text content. I'm `Ctrl+F`ing the [HTML5 spec](https://html.spec.whatwg.org/) for `Content Model`, and I see: `<style>` (which maybe should be excluded from the overlays logic entirely), `<rp>`, and `<textarea>`, as well as  [`<time>`](https://html.spec.whatwg.org/#the-time-element) and [`<option>`](https://html.spec.whatwg.org/#the-option-element), but only under some conditions.

While we're at it, it might also be a good idea to protect elements with the ["nothing" content model](https://html.spec.whatwg.org/#the-nothing-content-model), most of which are void elements. We already do it [in `@fluent/react`](https://github.com/projectfluent/fluent.js/blob/f8cc0f45428e72afa89c0600b24186896e38e4d1/fluent-react/src/localized.js#L141-L147).

I'm not sure how far we want to go enforcing HTML's content model rules. They are quite complex: https://html.spec.whatwg.org/#elements-3. If we don't enforce them at all, or don't enforce them too strictly, I would still be in favor of making an exception for `<title>` since it's such an important and basic HTML element.

zbraniecki:
Updated to add namespace check.

PR #440, opened 2019-12-11
stop using `componentWillReceiveProps()` (#425)

spacemeowx2:


stasm:
Thank you for the PR. We're re-writing the provider in #406, which is close to being complete and will also fix the `componentWillReceiveProps` warning.

spacemeowx2:
That's cool! Can't wait to see the new release

PR #441, opened 2019-12-16
Use Object.prototype.hasOwnProperty.call

### LINKED ISSUE #428: Cannot pass variable named `hasOwnProperty`

stasm:
Fix #428.


Pike:
I like the idea, I'm wondering if we should use `Object.create(null)` for the `scope.params`?

stasm:
I've been thinking about this, and the same about named arguments to functions. But in the end, the `hasOwnProperty` check makes the distinction between `{}` and `Object.create(null)` irrelevant, I think?

Pike:
Probably.

stasm:
I used `Object.create(null)` for named parameters after all. It can't hurt and might actually help when writing custom functions.

PR #445, opened 2020-01-10
Remove unneeded Babel plugins

stasm:
Since we require Node.js 10 now, the following Babel plugins are not necessary:

  - `@babel/plugin-proposal-async-generator-functions`
  - `@babel/plugin-proposal-object-rest-spread`

I also removed `@babel/plugin-transform-modules-commonjs` which we used in tests, in favor of a simpler to use `esm`.


stasm:
@Pike Cleanups continued. These changes are related to Node.js 8 EOL. Would you like to take a look? I don't think there's anything tricky here, tbh.

babel_config.js
```diff
@@ -23,11 +23,5 @@ export default {
         "transform-regenerator"
       ],
     }]
-  ],
-  plugins: [
-    // Shipped in Firefox 57, Chrome 63
-    "@babel/plugin-proposal-async-generator-functions",
-    // Shipped in Firefox 55, Chrome 60
-    "@babel/plugin-proposal-object-rest-spread"
```
stasm: I'm not really sure why we included these plugins here. I introduced them in #193, but then I don't know why I didn't remove them in #225 a few weeks later. After all, it would seem that the `preset-env` config would have sufficed?

PR #446, opened 2020-01-14
Make ReactLocalization the value of FluentContext

stasm:
The value of `FluentContext` is now an instance of `ReactLocalization` which gained the `parseMarkup` method.

I also exposed `FluentContext` as part of the public API in case someone prefers to use it via `FluentContext.Consumer` or the `useContext` hook. In fact, I think we should declare these methods as preferrable to the `withLocalization` decorator, which suffers from not hoisting static members of the inner component, if any are present. #430 has a fix for it but perhaps we could live with this shortcoming of `withLocalization` if other methods are available.

### DIFF
--- a/fluent-react/src/context.js
+++ b/fluent-react/src/context.js
@@ -1,7 +1,4 @@
 import { createContext } from "react";
 import ReactLocalization from "./localization";
 
-export default createContext({
-  l10n: new ReactLocalization([]),
-  parseMarkup: null
-});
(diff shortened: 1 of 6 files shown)


stasm:
@Gregoor Would you mind taking a look at this?

Gregoor:
The changes look good to me 👍

stasm:
Thanks :)

PR #450, opened 2020-01-27
Migrate @fluent/sequence to TypeScript

stasm:
See #376.


stasm:
With this PR, I've fleshed out the process of migrating our packages to TS. It's not a lot of work; I hope other packages will follow shortly. @Pike Would you like to take a look, please?

fluent-sequence/src/map_async.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export function mapBundleAsync(
+  bundles: AsyncIterable<FluentBundle>,
+  ids: string
+): Promise<FluentBundle | null>
+
+export function mapBundleAsync(
+  bundles: AsyncIterable<FluentBundle>,
+  ids: Array<string>
+): Promise<Array<FluentBundle | null>>;
```
Pike: Is there documentation on what these do, in combination with the impl below??

stasm: They're function overloads: https://www.typescriptlang.org/docs/handbook/functions.html#overloads. They allow specifying extra typing constraints when the implementation is quite loose. E.g. "if you pass an array, expect an array as a result".

PR #453, opened 2020-01-30
(@fluent/react) Don't call createParseMarkup() in the default context

### LINKED ISSUE #452: @fluent/react v0.11.0 breaks SSR
I tried to fix this but it's more involved than I initially thought.

`ReactContext` always initializes a `ReactLocalization` with no params. This in turn triggers a `createMarkupParser()` call which requires a `document` and blows up during SSR:
https://github.com/projectfluent/fluent.js/blob/c7c39dadb8adeb47a542ddb9e3821dfec5a6f38d/fluent-react/src/context.js#L4
https://github.com/projectfluent/fluent.js/blob/c7c39dadb8adeb47a542ddb9e3821dfec5a6f38d/fluent-react/src/localization.js#L17

In summary, I can't seem to avoid this error during SSR:
https://github.com/projectfluent/fluent.js/blob/c7c39dadb8adeb47a542ddb9e3821dfec5a6f38d/fluent-react/src/markup.js#L10-L17

stasm:
Allow passing `null` as `parseMarkup` to disable the support for any markup in translations.

Fix #452.

### DIFF
--- a/fluent-react/src/context.js
+++ b/fluent-react/src/context.js
@@ -1,4 +1,4 @@
 import { createContext } from "react";
 import ReactLocalization from "./localization";
 
-export default createContext(new ReactLocalization([]));
+export default createContext(new ReactLocalization([], null));
--- a/fluent-react/src/localized.js
+++ b/fluent-react/src/localized.js
@@ -135,7 +135,7 @@ function Localized(props) {
 
   // If the message value doesn't contain any markup nor any HTML entities,
   // insert it as the only child of the wrapped component.
-  if (!reMarkup.test(messageValue)) {
+  if (!reMarkup.test(messageValue) || l10n.parseMarkup === null) {
     return cloneElement(child, localizedProps, messageValue);
   }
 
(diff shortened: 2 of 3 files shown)


stasm:
@rlr Do you maybe have some way to verify this fix before I merge it?

rlr:
sure. I’ll try it tomorrow. Thanks!

fluent-react/src/context.js
```diff
@@ -1,4 +1,4 @@
 import { createContext } from "react";
 import ReactLocalization from "./localization";
 
-export default createContext(new ReactLocalization([]));
+export default createContext(new ReactLocalization([], null));
```
stasm: @Gregoor Do you think this is the right approach for the default context value?

rlr:
@stasm this patch fixes the issue for me. r+ thanks!

stasm:
@rlr Thanks for verifying. I'll land and publish the fix later today.

fluent-react/src/context.js
Gregoor: I think I'd initialize it with `null` in the constructor of `ReactLocalization` instead, imo that class is more concerned with that detail than the context.

PR #461, opened 2020-03-27
(@fluent/react) Pass instances of ReactLocalization to LocalizationProvider

stasm:
This PR changes the API from ↓

```ts
<LocalizationProvider bundles={bundlesIterable} />
```

to ↓

```ts
let l10n = new ReactLocalization(bundlesIterable);
<LocalizationProvider l10n={l10n} />
```

In the old API, the `LocalizationProvider ` would automatically create a `ReactLocalization` instance. In the new API, the instance must be explicitly created by the developer. The reasoning behind this change are as follows:

- It's reasonable to expect that consumers will store `bundlesIterable` in their app's state.
- The app might re-render multiple times given the same state, and we should then produce the same result for each render.
- When `bundlesIterable` is an array, this works as expect.
- When `bundlesIterable` is a generator, it keeps its own state, and re-renders might change it, or even deplete it. 
  - The current implementation of `LocalizationProvider` internally creates a new `ReactLocalization` using `bundlesIterator` and uses it as the value for the `FluentContext`.
  - The new value of the context triggers re-renders of `Localized` descendants. 
  - The bundles yielded by the generator are cached in the internal `ReactLocalization` instance, but that instance is then thrown away when `LocalizationProvider` re-renders.
  - This results in `Localized` translating the app into a fallback language, or into the fallbacks stored in the source.
- React's `PureComponent` and `memo` might offer a away to remedy this situation, but I haven't been able to get them to work reliably. More importantly, the React docs are very clear about the fact that `memo` should not be used to fix these kind of bugs:

    > This method only exists as a performance optimization. Do not rely on it to “prevent” a render, as this can lead to bugs.

    And a similar note about the `useMemo` hook:

    > You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

- In contrast to generators, instances of `ReactLocalization` are safe to store in the app's state. Internally, the cached bundles yielded from `bundlesIterator` persist as long as the same `ReactLocalization` instance is used for all re-renders.

Interestingly, with this PR, we might not even need `LocalizationProvider` at all. All it's doing is render the `FluentContext.Provider`, which we could expose in the public API of `@fluent/react`:

```ts
let l10n = new ReactLocalization(bundlesIterable);
<FluentContext.Provider value={l10n}>
    <App />
</FluentContext.Provider>
```

### DIFF
--- a/fluent-react/src/provider.ts
+++ b/fluent-react/src/provider.ts
@@ -25,14 +25,6 @@ interface LocalizationProviderProps {
  * under the provider.
  */
 function LocalizationProvider(props: LocalizationProviderProps): ReactElement {
-  if (props.l10n === undefined) {
-    throw new Error("LocalizationProvider must receive the l10n prop.");
-  }
-
-  if (!(props.l10n instanceof ReactLocalization)) {
-    throw new Error("The bundles prop must be an iterable.");
-  }
-
   return createElement(
     FluentContext.Provider,
     {
@@ -44,30 +36,7 @@ function LocalizationProvider(props: LocalizationProviderProps): ReactElement {
 
 LocalizationProvider.propTypes = {
   children: PropTypes.element.isRequired,
-  l10n: isReactLocalization,
+  l10n: PropTypes.instanceOf(ReactLocalization).isRequired,
 };
-
-function isReactLocalization(
-  props: Record<string, unknown>,
-  propName: string,
-  componentName: string
-): null | Error {
-  const prop = props[propName];
-
-  if (!prop) {
-    return new Error(
-      `The ${propName} prop supplied to ${componentName} is required.`
-    );
-  }
-
-  if (prop instanceof ReactLocalization) {
-    return null;
-  }
-
-  return new Error(
-    `The ${propName} prop supplied to ${componentName} must be an instance \
-of ReactLocalization.`
-  );
-}
 
 export const MemoLocalizationProvider = memo(LocalizationProvider);
--- a/fluent-react/src/provider.ts
+++ b/fluent-react/src/provider.ts
@@ -1,4 +1,4 @@
-import { createElement, ReactNode, ReactElement, memo } from "react";
+import { createElement, ReactNode, ReactElement } from "react";
 import PropTypes from "prop-types";
 import { FluentContext } from "./context";
 import { ReactLocalization } from "./localization";
@@ -24,7 +24,9 @@ interface LocalizationProviderProps {
  * `l10n` prop. This instance will be made available to `Localized` components
  * under the provider.
  */
-function LocalizationProvider(props: LocalizationProviderProps): ReactElement {
+export function LocalizationProvider(
+  props: LocalizationProviderProps
+): ReactElement {
   return createElement(
     FluentContext.Provider,
     {
@@ -38,5 +40,3 @@ LocalizationProvider.propTypes = {
   children: PropTypes.element.isRequired,
   l10n: PropTypes.instanceOf(ReactLocalization).isRequired,
 };
-
-export const MemoLocalizationProvider = memo(LocalizationProvider);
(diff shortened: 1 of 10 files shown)


fluent-react/src/provider.ts
```diff
@@ -66,13 +60,14 @@ function isIterable(
...
 
   return new Error(
-    `The ${propName} prop supplied to ${componentName} must be an iterable.`
+    `The ${propName} prop supplied to ${componentName} must be an instance \
+of ReactLocalization.`
   );
 }
 
-export let MemoLocalizationProvider = memo(LocalizationProvider);
+export const MemoLocalizationProvider = memo(LocalizationProvider);
```
stasm: I kept `memo` here for now, but I think it's not needed. If I'm reading the code right, [`Context.Provider` is already memoized](https://github.com/facebook/react/blob/9d67847f7b41e408016cfa16450323b3fee8bc56/packages/react-reconciler/src/ReactFiberBeginWork.js#L2694), effectively.

fluent-react/test/provider_valid.test.js
```diff
@@ -28,29 +28,37 @@ describe("LocalizationProvider - validation", () => {
...
     act(() => {
-      renderer = renderer.update(<LocalizationProvider bundles={bundles} />);
+      renderer = renderer.update(
+        <LocalizationProvider l10n={l10n}>
+          <MockComponent />
+        </LocalizationProvider>
+      );
     });
-    expect(spy).toHaveBeenCalledTimes(1);
+    expect(mockFn).toHaveBeenCalledTimes(1);
```
stasm: This test currently fails. I'm not sure how to test that the tree under the `Context.Provider` hasn't re-rendered. @Gregoor, do you have any ideas for how to test it?

stasm:
@Gregoor, what do you think about this?

stasm:
> Interestingly, with this PR, we might not even need `LocalizationProvider` at all. All it's doing is render the `FluentContext.Provider`, which we could expose in the public API of `@fluent/react`:
> 
> ```ts
> let l10n = new ReactLocalization(bundlesIterable);
> <FluentContext.Provider value={l10n}>
>     <App />
> </FluentContext.Provider>
> ```

This [tweet](https://twitter.com/kentcdodds/status/1229460648276328448) explains that the above wouldn't be a good for performance if consumers had to use it instead of the dedicated `LocalizationProvider`.

fluent-react/test/provider_valid.test.js
Gregoor: Don't have a fluent fork running atm, so this is a little speculative. Shouldn't you be able to do?

```jsx
const child = <MockComponent />;
let renderer = TestRenderer.create(
  <LocalizationProvider l10n={l10n}>
    {child}
  </LocalizationProvider>
);
act(() => {
  renderer = renderer.update(
    <LocalizationProvider l10n={l10n}>
      {child}
    </LocalizationProvider>
  );
});
expect(mockFn).toHaveBeenCalledTimes(1);
```
Which hopefully should be true and reflect the way it's being used?

stasm: Ah yes, thank you, this works! I spent more time looking at this test that I'm willing to admit, and somehow I missed the fact that I was passing a new `<MockComponent />` in the `update` call... 🤦‍♂️

stasm: That said, I'm not sure if this test tests what I'd like it to test :) I'll dig into this tomorrow.

Pike:
Should the tests keep the localization outside of the `<LocalizationProvider>` element? Just to set a good example? Also, of course, examples will need an update, that's still coming?

stasm:
I think it's OK to leave the tests the way they are right now. As far as examples are concerned, unfortunately they're quite outdated even right now. I'll update them once this lands.

PR #462, opened 2020-03-30
Migrate @fluent/langneg to TypeScript

stasm:
See #376.


fluent-langneg/src/negotiate_languages.ts
```diff
@@ -1,23 +1,8 @@
-import filterMatches from "./matches";
+import {filterMatches} from "./matches";
 
-function GetOption(options, property, type, values, fallback) {
```
stasm: I removed this function because I don't think the implementation of `@fluent/langneg` needs to be 100% defensive. For TypeScript consumers, TS already gives us some of those guaranties. For non-TS consumers, we don't usually verify all inputs in other `@fluent` packages as thoroughly as this function did it.

fluent-langneg/src/negotiate_languages.ts
```diff
@@ -63,38 +48,32 @@ function GetOption(options, property, type, values, fallback) {
...
+    strategy = "filtering",
+    defaultLocale,
+  }: NegotiateLanguagesOptions = {}
+): Array<string> {
 
   const supportedLocales = filterMatches(
-    resolvedReqLoc,
-    resolvedAvailLoc, strategy
+    Array.from(Object(requestedLocales)).map(String),
+    Array.from(Object(availableLocales)).map(String),
```
stasm: I'm tempted to remove the string and array normalization here as well, because again, I think we can make reasonable assumptions about the quality of the input. There are tests which fail if I do that, however.

fluent-langneg/test/langneg_test.js
```diff
@@ -111,11 +111,11 @@ const data = {
...
 suite('Language Negotiation', () => {
+  for (const strategy in data) {
+    for (const groupName in data[strategy]) {
+      const group = data[strategy][groupName];
 
-  test('test suite matches', () => {
-    for (const strategy in data) {
-      for (const groupName in data[strategy]) {
-        const group = data[strategy][groupName];
+      test(`${strategy} - ${groupName}`, () => {
```
stasm: I moved the `test` call into the nested for loop so that each case is run as a separate Mocha test, resulting in better test summary and error logging.

stasm:
@zbraniecki Would you have a moment to review this? Other than adding types, I also made a few small changes to the implementation to make it a bit less defensive.

stasm:
Thanks for the review, @zbraniecki!

PR #464, opened 2020-03-31
(@fluent/bundle) Forbid setting NUMBER's currency

stasm:
If one of the named arguments passed into `NUMBER` is `currency`, we should remove it from the options bag. Right now this PR just ignores it silently, but perhaps it would be better to explicitly throw in `NUMBER` instead. @Pike, wdyt?

I also wonder if we should disallow `NUMBER`'s `style`, as well. Are there scenarios when we'd want the localization to choose a different style (e.g. `percent`) than the base (e.g. `currency`)? I guess one such example is when the base is not explicitly defined in the source, in which case `decimal` is used. This can happen in the declarative bindings where there's no way to pass an instance of `FluentNumber` (like in `DOMLocalization`).

### DIFF
--- a/fluent-bundle/src/builtins.ts
+++ b/fluent-bundle/src/builtins.ts
@@ -18,10 +18,15 @@ import {
   FluentDateTime
 } from "./types.js";
 
-function values(opts: Record<string, FluentValue>): Record<string, unknown> {
+function values(
+  opts: Record<string, FluentValue>,
+  exclude?: string
+): Record<string, unknown> {
   const unwrapped: Record<string, unknown> = {};
   for (const [name, opt] of Object.entries(opts)) {
-    unwrapped[name] = opt.valueOf();
+    if (exclude === undefined || name !== exclude) {
+      unwrapped[name] = opt.valueOf();
+    }
   }
   return unwrapped;
 }
@@ -37,7 +42,10 @@ export function NUMBER(
   }
 
   if (arg instanceof FluentNumber || arg instanceof FluentDateTime) {
-    return new FluentNumber(arg.valueOf(), { ...arg.opts, ...values(opts) });
+    return new FluentNumber(arg.valueOf(), {
+      ...arg.opts,
+      ...values(opts, "currency")
+    });
   }
 
   throw new TypeError("Invalid argument to NUMBER");
--- a/fluent-bundle/src/builtins.ts
+++ b/fluent-bundle/src/builtins.ts
@@ -18,13 +18,30 @@ import {
   FluentDateTime
 } from "./types.js";
 
-function values(opts: Record<string, FluentValue>): Record<string, unknown> {
-  const unwrapped: Record<string, unknown> = {};
+function values(
+  opts: Record<string, FluentValue>,
+  allowed: Array<string>
+): Record<string, unknown> {
+  const unwrapped: Record<string, unknown> = Object.create(null);
   for (const [name, opt] of Object.entries(opts)) {
-    unwrapped[name] = opt.valueOf();
+    if (allowed.includes(name)) {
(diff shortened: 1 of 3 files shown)


stasm:
I added a second commit which changes the silent error behavior to an explicit one: the `NUMBER` built-in throws when one of the options is named `currency`. While I think it's a good idea to be more strict about this now, I dislike how it makes `NUMBER(1234, currency: "EUR")` produce `NUMBER()` (because the entire builtin throws) while `NUMBER(1234, style: "currency")` produces an un-formatted `1234` because the `TypeError: Currency code is required with currency style.` happens later in `FluentNumber.toString()`.

Pike:
Oh rabbit hole, where are you when I need you? Oh, nice, there you are.

I'd go as far as to whitelist valid arguments. But that's hard for backwards compat. The story is full of sadness, like plain calls to `NUMBER($FOO)` in various places, 🤷‍♂ .

Note, there's already pre-existing documentation on restricting the options in https://projectfluent.org/fluent/guide/functions.html.

I'd blacklist `unit` and `notation`, additionally, and whitelist `unitDisplay`.

stasm:
Nice documentation find. It looks it was written a long time ago, in a rather wishful tone :)

If I'm understanding correctly, you're suggest to do the following?

- Throw when a forbidden option is passed to `NUMBER()`:

        localeMatcher
        style
        currency
        unit
        notation

- Ignore all options other than the known ones: 

        numberingSystem (not sure about this one)
        currencyDisplay
        unitDisplay
        useGrouping
        minimumIntegerDigits
        minimumFractionDigits
        maximumFractionDigits
        minimumSignificantDigits
        maximumSignificantDigits

- Don't require the known options so that `NUMBER($foo)` continues to work.

I like this, but I'm afraid of breaking a combination of `data-l10n-args` and `style` somewhere where I'm not aware of them being used:

    foo = Progress: {NUMBER($prog, style="percent")}
    <span data-l10n-id="foo" data-l10n-args='{"prog": 0.75}' />

FWIW, I haven't found anything like that in gecko-strings not in l10n-central.

stasm:
I changed the approach again to explicitly listing all allowed options, and silently ignoring all other options. This approach gives us good fallback for both unknown and forbidden options, and lets us strictly control allowed options in the future, too.

fluent-bundle/src/builtins.ts
```diff
@@ -37,12 +54,35 @@ export function NUMBER(
...
   }
 
   throw new TypeError("Invalid argument to NUMBER");
 }
 
+const DATETIME_ALLOWED = [
+  "dateStyle",
+  "timeStyle",
+  "fractionalSecondDigits",
+  "calendar",
```
Pike, line 70: I'm torn on this one. This sounds like a scary thing for localizers to modify, in particular when they're doing it inconsistently.

I'd go with restricting this for now.

fluent-bundle/src/builtins.ts
```diff
@@ -37,12 +54,35 @@ export function NUMBER(
...
+  "dayPeriod",
+  "numberingSystem",
```
Pike, line 72: Yeah, I'm torn on this one, too. I'd go with restrictive for now, and not allow this one as well (same for `NUMBER`).

fluent-bundle/src/builtins.ts
```diff
@@ -37,12 +54,35 @@ export function NUMBER(
...
+  "hour12",
+  "hourCycle",
```
Pike, line 74: Given this changes the meaning of `10`, I'd exclude this one, too.

stasm:
Thanks, @Pike. I agree about being more restrictive in terms of which options are allowed; I pushed the changes you requested. Good points about the tests, too.

PR #465, opened 2020-04-03
(@fluent/react) Use FluentContext.Consumer instead of useContext

stasm:
The way hooks are implemented in React, they require that all components
use the same instance of React, likely due to a shared global state used
to store components' state. Unfortunately, this approach doesn't work
well when `@fluent/react` is installed into a project via `npm link`,
for instance for testig purposes.

We don't make a heavy use of hooks. In fact, we only currently use the
`useContext` hook which can be easily replaced by `Context.Consumer` and
a render prop, thus making testing a lot easier.

References:
    https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react
    https://github.com/facebook/react/issues/13991

### DIFF
--- a/fluent-react/src/localized.ts
+++ b/fluent-react/src/localized.ts
@@ -5,12 +5,12 @@ import {
   cloneElement,
   createElement,
   isValidElement,
-  useContext
 } from "react";
 import PropTypes from "prop-types";
 import voidElementTags from "../vendor/voidElementTags";
 import { FluentContext } from "./context";
 import { FluentArgument } from "@fluent/bundle";
+import { ReactLocalization } from "./localization";
 
 // Match the opening angle bracket (<) in HTML tags, and HTML entities like
 // &amp;, &#0038;, &#x0026;.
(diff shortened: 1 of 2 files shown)


stasm:
@Gregoor I've had some issue with testing `@fluent/react` in other projects via `npm link` due to how hooks work internally. What do you think about removing the `useContext` hook in favor of a more tranditional `<FluentContext.Consumer>{l10n => ...}</FluentContext.Consumer>`?

Gregoor:
I guess this wouldn't affect the propspects of consumers of Fluent being able to use hooks in the future?

I can't say I have a good grasp on the issue at hand yet, i.e. I wonder if having multiple versions of React is not a problem in and of itself, that should be solved instead? But given my current limited understanding, you might've already evaluated that and deemed that to be infeasible?

Internally the only case I'd have for hooks is that I find them more readable as it doesn't introduce an additional scope. If you see testing benefits in the <Consumer> approach, that might just trump it 🎺

stasm:
> I guess this wouldn't affect the propspects of consumers of Fluent being able to use hooks in the future?

That's a good question. Initially I thought the answer would be _it wouldn't affect them_. This morning I realized, however, that if we add, say, a `useLocalization` hook in the future and if that hook makes use of `useContext`, then we're back to square one.

> I can't say I have a good grasp on the issue at hand yet, i.e. I wonder if having multiple versions of React is not a problem in and of itself, that should be solved instead? But given my current limited understanding, you might've already evaluated that and deemed that to be infeasible?

I've only run into this problem when trying to use an `npm linked` `@fluent/react` in some other codebase. The problem stems from the fact that there's one `node_modules/react` in the `fluent.js` working directory where I work on the `@fluent/react` code, and another one in `node_modules` of the othter codebase which I wanted to test changes to `@fluent/react` on. 

This sounds like a bug in the design of React Hooks, and it might be that it would be impossible to completely fix it in React.

There are two other work-arounds which I found:

- Build `@fluent/react` locally and copy it into `node_modules/@fluent/react` of the other codebase. This needs to be repeated after every change.

- Edit `package.json` of `@fluent/react` and point React deps to the other project's ones. (Or the other way around.)

    ```diff
    -    "react": "^16.13.1",
    -    "react-dom": "^16.13.1"
    +    "react": "file:../node_modules/react/",
    +    "react-dom": "file:../node_modules/react-dom/"
    ```

stasm:
After a bit more testing, I think the approach no. 1 from my comment above is the most fail-proof way of testing, despite the inconvenience of having to copy files after each change. I found an [article](https://medium.com/@vcarl/problems-with-npm-link-and-an-alternative-4dbdd3e66811) explaining how one can use `npm pack` and `npm install` to streamline the process a little bit.

stasm:
I'm going to close this without merging. Let's assume we can continue to use hooks as normal in `@fluent/react`'s codebase.

PR #471, opened 2020-04-09
(@fluent/gecko) Prune obsolete build targets

stasm:
Currently, only the following targets are still used in Gecko:

  - `FluentSyntax.jsm` is used in tests,
  - `fluent-react.js` is used in Devtools.

I wondered about `Fluent.jsm` -- it's still used on the release channel in Firefox 75 today. If there's a any problem with it in the nex few weeks, I'll bring the `Fluent.jsm` target back.


fluent-gecko/makefile
```diff
@@ -7,52 +7,20 @@ version = $(1)@$(shell node -e "\
 	console.log(require('../$(1)/package.json').version)")
 
 test:
```
Pike, line 9: Should this be an empty recipe? https://www.gnu.org/software/make/manual/html_node/Empty-Recipes.html

fluent-gecko/makefile
```diff
@@ -66,5 +34,8 @@ clean:
 	@rm -rf .nyc_output coverage
 	@echo -e " $(OK) clean"
 
-lint: _lint
-html: _html
+lint:
+	@eslint --config $(ROOT)/eslint_src.json --max-warnings 0 src/
+	@echo -e " $(OK) lint"
+
+html:
```
Pike, line 41: Same here.

fluent-gecko/makefile
stasm: Ah, thanks, I forgot about this syntax.

stasm: I also have a Makefule cleanup in store, but that's for later :)

stasm:
@zbraniecki Would you mind taking a quick look at this? It's just some removals that I'd like you to be aware of. Thanks.

zbraniecki:
lgtm!

Issue #467, opened 2020-04-05
Add React Hooks


macabeus:
`fluent-react` provides a simpler way to use fluent on a React application, but it's laking a built-in function to get messages using react hooks.
Currently is possible to get the messages just using a HOC using `withLocalization` or with `<Localized />`.

It's a similar approach that [`react-i18next`](https://react.i18next.com/latest/usetranslation-hook) already do.

Sometimes is better just to call a function instead of increasing the JSX code wrapping each message with the `<Localized />` component. For example, [here](https://github.com/mozilla/activate.mozilla.community/blob/master/src/components/Header.js#L82-L84):
```jsx
<Localized id="nav-campaigns">
  <span></span>
</Localized>

<!-- would be just... -->
<span>{t('nav-campaigns')}</span>
```

Related: https://github.com/projectfluent/fluent.js/issues/182

stasm:
Thanks for your work on this, @macabeus! I released `@fluent/react` 0.13 last week with your changes: https://github.com/projectfluent/fluent.js/blob/master/fluent-react/CHANGELOG.md#fluentreact-0130-july-2-2020. I also added a note about `useLocalization` on the wiki: https://github.com/projectfluent/fluent.js/wiki/useLocalization

PR #475, opened 2020-04-11
(@fluent/react) Add useTranslate

macabeus:
Closes this issue: https://github.com/projectfluent/fluent.js/issues/467

Add `useTranslate` custom hook, that uses the `FluentContext` to return two functions to get the translated messages from Fluent's bundle.
- `t`: get the root message from the given id
- `tAttributes`: get the attributes message from the given id

This PR also updates one example file on `@fluent/react` in order to show how to use that.

### DIFF
--- a/fluent-react/example/src/Hello.tsx
+++ b/fluent-react/example/src/Hello.tsx
@@ -1,29 +1,23 @@
 import React, { useState } from "react";
-import { Localized } from "@fluent/react";
+import { useTranslate } from "@fluent/react";
 
 export function Hello() {
     let [userName, setUserName] = useState("");
+    const { t, tAttributes } = useTranslate()
 
     return (
         <div>
             {userName ?
-                <Localized id="hello" vars={{ userName }}>
-                    <h1>{'Hello, { $userName }!'}</h1>
-                </Localized>
-                :
-                <Localized id="hello-no-name">
-                    <h1>Hello, stranger!</h1>
-                </Localized>
+                <h1>{t('hello', { userName })}</h1>
+                : <h1>{t('hello-no-name')}</h1>
             }
 
-            <Localized id="type-name" attrs={{ placeholder: true }}>
-                <input
-                    type="text"
-                    placeholder="Type your name"
-                    onChange={evt => setUserName(evt.target.value)}
-                    value={userName}
-                />
-            </Localized>
+            <input
+                type="text"
+                placeholder={tAttributes('type-name').placeholder}
+                onChange={evt => setUserName(evt.target.value)}
+                value={userName}
+            />
         </div>
     );
 }
--- a/fluent-react/example/src/Hello.tsx
+++ b/fluent-react/example/src/Hello.tsx
@@ -1,23 +1,29 @@
 import React, { useState } from "react";
-import { useTranslate } from "@fluent/react";
+import { Localized } from "@fluent/react";
 
 export function Hello() {
     let [userName, setUserName] = useState("");
-    const { t, tAttributes } = useTranslate()
 
     return (
         <div>
             {userName ?
-                <h1>{t('hello', { userName })}</h1>
-                : <h1>{t('hello-no-name')}</h1>
+                <Localized id="hello" vars={{ userName }}>
+                    <h1>{'Hello, { $userName }!'}</h1>
+                </Localized>
+                :
+                <Localized id="hello-no-name">
+                    <h1>Hello, stranger!</h1>
+                </Localized>
             }
 
-            <input
-                type="text"
-                placeholder={tAttributes('type-name').placeholder}
-                onChange={evt => setUserName(evt.target.value)}
-                value={userName}
-            />
+            <Localized id="type-name" attrs={{ placeholder: true }}>
+                <input
+                    type="text"
+                    placeholder="Type your name"
+                    onChange={evt => setUserName(evt.target.value)}
+                    value={userName}
+                />
+            </Localized>
         </div>
     );
 }
--- a/fluent-react/src/useTranslate.ts
+++ b/fluent-react/src/useTranslate.ts
@@ -1,0 +1,96 @@
+import { useContext } from "react";
+import { FluentContext } from "./context";
+import { ReactLocalization } from "./localization";
+import { FluentArgument, FluentBundle } from "@fluent/bundle";
+
+type Message = Exclude<ReturnType<FluentBundle["getMessage"]>, undefined>
+
+// Using ReactLocalization and a id, return its bundle and message object
+// if both exists, otherwise return null for boths
+type getBundleAndMessage = (
+  (l10n: ReactLocalization, id: string) =>
+  ({ bundle: FluentBundle; msg: Message } | { bundle: null; msg: null })
+)
+const getBundleAndMessage: getBundleAndMessage = (l10n, id) => {
+  const bundle = l10n.getBundle(id);
+  const msg = bundle?.getMessage(id);
+
+  if (bundle === null || msg === undefined) {
+    return { bundle: null, msg: null };
+  }
+
+  return { bundle, msg };
+};
+
+type TFunction = (id: string, vars?: Record<string, FluentArgument>) => string
+type TAttributesFunction = (
+  (id: string, vars?: Record<string, FluentArgument>) =>
+  { [key in string]: string }
+)
+
+type useTranslate = () => { t: TFunction; tAttributes: TAttributesFunction }
+
+/*
+ * The `useTranslate` hook uses the FluentContext to return two functions
+ * to get the translated messages from Fluent's bundle:
+ * - t: get the root message from the given id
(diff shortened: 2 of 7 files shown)


fluent-react/example/src/Hello.tsx
```diff
@@ -1,29 +1,23 @@
...
-                <input
-                    type="text"
-                    placeholder="Type your name"
-                    onChange={evt => setUserName(evt.target.value)}
-                    value={userName}
-                />
-            </Localized>
+            <input
+                type="text"
+                placeholder={tAttributes('type-name').placeholder}
```
macabeus, line 17: I don't know if add a function `tAttributes` is the best way, but I think that it's enough.
At the first moment I tried adding a parameter on `t` function to return the attributes (`t('type-name'), undefined, { attributes: true })`), but it's very weird and bad to type.

fluent-react/src/useTranslate.ts
```diff
@@ -0,0 +1,96 @@
...
+export const useTranslate: useTranslate = () => {
+  const l10n = useContext(FluentContext);
+
+  if (!l10n) {
+    // Return fallbacks functions
+    const t: TFunction = () => "";
+    const tAttributes: TAttributesFunction = () => ({});
+
+    return { t, tAttributes };
+  }
```
macabeus, line 48: I'm returning an empty string and empty object because I'm following the same approach of `<Localized />`, but maybe would be better to return the `id`.
On the case of `tAttributes`, we could use Proxy to return the attribute name. For example:

```js
tAttributes('invalid-key').foo // should return `"foo"`
```

stasm:
Thank you for the PR, @macabeus! I've been thinking about adding a hook-based API too, and it's great to see a contribution in this area. I'd like to spend a bit of time discussing the API design and the use-cases, and then I'll be happy to do a code review. CC'ing @Gregoor who re-wrote `@fluent/react` to function components which use hooks internally.

----

The API I had in mind was a bit different, and I wonder you think about it. Rather than return `t` and `tAttributes`, could we directly return the `ReactLocalization` instance? It already has the `getString` method which is similar to your `t` function. We could then do:

```js
let l10n = useLocalization();
alert(l10n.getString("hello"));
```

Or even:

```js
let l10n = useContext(FluentContext);
alert(l10n.getString("hello"));
```

What I like in this approach is that it mostly relies on abstractions already built into React and into `@fluent/react`. I usually try to look for solutions which avoid adding more code :)

----

I also have a suggestion regarding the `tAttributes` function. In Fluent, querying individual attributes alone shouldn't be needed. In other words, the most common operation is either querying the value of the message (like in `getString), or querying the value _and_ the attributes.

In `@fluent/dom` we have two functions corresponding to these two use-cases: `valueFromBundle` (which is functionally equivalent to `@fluent/react`'s `getString`, and `messageFromValue`, which is implemented like this:

https://github.com/projectfluent/fluent.js/blob/cff95f6f42338974e90e58607c9aaa87c92021d3/fluent-dom/src/localization.js#L215-L235

Perhaps we could add a similar method to `ReactLocalization` to allow formatting the value _and_ the attributes in one go?

----

 That said, I might suggest startig with just `getString` for now. For more complex usage, I think wrapping components in `<Localized>` should be the preferred way of localizing them together with attributes.

macabeus:
Thank you for commenting on the PR. And yeah, I also want to discuss more about the API =]

<hr />

> Rather than return `t` and `tAttributes`, could we directly return the `ReactLocalization` instance? It already has the `getString` method which is similar to your `t` function.

I didn't notice about `ReactLocalization`.
I'm making some prototypes following this approach, and yeah, I think that returning a `ReactLocalization` would be enough.

I like the mindset of creating a new feature with few new line of codes, but:
- I think that the name `getString` is too big. Is very common to use many times the function that translate a string, and because of that is a common convention to call this function just as `t`. A smaller name is good on this case. Maybe we could rename `getString` to `t`?
- I'm a little afraid to `useTranslate` just returning `ReactLocalization` instance instead of a plain object, because it'll would bad if we want to change its return on future to returning something more. Do you think that it could be a problem?

<hr />

> Perhaps we could add a similar method to `ReactLocalization` to allow formatting the value _and_ the attributes in one go?

Yeah, adding a new method that returns everything would be nice. Looks like a better approach following the fluent mindset.

stasm:
> I think that the name getString is too big. Is very common to use many times the function that translate a string, and because of that is a common convention to call this function just as t. A smaller name is good on this case. Maybe we could rename getString to t?

I think `t` is a good choice when the function is expected to be used a lot, like in i18next. For `@fluent/react`, I think we should stick to composing components via `<Localized>` and only use `getString` in exceptional scenarios, when the imperative API is the only way to solve a problem. The way I think about the API I'd like to add here is that it's a hook equivalent of the [`withLocalization`](https://github.com/projectfluent/fluent.js/wiki/withLocalization) HOC, catering to the same use-caes.

That's why I'd suggest to only focus on the `getString` use-case. For attributes, let's leave that out of this PR, please. We might want to add a method for formatting an entire message, together with its attributes, later on, but for now, I'd like to focus on `<Localized>` as the work horse.

> I'm a little afraid to useTranslate just returning ReactLocalization instance instead of a plain object, because it'll would bad if we want to change its return on future to returning something more. Do you think that it could be a problem?

Hmm, that's an interesting point. I think you're right that it might be safer to return a plain object, like so:

```js
let {l10n} = useLocalization();
```

In the future, we might want to add more properties, for example:

```js
let {l10n, changeLocales} =  useLocalization();
```

I'd like to think about this a little bit and see what's possible. Let's not block this PR on this; both approaches will be fine :)

macabeus:
@stasm Okay. I just changed the PR to `useTranslate` return an instance of `ReactLocalization` (and a fallback object if there isn't that, following the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern)).

Also I changed the example to just replace `withLocalization` with the new hook.

fluent-react/src/useTranslate.ts
```diff
@@ -0,0 +1,28 @@
+import { useContext } from "react";
+import { FluentContext } from "./context";
+import { ReactLocalization } from "./localization";
+
+// Implementation of null object pattern for ReactLocalization
+const fallbackL10n: ReactLocalization = {
```
stasm, line 6: Is this actually needed? I understand and agree with the intent, but it also seems to me that `useContext` is guaranteed to return an empty `ReactLocalization` thanks to how it's initialized here:

https://github.com/projectfluent/fluent.js/blob/c27994d75a8a708540a3b631617a271ec48253b9/fluent-react/src/context.ts#L4

fluent-react/src/useTranslate.ts
```diff
@@ -0,0 +1,28 @@
...
+  bundles: [],
+  parseMarkup: null,
+  reportError: () => undefined,
+};
+
+/*
+* The `useTranslate` hook returns the FluentContext
+*/
+type useTranslate = () => { l10n: ReactLocalization }
+export const useTranslate: useTranslate = () => {
```
stasm, line 18: Would you mind renaming this to `useLocalization`, to keep it in sync with the `withLocalization` HOC?

fluent-react/src/useTranslate.ts
```diff
@@ -0,0 +1,28 @@
...
+  const l10n = useContext(FluentContext);
+
+  if (!l10n) {
+    return { l10n: fallbackL10n };
+  }
+
+  return { l10n };
+};
+
+export default useTranslate;
```
stasm, line 28: Let's not mix default exports with named exports. In fact, I try to avoid default exports in TypeScript altogether.

fluent-react/src/useTranslate.ts
macabeus: Okay, I removed that. I didn't notice that we are setting a default value on `createContext`

fluent-react/src/useTranslate.ts
macabeus: Okay. Makes sense to me.

macabeus:
@stasm I made the changes that you proposed.
Also I added some very simple tests case for `useLocalization`. I added just three new tests because I'm avoiding to add redundant tests.

fluent-react/test/use_localization.test.js
```diff
@@ -0,0 +1,57 @@
...
+  test("render outside of a LocalizationProvider", () => {
+    const renderer = TestRenderer.create(<DummyComponent />);
+    expect(renderer.toJSON()).toMatchInlineSnapshot(`
+      <p>
+        foo
+      </p>
+    `);
+  });
+
+  test("useLocalization should exposes getString from ReactLocalization", () => {
```
stasm, line 37:
```suggestion
  test("useLocalization exposes getString from ReactLocalization", () => {
```

stasm:
@macabeus Let me know when you're ready for getting this merged. I'll also update the docs on the wiki once this is released. Thanks again!

macabeus:
@stasm It's fine to me to merge =]
Thank you

PR #480, opened 2020-04-21
Index.cjs

stasm:
This is an attempt to use `type: "module"` combined with [conditional exports](https://nodejs.org/api/esm.html#esm_conditional_exports) to ship both CommonJS and ES modules code.


eemeli:
Closing in favour of #653.

PR #485, opened 2020-05-18
Use GitHub Actions

stasm:
Let's try something simple first: one workflow per package, using `make dist`.

### DIFF
--- a/.github/workflows/fluent-bundle.yml
+++ b/.github/workflows/fluent-bundle.yml
@@ -1,0 +1,22 @@
+name: "@fluent/bundle"
+on: [push, pull_request]
+jobs:
+  dist:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [10.x, 12.x, 14.x]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{matrix.node}}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{matrix.node}}
+      - name: Install build tools
+        run: npm ci
+      - name: Install package dependencies
+        working-directory: ./fluent-bundle
+        run: npm ci
+      - name: make dist
+        working-directory: ./fluent-bundle
+        run: make dist
--- a/.github/workflows/fluent-bundle.yml
+++ b/.github/workflows/fluent-bundle.yml
@@ -1,5 +1,9 @@
 name: "@fluent/bundle"
-on: [push, pull_request]
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
 jobs:
   dist:
     runs-on: ubuntu-latest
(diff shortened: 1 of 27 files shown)


stasm:
The actions pass in my fork: https://github.com/stasm/fluent.js/actions?query=branch%3Aactions

.github/workflows/fluent-bundle.yml
```diff
@@ -0,0 +1,22 @@
+name: "@fluent/bundle"
+on: [push, pull_request]
```
Pike, line 2: I'd restrict these, similar to what we've done in `python-fluent`. That way it's obvious what should happen on pushes to `gh-pages`. Or maybe just exclude `gh-pages`? I haven't figured out how much the action config of one branch blends others. Maybe not even remotely as much as I thought it would.

Also, do you want to start restricting when automation is run for what now or in a follow-up?

stasm: I pushed a commit which restricts the workflows to pushes to master and PRs against master.

stasm:
> Would you want to have wider compat-ranges on dependencies, and test those?
> 
> That would conflict with the use of `npm ci`, for one. Also, you could prep that in the same way we did for python-fluent, and add a matrix for that already.

Can you explain what you mean by this? What kind of matrix does `python-fluent` define for this purpose?

Pike:
> Can you explain what you mean by this? What kind of matrix does `python-fluent` define for this purpose?

In https://github.com/projectfluent/python-fluent/blob/9f88c05094fbe34899775df19184589b962a42fa/.github/workflows/fluent.runtime.yml#L28 I'm testing across all the versions of `fluent-syntax` listed. Right now, just one. If you look at the resulting jobs, https://github.com/projectfluent/python-fluent/runs/678936556 shows that it's running the tests against various python versions and syntax 0.17.0.

stasm:
Oh, interesting! I think I must have missed this last week.

I think this could be useful for testing different versions of React in `@fluent/react`, but I'm concerned about the list of version getting out of date quickly.

Pike:
In `python-fluent`, we have comments in a few places that ask developers to sync dependency and automation configuration. Not that I read those comments last time I changed something ;-).

On a positive note, dependency information for node is machine readable, so one could check. I started thinking about moving the test deps into `setup.cfg`, but there's a yak in that rabbit hole.

I ended up not worrying about that too much, as the job names are pretty explicit about what they test. Could even make that expliciter by putting `fluent.syntax==0.17.0` into the matrix value instead of just the version number.

stasm:
Hmm, I used the _workflow file path_ variant from https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#example-using-a-workflow-file-path because the workflow names have funny characters in them. It looks like it doesn't work.

I'm surprised to see that I can just use `/` verbatim in the URL there. I'll change the paths to use names rather than paths.

Issue #486, opened 2020-05-18
macos make fails on simple node commands


Pike:
macos is super eager to not pass commands to the shell at all, if it thinks they're simple.

Thus the PATH logic doesn't work for things like `tsc`, or *some* invocations of eslint. Yes, *some*.

There are at least three ways to fix these:

```
lint-1:
    @: ; eslint --config $(ROOT)/eslint_test.json --max-warnings 0 test/
lint-2:
    @eslint --config $(ROOT)/eslint_test.* --max-warnings 0 test/
lint-3:
    @../.node_modules/.bin/eslint --config $(ROOT)/eslint_test.json --max-warnings 0 test/
```

`common.mk` could probably also define `ESLINT` and `TSC` to the path in `lint-3`, which would double-fix the issue, probably, as the shell variable would trigger the shell, and it'd be a real path.

@stasm, opinions?

stasm:
I'm leaning towards option no. 3: `../.node_modules/.bin/eslint `. It's simple (as in _the opposit of _indirect_), explicit, self-explanatory, and the visual noise isn't much of a problem, because we don't spend too much time looking at Makefiles anyways. I can work on a PR if you agree.

Pike:
SGTM. I'll make a patch, 'cause I need to test which things pass and which fail anyway

Pike:
This is fixed by #487

PR #487, opened 2020-05-18
Build system fixes for macos (#486)

Pike:
On macos, GNU make eagerly avoids calling into the shell,
so exposing node commands through $PATH doesn't work half of the
time.

I've put two alternatives up in the PR to give some context.

In #486, you mentioned you'd prefer absolute paths. I've done that in `fluent-dedent`. I've done it for commands that need it, so `tsc` and `eslint`. I've done it consistently for `eslint`, though only one needs it. I've not done it for `rollup` or `nyc`. Looking at my changes, I'm not fond.

I've also thrown out the start of doing `$(ESLINT)` in `fluent-dom`. Turns out that's actually much more *make*ier, now that I think about it.

@stasm, mind if I go for the latter consistently, and also drop the `$(ROOT)/node_modules/.bin` from `$PATH`?

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -4,6 +4,7 @@
 ROOT := $(dir $(lastword $(MAKEFILE_LIST)))
 SOURCES := $(wildcard src/*)
 VERSION := $(shell node -pe "require('./package.json').version")
+DOC_DESTINATION := $(subst @fluent, ../html, $(PACKAGE))
 
 export SHELL := /bin/bash
 ESLINT ?= $(ROOT)node_modules/.bin/eslint
@@ -19,6 +20,15 @@ ROLLUP_CMD = $(ROLLUP) $(CURDIR)/esm/index.js \
 	--name $(GLOBAL) \
 	--output.format umd \
 	--output.file \
+	$(NULL)
+
+TYPEDOC_CMD = $(TYPEDOC) src \
+	--out $(DOC_DESTINATION) \
+	--mode file \
+	--excludeNotExported \
+	--excludePrivate \
+	--logger none \
+	--hideGenerator \
 	$(NULL)
 
 # Common maintenance tasks.
--- a/common.mk
+++ b/common.mk
@@ -31,6 +31,12 @@ TYPEDOC_CMD = $(TYPEDOC) src \
 	--hideGenerator \
 	$(NULL)
 
+MOCHA_CMD =@$(NYC) --reporter=text --reporter=html $(MOCHA) \
+	--recursive --ui tdd \
+	--require esm $(TEST_REQUIRES) \
+	test/**/*_test.js \
+	$(NULL)
+
 # Common maintenance tasks.
 .PHONY: clean lint test build html
 
@@ -47,14 +53,5 @@ deps:
 depsclean:
 	@rm -rf node_modules
 	@echo -e " $(OK) deps clean"
-
-# Shared recipes
-.PHONY: mocha-test
-
-mocha-test:
-	@$(NYC) --reporter=text --reporter=html $(MOCHA) \
-	    --recursive --ui tdd \
-	    --require esm $(TEST_REQUIRES) \
-	    test/**/*_test.js
 
 OK := \033[32;01m✓\033[0m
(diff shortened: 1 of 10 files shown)


stasm:
My hope was that with time we could remove `common.mk` completely. I don't like the fact that the build system lives between three different Makefiles for each package.

My suspicion about the _makiness_ is that it comes from the need/desire to allow toolchain overrides like `CC=clang`. I don't think we'd need this for `eslint` or even `tsc`. But I do see the value of declaring build tools executables upfront at the beginning of the Makefile. Could we do that in package-specific Makefiles rather than in `common.mk`?

I'll be happy to see `export PATH` gone completely, if possible :)

Pike:
I never knew that `mocha` was actually a executable command, and not an option ;-)

Anyway, this passes locally now, and on Travis.

I'm not really sure if `mocha-test` should also be `$(MOCHA_CMD)`.

Given we talked about changing our docs, I like that they're now built consistently.

In all those factorizations, I left the top-level target untouched, and only called into a shared workflow from there, so that local customizations can be done w/out a global change.

I've split this into a few commits to allow for individual decisions.

fluent-dom/makefile
```diff
@@ -5,31 +5,27 @@ DEPS    := cached-iterable:CachedIterable
...
 	@echo -e " $(OK) esm/ compiled"
 
 .PHONY: test
 test: esm/.compiled
-	@nyc --reporter=text --reporter=html mocha \
-	    --recursive --ui tdd \
-	    --require esm \
-	    --require ./test/index \
-	    test/**/*_test.js
+	@$(MAKE) mocha-test TEST_REQUIRES="--require ./test/index"
```
stasm, line 22: Perhaps drop `TEST_REQUIRES` and leave the full command here as a special case, same as you did in `fluent-react`?

common.mk
```diff
@@ -1,12 +1,35 @@
...
+ROLLUP ?= $(ROOT)node_modules/.bin/rollup
+TYPEDOC ?= $(ROOT)node_modules/.bin/typedoc
+
+ROLLUP_CMD = $(ROLLUP) $(CURDIR)/esm/index.js \
+	--banner "/* $(PACKAGE)@$(VERSION) */" \
+	--amd.id $(PACKAGE) \
+	--name $(GLOBAL) \
+	--output.format umd \
+	--output.file \
+	$(NULL)
```
stasm, line 23: What is the `$(NULL)` for?

common.mk
```diff
@@ -25,4 +48,13 @@ depsclean:
 	@rm -rf node_modules
 	@echo -e " $(OK) deps clean"
 
+# Shared recipes
+.PHONY: mocha-test
+
+mocha-test:
```
stasm, line 54: I guess this is a target rather than `MOCHA_CMD` because you wanted to be able to resolve `TEST_REQUIRES` lazily? If instead we use a custom `test` target in `fluent-dom` (see my other suggestion), that problem would go away.

common.mk
Pike: "Best practices", they said. https://searchfox.org/mozilla-central/search?q=%24%28NULL%29&path=&case=false&regexp=false is where I got this from.

Also allows you to add a line at the end w/out having to change the line before, as they all end in `\`.

common.mk
Pike: It's a target 'cause I done this one first, and then the other ones didn't work as targets.

I'm pretty confident that this would work just as well as a `CMD`. Which is why I kinda looked at it with an angle when pushing.

I think I should just align that.

common.mk
stasm: Thanks, that makes sense.

fluent-dom/makefile
Pike: I left this in. This is a pretty common practice to customize build recipes in the world of make.

Pike:
I was torn on whether I should call it `MOCHA_CMD` or `NYC_CMD`, but in the end I figured we're running mocha tests inside a nyc test runner.

If that's just me and the thing I learned new while writing this patch, I'm happy to change that.

common.mk
```diff
@@ -1,12 +1,41 @@
...
+	--mode file \
+	--excludeNotExported \
+	--excludePrivate \
+	--logger none \
+	--hideGenerator \
+	$(NULL)
+
+MOCHA_CMD =@$(NYC) --reporter=text --reporter=html $(MOCHA) \
+	--recursive --ui tdd \
+	--require esm $(TEST_REQUIRES) \
```
stasm, line 36: Nit: perhaps call it `MOCHA_REQUIRES`?

Issue #500, opened 2020-07-06
fluent-react: setL10n in useLocalization?


leesiongchan:
Can we have `setL10n` in `useLocalization` so we can easily set the new value in a nested component?

stasm:
Thanks for filing the issue! I've been thinking about something similar: `let {l10n, changeLanguages} = useLocalization();` but I think `setL10n` might be easier. @macabeus What do you think?

macabeus:
I'm thinking... And yeah, maybe it could be useful.

_On the client side_, using the example from [`fluent-react`](https://github.com/projectfluent/fluent.js/blob/f41d910dd3bbf13ac057026060a0666426796870/fluent-react/example/src), on `l10n.tsx` we'll need to _something like_:
```diff
- <LocalizationProvider l10n={l10n}>
+ <LocalizationProvider l10n={l10n} setL10n={changeLocales}>
```

Then we'll be able to use `const { setL10n } = useLocalization()` anywhere. It could be useful if you want to create one or more components on deeper level to change the language. Nice.

Just adding the prop `setL10n` on `<LocalizationProvider />` and updating the value of `FluentContext` from `l10n` to an object `{ l10n, setL10n }` is the most simple implementation.
But let’s brainstorm. Maybe we can to do something even better:

Currently `setL10n` receives as parameter an array. Is it make sense on the use case from a hook? Maybe a string could be simpler and fit better (`setL10n('pt')` instead of `setL10n(['pt'])`)?
Also, if you set an unknown language (`setL10n(['blaah'])`), it'll fails silently... is it the best behaviour on this case?

Also and most important IMHO... Currently is necessary to write a big boilerplate to use Fluent. `l10n.ts` has 78 lines!
I think that we could improve that setting a default `setL10n` function on `LocalizationProvider`. So the client won't need to write again a `setL10n` function and pass it to `<LocalizationProvider />` - only if they want to use a custom `setL10n`, of course.

What do you think about that, @stasm ?

stasm:
> Just adding the prop `setL10n` on `<LocalizationProvider />` and updating the value of `FluentContext` from `l10n` to an object `{ l10n, setL10n }` is the most simple implementation.

Yeah, I was thinking along the same lines. That's why I liked your idea in #475 to return a plain `{l10n}` object from the hook.

> Maybe a string could be simpler and fit better (`setL10n('pt')` instead of `setL10n(['pt'])`)?

[...]

PR #501, opened 2020-07-08
(@fluent/react) Add changeLocales and currentLocales on provider

macabeus:
Closes this issue: https://github.com/projectfluent/fluent.js/issues/500

This PR aims to add on `useLocalization` a function to change the current locale, `changeLocales`. So on a nested component will can to updates the locale. It's useful for a "choice language" selector.

In addition to returning `changeLocales` on `useLocalization`, I also added to return `currentLocales`, in order to avoid two sources of truth. So a "choice language" selector will need just to read the current language from the provider.

I also updated the example adding the component "choice language".


fluent-react/src/context.ts
```diff
@@ -1,4 +1,10 @@
 import { createContext } from "react";
 import { ReactLocalization } from "./localization";
 
-export let FluentContext = createContext(new ReactLocalization([], null));
+const defaultValue = {
+  l10n: new ReactLocalization([], null),
+  changeLocales: (_changeLocales: string[]) => undefined as void,
+  currentLocales: [] as string[],
```
macabeus, line 7: Since we have a setter and a reader, maybe we could call they as `locales` and `setLocales`?

fluent-react/src/provider.ts
```diff
@@ -27,10 +29,21 @@ interface LocalizationProviderProps {
 export function LocalizationProvider(
   props: LocalizationProviderProps
 ): ReactElement {
+  let [locales, setLocales] = useState(props.initialLocales);
+
+  function changeLocales(locales: string[]) {
+    props.changeLocales(locales);
+    setLocales(locales);
```
macabeus, line 36: There aren't any check if the `prop.changeLocales` was succeeded before to call `setLocales`. Maybe we could return a bool on `changeLocales`? Or use `await props.changeLocales(locales)` in order to fail on except?

fluent-react/src/use_localization.ts
```diff
@@ -5,9 +5,9 @@ import { ReactLocalization } from "./localization";
 /*
 * The `useLocalization` hook returns the FluentContext
 */
-type useLocalization = () => { l10n: ReactLocalization }
+type useLocalization = () => { l10n: ReactLocalization, changeLocales: (locales: string[]) => void, currentLocales: string[] }
```
macabeus, line 8: I'm writing the type `(locales: string[]) => void` 2~3 times. Maybe we could to name this type and export/import from somewhere?

PR #504, opened 2020-07-17
Add per locale Intl memoizer

Demivan:


fluent-bundle/src/memoizer.ts
```diff
@@ -0,0 +1,14 @@
+type IntlCache = WeakMap<object, Record<string, object>>;
```
macabeus, line 1: Just a question: why is better to use a `WeakMap` here instead of just a plain object?

Demivan: I don't know :) Kept it as it is right now.
Seeing that map keys are `Intl.NumberFormat` and `Intl.DateTimeFormat` and will not be garbage collected it should be regular `Map` instead

stasm: That's a good point. @zbraniecki, do you remember why we went with a `WeakMap` here?

Pike: I didn't fully finish my doodling, but it seems that the keys are objects and not strings?

Demivan:
Sorry for pinging. Anything need to be changed for this to be merged?
I can see that this pull requests fixes same issue #254 tries to fix. If that is preferred API I can change this pull request to match #254 API.

zbraniecki:
@Demivan we're currently focused on finishing the Rust translation of this library and those PRs for JS are a bit on a backburner. I'm sorry for it. Within a month we'll start looking again at them and hope to start addressing all the backlog here.

In the meantime, if you have a suggestion wrt. the `WeakMap`, feel free to take over the PR :)

fluent-bundle/src/memoizer.ts
Demivan: I now added commit that switches from `WeakMap` to `Map` as keys that are used here cannot be garbage collected anyway.

Demivan:
Hi.
Any chance this gets merged before next release? Anything need to be changed? I have rebased this PR and fixed conflicts.
Is there planned date for next release?

eemeli (approved):
Sorry for the delay; finally got around to looking at this. And it's good! I think it might make sense to eventually separate the memoiser even further form the bundle, as I at least find it a little surprising how the class, locale arg, and options arg are now all handled in very different parts of the API. But that'd be a larger change than this, which makes sense just by itself.

PR #513, opened 2020-08-10
@fluent/syntax: Don't drop to newline when serializing patterns starting with special chars

### LINKED ISSUE #512: @fluent/syntax: Variant values with `[]` don't roundtrip through the parser/serializer
The following Fluent snippet doesn't round-trip right:

```properties
shared-photos =
    { $userName } { $photoCount ->
        [one][x]added a new photo { $item ->
                [bucket] to the bucket
               *[folder] to the folder
            }
       *[other] added { $photoCount } new photos
    }
```

The result is

```properties
shared-photos =
    { $userName } { $photoCount ->
        [one]
            [x]added a new photo { $item ->
                [bucket] to the bucket
               *[folder] to the folder
            }
       *[other] added { $photoCount } new photos
    }
```
Filing here, but it affects all of `js`, `python`, and `kotlin`.

stasm:
Fix #512.

### DIFF
--- a/fluent-syntax/src/serializer.ts
+++ b/fluent-syntax/src/serializer.ts
@@ -13,6 +13,29 @@ function isSelectExpr(elem: AST.PatternElement): boolean {
     && elem.expression instanceof AST.SelectExpression;
 }
 
+function shouldStartOnNewLine(pattern: AST.Pattern): boolean {
+  const isMultiline =
+    pattern.elements.some(isSelectExpr) ||
+    pattern.elements.some(includesNewLine);
+
+  if (isMultiline) {
+    const firstElement = pattern.elements[0];
+    if (firstElement instanceof AST.TextElement) {
+      const firstChar = firstElement.value[0];
+      // Due to the indentation requirement these text characters may not appear
+      // as the first character on a new line.
+      if (firstChar === "[" || firstChar === "." || firstChar === "*") {
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  return false;
+}
+
+
 // Bit masks representing the state of the serializer.
 export const HAS_ENTRIES = 1;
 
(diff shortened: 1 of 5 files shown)


fluent-syntax/src/serializer.ts
```diff
@@ -136,19 +136,29 @@ function serializeTerm(term: AST.Term): string {
...
   const content = pattern.elements.map(serializeElement).join("");
-  const startOnNewLine =
+  const maybeStartOnNewLine =
     pattern.elements.some(isSelectExpr) ||
     pattern.elements.some(includesNewLine);
 
-  if (startOnNewLine) {
-    return `\n    ${indent(content)}`;
+  if (maybeStartOnNewLine) {
+    const firstElement = pattern.elements[0];
```
Pike, line 151: I'd prefer this code to be a function like the other two.

fluent-syntax/src/serializer.ts
```diff
@@ -136,19 +136,29 @@ function serializeTerm(term: AST.Term): string {
...
-  const value = indent(serializePattern(attribute.value));
+  const value = indentExceptFirstLine(serializePattern(attribute.value));
   return `\n    .${attribute.id.name} =${value}`;
 }
 
 
 function serializePattern(pattern: AST.Pattern): string {
   const content = pattern.elements.map(serializeElement).join("");
-  const startOnNewLine =
+  const maybeStartOnNewLine =
```
Pike, line 146: I was puzzled, and then I realized that this var is not named right. I took this name face-value, and was very confused that you need to do two different checks, aka, this fix is not the fix I had in mind ;-)

How about `needsIndent` ?

fluent-syntax/src/serializer.ts
stasm: I used to have it in a separate function which returned a boolean, but I then inlined it here because the `false` case resulted in two outcomes: `indentExceptFirstLine(content)` in line 157 and `content` in 164, and I didn't want to add any special sentinel return values.

However, when `content` is on a single line, `indentExceptFirstLine(content)` is a no-op, so I can also use it in line 164.

fluent-syntax/src/serializer.ts
stasm: I moved this to a function called `shouldStartOnNewLine`, which I think is the right name this time.

PR #514, opened 2020-08-17
[@fluent/bundle] Export missing FluentNone

danilobuerger:


stasm (approved):
Good catch, thanks for the PR!

Issue #519, opened 2020-12-07
In fluent-react, when an id is not found, it should report an error


gregtatum:
https://github.com/projectfluent/fluent.js/blob/72f59706128e648a4aaff5ca294bdc39a7d366fc/fluent-react/src/localized.ts#L63-L68

I'm currently using fluent-react and my UI was not translating correctly with no errors reported. The issue turned out to be missing `id`s in my bundle.

For example:

<img width="736" alt="image" src="https://user-images.githubusercontent.com/1588648/101388802-10967100-3886-11eb-8a86-2a790c88708b.png">

gregtatum:
I'll work on this.

julienw:
May I suggest to render the l10n-id when the child is null (in addition to report an error to the console maybe)? I believe this is what the "main" fluent library is doing, it would be nice that the react binding follows the same behavior.

PR #520, opened 2020-12-10
Explicitly test reporting errors in fluent-react

gregtatum:
In preparation for #519, this PR adds explicit testing to the use of errors in fluent-react. I split this out into a separate PR, as I had to do a few different steps to get the proper testing in place to where I would feel comfortable adding the new feature. Each commit should be reviewed separately, and has a full commit message describing the decisions and process I went through.

There is no behavior change for the project, these changes are test only.

### DIFF
--- a/fluent-react/package.json
+++ b/fluent-react/package.json
@@ -44,6 +44,12 @@
   "engines": {
     "node": ">=10.0.0"
   },
+  "jest": {
+    "setupFilesAfterEnv": [
+      "./test/setup.js"
+    ],
+    "restoreMocks": true
+  },
   "dependencies": {
     "@fluent/sequence": "0.5.0",
     "cached-iterable": "^0.2.1",
--- a/fluent-react/test/localized_overlay.test.js
+++ b/fluent-react/test/localized_overlay.test.js
@@ -2,16 +2,21 @@ import React from "react";
 import TestRenderer from "react-test-renderer";
 import { FluentBundle, FluentResource } from "@fluent/bundle";
 import { createParseMarkup } from "../esm/markup";
-import { ReactLocalization, LocalizationProvider, Localized }
-  from "../esm/index";
+import {
+  ReactLocalization,
+  LocalizationProvider,
+  Localized
+} from "../esm/index";
 
 describe("Localized - overlay", () => {
   test("< in text", () => {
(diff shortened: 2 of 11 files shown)


gregtatum:
Looks like CI is not passing, I will investigate.

fluent-react/test/localized_overlay.test.js
```diff
@@ -75,67 +84,79 @@ two = First &middot; Second
...
+          id="foo"
+          elems={{ button: <button onClick={alert}></button> }}
+        >
           <div />
         </Localized>
       </LocalizationProvider>
     );
 
     expect(renderer.toJSON()).toMatchInlineSnapshot(`
-                  <div>
```
gregtatum, line 91: I'm a bit surprised by these whitespace changes to be honest.

fluent-react/package.json
```diff
@@ -44,6 +44,12 @@
   "engines": {
     "node": ">=10.0.0"
   },
+  "jest": {
+    "setupFilesAfterEnv": [
+      "./test/setup.js"
+    ],
+    "restoreMocks": true
```
gregtatum, line 51: This removes the mock after every run, which helps with reproducibility of tests.

fluent-react/test/localized_valid.test.js
```diff
@@ -7,16 +7,6 @@ import {
 } from "../esm/index";
 
 describe("Localized - validation", () => {
-  let consoleError = console.error;
```
gregtatum, line 6: I made this more targeted.

stasm (approved):
LGTM. Thanks for separating the Prettier-related changes to their own commit. Out of curiosity, have you considered spying on `ReactLocalization.reportError` instead?

gregtatum:
> ReactLocalization.reportError

I believe I went for `console.warn` rather than `ReactLocalization.reportError` as I tend to prefer to test again user-facing APIs. In the Firefox Profiler UI this has been our common practice, to test the behavior as a user would actually do things. In this mode of thought, the `console.warn` is more of the endpoint that the user will actually see. Plus there is the added benefit of being in control (at least through the tests) of when information is being logged in a user-visible space. These calls are then explicitly tested so it's easy to understand the behavior with different permutations.

PR #521, opened 2020-12-13
Nested children support in overlays 

joepie91:
This change makes it possible for translations to contain nested markup tags, like so:

```ftl
foo = This is some <em><strong>nested markup!</strong></em>
```

It also fixes the test which explicitly checks that this *doesn't* work, because that test was broken and incorrectly still passed after making this change. ¯\\\_(ツ)\_/¯ That's a separate commit though, so it can be picked out if desired.

Current state of tests: all tests pass, except the one which explicitly disallows this behaviour (obviously), as I'm not sure whether there's a specific reason why nested children have not been allowed up til now. I have not added new tests for this functionality, as I find it very frustrating to deal with TS and again, I have no idea whether this PR will be accepted or not :)

So yeah, this is a bit of a draft PR, I suppose. I need this in my own application (for basic Markdown support in translation strings), hence having created a fork, but I don't know whether you're also interested in it as the upstream.

### DIFF
--- a/fluent-react/src/localized.ts
+++ b/fluent-react/src/localized.ts
@@ -143,46 +143,66 @@ export function Localized(props: LocalizedProps): ReactElement {
     }
   }
 
+  function transformTranslatedNodes(nodes: Node[]): ReactNode[] {
+    return nodes.map(childNode => {
+      if (childNode.nodeName === "#text") {
+        return childNode.textContent;
+      }
+
+      const childName = childNode.nodeName.toLowerCase();
+
+      // If the child is not expected just take its textContent.
+      if (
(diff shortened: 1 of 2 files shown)


zbraniecki:
Thank you for that draft!

DOM Overlays are actually quite sophisticated as a system. I wouldn't want to change one piece without considering others.

We want to think about it this year, and here are features we want to think about - https://github.com/zbraniecki/fluent-domoverlays-js/wiki/New-Features-(rev-3)

In issues in that repo you can read about prototypes and architectural considerations around each feature.

Would you be interested in participating in such architecture design round for rev3? It will take a looong time tho :)

PR #525, opened 2021-03-17
Typing fixes for Pontoon TypeScript migration, fix #524

### LINKED ISSUE #524: Improve typing support to aid Pontoon flow-to-ts migration
Some of the issues we see when trying to migrate Pontoon's frontend to typescript come from Fluent.

Let's have an issue to track that work.

I've got a first step for `@fluent/langneg`, but that might not be all.

CC  @mathjazz @adngdb

Pike:
* Stricter typing for @fluent/langneg

### DIFF
--- a/fluent-syntax/src/parser.ts
+++ b/fluent-syntax/src/parser.ts
@@ -1,6 +1,8 @@
 /*  eslint no-magic-numbers: [0]  */
 
 import * as AST from "./ast.js";
+// eslint-disable-next-line no-duplicate-imports
+import type {Resource, Entry} from "./ast.js";
 import { EOF, EOL, FluentParserStream } from "./stream.js";
 import { ParseError } from "./errors.js";
 
@@ -11,10 +13,6 @@ const trailingWSRe = /[ \t\n\r]+$/;
 type ParseFn<T> =
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   (this: FluentParser, ps: FluentParserStream, ...args: Array<any>) => T;
-
-type Comment = AST.Comment | AST.GroupComment | AST.ResourceComment
-type Entry = AST.Message | AST.Term | Comment;
-type EntryOrJunk = Entry | AST.Junk;
 
 function withSpan<T extends AST.SyntaxNode>(fn: ParseFn<T>): ParseFn<T> {
   return function (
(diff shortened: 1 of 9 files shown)


fluent-syntax/src/parser.ts
```diff
@@ -669,7 +667,9 @@ export class FluentParser {
     return selector;
   }
 
-  getInlineExpression(ps: FluentParserStream): AST.Expression | AST.Placeable {
+  getInlineExpression(
+    ps: FluentParserStream
+  ): AST.InlineExpression | AST.Placeable {
```
zbraniecki, line 672: `suggestion (if-minor)`: this could be a bit cleaner if we separated `getPlaceable` from `getInlineExpression`.

zbraniecki (approved):
lgtm. one suggestion, optional, if-minor.

Pike:
I've also tested this with temporarily installing my branch on the corresponding Pontoon PR, https://github.com/mozilla/pontoon/pull/1902. When doing the TS conversion there, I end up with just two remaining issues, where the Transformer claims to not have the `visit` and `genericVisit` methods. Given they're there, I don't think that should block.

Which is a long winding way of saying that this doesn't just look nice in theory, but also stood the test of a TypeScript consumer.

fluent-syntax/src/parser.ts
Pike: I gave this a look, and my intuition isn't strong here. And I don't think our tests are exhaustive. Like, I was rather surprised that 

```
shared-photos = {NUMBER({1 ->
  *[1] 2
})}
```

works. Which made me wonder if it's not about expressions when it comes down to block vs inline, but placeable.

I've paged most of that context out of my brain, so it's not minor to me.

Playground also seems to behave differently to my local repo, and I'll call shotgun on that one. It still uses the `0.14` compat build.

zbraniecki: ouch. That's not spec compliant, right?

Spec doesn't allow for it - https://github.com/projectfluent/fluent/blob/master/spec/fluent.ebnf#L74
Rust doesn't allow for it - https://github.com/projectfluent/fluent-rs/blob/master/fluent-syntax/src/parser/expression.rs#L176

maybe it's time to consider Rust->JS export for the parser?

Pike: I think the spec allows it, https://github.com/projectfluent/fluent/blob/master/spec/fluent.ebnf#L59 has the `inline_placeable` as inline expression. And I tested the fluent-rs parser, too, and it was fine.

```
jazz = {{{1->
  *[1] 2
  }}->
  *[1] 3
}
```

OTH is illegal. Works on playground as in not only in the resolver, but also shows an AST.

zbraniecki: ah, right. The `placeholder` loophole in inline :) retracting my previous statement.

PR #527, opened 2021-03-28
Separate FluentNumber/Datetime opts when used in DATETIME/NUMBER, fix #526

### LINKED ISSUE #526: Don't pass Intl.DatetimeFormatOptions to NUMBER and vice versa
I tracked the current behavior down to #410, where we pass the options tied to a `FluentNumber` to `DATETIME`. and those tied to a `FluentDateTime` to `NUMBER`

That doesn't sound like a good idea, and TypeScript started to tell me while trying to update.

Pike:
The FluentType opts are not compatible between dates and numbers, don't pass one to the formatter of the other.

As I mentioned in #526, this is compatible with what gecko does for fluent-rs, AFAICT.

### DIFF
--- a/fluent-bundle/src/builtins.ts
+++ b/fluent-bundle/src/builtins.ts
@@ -80,9 +80,15 @@ export function NUMBER(
     return new FluentNone(`NUMBER(${arg.valueOf()})`);
   }
 
-  if (arg instanceof FluentNumber || arg instanceof FluentDateTime) {
+  if (arg instanceof FluentNumber) {
     return new FluentNumber(arg.valueOf(), {
       ...arg.opts,
+      ...values(opts, NUMBER_ALLOWED)
+    });
+  }
+
+  if (arg instanceof FluentDateTime) {
+    return new FluentNumber(arg.valueOf(), {
       ...values(opts, NUMBER_ALLOWED)
     });
   }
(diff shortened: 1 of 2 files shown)


fluent-bundle/src/builtins.ts
```diff
@@ -80,13 +80,19 @@ export function NUMBER(
...
       ...arg.opts,
       ...values(opts, NUMBER_ALLOWED)
     });
   }
 
+  if (arg instanceof FluentDateTime) {
+    return new FluentNumber(arg.valueOf(), {
+      ...values(opts, NUMBER_ALLOWED)
+    });
+  }
```
zbraniecki, line 94: `question`: Seems like Gecko doesn't handle that conversion. Do you think it's useful to add or is it something TypeScript specific?

zbraniecki (approved):
lgtm!

fluent-bundle/src/builtins.ts
Pike: TBH, I don't remember if we were intentional about this.

If you pass a datetime to `NUMBER`, you can either do an error or recover by formatting the date as number. I guess in js land, `Number(new Date())` works by design. Also compatible with what `Intl.NumberFormat` does.

Not sure if there are actually good-Fluent arguments for this, though.

PR #528, opened 2021-03-28
Update TypeScript and typedoc

Pike:
For #524, I'll want new TypeScript features, so updating TypeScript.
The new TypeScript requires a new typedoc. Which removed the setup we used to generated the docs (--mode=file) because it was duh.
Now the docs are actually for the exposed symbols of the module.

I've staged the generated documentation for this on https://pike.github.io/fluent.js/. The index.html is fake. The actual changes are in https://github.com/Pike/fluent.js/commit/1847d1b5c1a57cb5f403480f2fab6f75e31cb5e7, but that's big.

I reviewed each module, and I think the changes are good. The biggest difference was in `langneg`, where `filterMatches` dropped from the docs. I wasn't comfortable about that, so I added it to the module exports.

TechDebt is beautiful.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -22,13 +22,11 @@ ROLLUP_CMD = $(ROLLUP) $(CURDIR)/esm/index.js \
 	--output.file \
 	$(NULL)
 
-TYPEDOC_CMD = $(TYPEDOC) src \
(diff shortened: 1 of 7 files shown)


zbraniecki (approved):
lgtm

PR #541, opened 2021-04-06
[@fluent/react] Allow the `Localized` component to have an array with one element in its children property

### LINKED ISSUE #539: `Localized` component throws an error when the `children` prop is an array with only one element
This is an error we are encountering while we are working on the i18n of the Firefox Profiler front-end (profiler.firefox.com).

We are aware that you should only have one child inside the `Localized` component which is reasonable. And that was working well for us. But recently, we've hit a problem related to this. Some third party components are processing the props of children components, and for our case, it's always changing the single child into an array with only one element that contains this child. This third part library is `react-contextmenu` in our case.

For example, this code won't work because of that problem:
```jsx
<ContextMenu>
  <MenuItem>
    <Localized id="test">
      <span></span>
    </Localized>
  </MenuItem>
</ContextMenu>
```
Even though there is only one span inside the `Localized`, `ContextMenu` changes the `Localized`'s children into an array that contains only the span.

This example triggers this error:
https://github.com/projectfluent/fluent.js/blob/bc796f8b33cd8f0f1f592e0e02d1a34cc91eb68d/fluent-react/src/localized.ts#L52-L56

In here, we are only checking if it's an array, but it can be also an array that contains only one element like our case.
So for the proposed solution, we can change this code to check if it contains only one element, and if so, we should continue with that element instead of throwing an error.

I can also help on this by sending a PR with my proposed solution, what do you think?

canova:
Fixes #539.

I've created the mentioned issue to explain the edge case we are having in the Firefox Profiler code base. I thought it would be better to talk about my proposed solution as well. So, I created this PR. Please let me know what you think!

### DIFF
--- a/fluent-react/src/localized.ts
+++ b/fluent-react/src/localized.ts
@@ -19,7 +19,7 @@ const reMarkup = /<|&#?\w+;/;
 export interface LocalizedProps {
   id: string;
   attrs?: Record<string, boolean>;
-  children?: ReactNode;
+  children?: ReactNode | Array<ReactNode>;
   vars?: Record<string, FluentVariable>;
   elems?: Record<string, ReactElement>;
 }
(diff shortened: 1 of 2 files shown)


gregtatum (approved):
Thanks, this looks good to me.

PR #545, opened 2021-04-09
Remove circular dependency in fluent-langneg

### LINKED ISSUE #459: langneg has a circular dependency
Hello, there is a circular dependency between locale.js and subtags.js.
`locale.js` calls [`subtags.js`](https://github.com/projectfluent/fluent.js/blob/d986155507fcb6c160695cfbfc8d368565313b51/fluent-langneg/src/locale.js#L3) which calls back [`locale.js`](https://github.com/projectfluent/fluent.js/blob/d986155507fcb6c160695cfbfc8d368565313b51/fluent-langneg/src/subtags.js#L1)

pago:
fixes #459

I've stumbled upon the warning presented in #459 today in a new svelte-kit project (which uses Vite) I am currently working on.
In that project, however, the problem wasn't just the warning about a circular dependency. Rather, code execution generally failed as `Locale` was undefined.

As mentioned by @stasm the problem is easy enough to fix by moving the code into the `locale.ts` file.
I decided to just keep `getLikelySubtagsMin` as a function, rather than making it a method on the `Locale` class to not cause any unnecessary API changes.

Please let me know if you'd prefer a different solution.


Pike:
I'm a bit concerned about pushing the full dataset into `locale.ts`.

@zbraniecki, do you have an updated perspective on the likelySubtags dataset? I  read #5, and I didn't find a lot about how we chose the dataset we currently ship. Maybe also the concerns we had then aren't as much of a problem today? Or maybe we can try to load the cldr-json dataset, and fall back to an internal one or so?

pago:
You didn't ask for my opinion but I would like to share it:
Including a full dataset for this purpose would make it very hard to use the package on the (web) client side. Going for a fallback based solution where its possible to optionally use the full data set from some source would be fine but is likely a bigger change and probably deserves some thinking on how to structure the API.

If possible, I would appreciate if we could look at this PR for what it is: A rather small refactoring that fixes a problem.
If, at some point, you'd like to rethink #5 in a bigger scope and make larger changes to the package, then that would also be perfectly fine and should not be negatively impacted by this change. But in the meantime my problem would be fixed. 😉 

Do you think there's any harm from this change? I could cut things differently to keep the JSON data separate from the `getLikleySubtagsMin` function if that would make you feel more comfortable.

Pike:
Deferring to Zibi on this one.

zbraniecki:
Handing over to Greg

gregtatum (approved):
This seems like a reasonable enough change to me.

PR #549, opened 2021-04-12
Provided better errors when components are mis-configured

### LINKED ISSUE #519: In fluent-react, when an id is not found, it should report an error

gregtatum:
Each commit here is well-ordered and can be considered separately.

Resolves #519.

The second commit isn't particularly risky, but the first introduces a new `throw` for misconfigured systems. See the commit message for more information.

dminor: As we discussed in Slack it would be great if you could look over the code. It's using TypeScript.

zibi: I also added you on here as a reviewer, as Dan wanted someone with JS experience to also look over it. Feel free to redirect if you know someone better to look at it.

### DIFF
--- a/fluent-react/src/context.ts
+++ b/fluent-react/src/context.ts
@@ -1,4 +1,5 @@
 import { createContext } from "react";
 import { ReactLocalization } from "./localization";
 
-export let FluentContext = createContext(new ReactLocalization([], null));
+export let FluentContext =
+  createContext(null) as React.Context<ReactLocalization | null>;
--- a/fluent-react/src/localized.ts
+++ b/fluent-react/src/localized.ts
@@ -66,8 +66,9 @@ export function Localized(props: LocalizedProps): ReactElement {
   }
 
   if (!l10n) {
-    // Use the wrapped component as fallback.
-    return createElement(Fragment, null, child);
+    throw new Error(
+      "The <Localized /> component was not properly wrapped in a <LocalizationProvider />."
+    );
   }
 
   const bundle = l10n.getBundle(id);
--- a/fluent-react/src/localized.ts
+++ b/fluent-react/src/localized.ts
@@ -74,6 +74,17 @@ export function Localized(props: LocalizedProps): ReactElement {
   const bundle = l10n.getBundle(id);
 
   if (bundle === null) {
+    if (id === undefined) {
+      l10n.reportError(
+        new Error("No id was provided for a <Localized /> component.")
+      );
+    } else {
+      if (l10n.areBundlesEmpty()) {
+        l10n.reportError(new Error(`A <Localized /> component was rendered when no localization bundles are present.`));
+      } else {
+        l10n.reportError(new Error(`The id "${id}" did not match any messages in the localization bundles.`));
+      }
+    }
     // Use the wrapped component as fallback.
     return createElement(Fragment, null, child);
   }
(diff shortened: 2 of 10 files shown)


fluent-react/src/context.ts
```diff
@@ -1,4 +1,5 @@
 import { createContext } from "react";
 import { ReactLocalization } from "./localization";
 
-export let FluentContext = createContext(new ReactLocalization([], null));
+export let FluentContext =
+  createContext(null) as React.Context<ReactLocalization | null>;
```
gregtatum, line 5: The addition of `ReactLocalization | null` means that the rest of the code will now need null checks.

fluent-react/src/localized.ts
```diff
@@ -66,8 +66,9 @@ export function Localized(props: LocalizedProps): ReactElement {
   }
 
   if (!l10n) {
```
gregtatum, line 68: This branch previously would never be hit, as `l10n` always existed. The change to `fluent-react/src/context.ts` means that this branch can now be exercised.

fluent-react/test/with_localization.test.js
```diff
@@ -100,7 +107,7 @@ bar = BAR {$arg}
     `);
   });
 
-  test("getString without access to the l10n context", () => {
+  test.skip("getString without access to the l10n context", () => {
```
gregtatum: These get re-enabled in the next commit.

fluent-react/test/localized_fallback.test.js
```diff
@@ -61,6 +61,8 @@ foo = FOO
 });
 
 test("falls back back for missing message", function() {
+  jest.spyOn(console, "warn").mockImplementation(() => {});
```
gregtatum, line 64: This activates Jests's "spy" feature, which allows the test to see if the function was called. In general I prefer this approach as it tests the "user interface" for the code.

fluent-react/test/localized_valid.test.js
```diff
@@ -64,17 +79,23 @@ describe("Localized - validation", () => {
     expect(console.error).toHaveBeenCalled();
   });
 
-  test("with single children inside an array", () => {
+  test("is valid to have a children array with a single element inside", () => {
```
gregtatum, line 82: I updated the examples to include valid configurations. Before the configurations would "just work" even when misconfigured.

fluent-react/test/with_localization.test.js
```diff
@@ -100,31 +100,60 @@ bar = BAR {$arg}
...
+        ],
         Array [
           "[@fluent/react] ReferenceError: Unknown variable: $arg",
         ],
       ]
     `);
   });
 
-  test.skip("getString without access to the l10n context", () => {
+  test("getString with an empty bundle list", () => {
```
gregtatum, line 113: It's still valid to provide an empty bundle list. To me this makes sense if for some reason you wanted to render you page with fallbacks. The LocalizationProvider still needs to be in place, but it will be configured with an empty list.

dminor (approved):
Praise: I think this will make debugging failures much easier! Thank you for adding the additional test cases and improving the names of the existing ones.

I don't see any issues here, beyond what the linter flagged. LGTM once those are fixed.

zbraniecki (approved):

Issue #482, opened 2020-05-05
Support Node 14's native ES module


stasm:
Currently, `@fluent` packages are published as `type: "commonjs"`. This means that in Node 14, you can only import then as if they were CommonJS modules:

- From other cjs files:
```js
let {FluentBundle} = require("@fluent/bundle");
let bundle = new FluentBundle();

// Or:
let fluent_bundle = require("@fluent/bundle");
let bundle = new fluent_bundle.FluentBundle();
```

- From mjs files:
```js
import fluent_bundle from "@fluent/bundle";
let bundle = new fluent_bundle.FluentBundle();
```

It would be nice to be able to use the following import form, too, just like we can do today in browsers:
```js
import {FluentBundle} from "@fluent/bundle";
let bundle = new FluentBundle();
```

stasm:
https://nodejs.org/api/esm.html#esm_dual_commonjs_es_module_packages has a good discussion of the available approaches. I think the most promising would be the `wrapper.mjs` approach.

ranisalt:
I think the packages export all that's needed for both module and commonjs imports, so omitting `type` should suffice

tobias47n9e:
Seems like **Angular 12** is now also having troubles compiling `@fluent`

```
    ./node_modules/@fluent/bundle/esm/index.js:9:0 - Error: Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (9:0)
    File was processed with these loaders:
     * ./node_modules/@angular-devkit/build-angular/src/babel/webpack-loader.js
     * ./node_modules/@angular-devkit/build-optimizer/src/build-optimizer/webpack-loader.js
     * ./node_modules/@ngtools/webpack/src/ivy/index.js
    You may need an additional loader to handle the result of these loaders.
    |  *
    |  */
    > export { FluentBundle } from "./bundle.js";
    | export { FluentResource } from "./resource.js";
    | export { FluentType, FluentNumber, FluentDateTime } from "./types.js";
```

zbraniecki:
hi all! We are a bit low on resources temporarily (most of the focus is on fluent-rs and MessageFormat 2.0 work), so I can't promise when I'll get to this bug.

If anyone can provide a PR, I'll be happy to review and merge it

Demivan:
Webpack 5 fix: https://github.com/projectfluent/fluent.js/pull/556

PR #556, opened 2021-05-28
Remove "type": "commonjs" from package.json

### LINKED ISSUE #517: Not able to compile with webpack 5
I've updated webpack to version 5-rc1 and since then I'm getting an error with fluent packages:

```
  ERROR in ./_canny/node_modules/@fluent/bundle/esm/index.js 9:0
  Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (9:0)
  You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
  |  *
  |  */
  > export { FluentBundle } from "./bundle.js";
  | export { FluentResource } from "./resource.js";
  | export { FluentType, FluentNumber, FluentDateTime } from "./types.js";
```
and
```  ERROR in ./_canny/node_modules/@fluent/langneg/esm/index.js 9:0
  Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (9:0)
  You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
  |  *
  |  */
  > export { negotiateLanguages } from "./negotiate_languages";
  | export { acceptedLanguages } from "./accepted_languages";
  |
```

I now found out, that last version of webpack, which was able to to compile it was 5.0.0-beta.29

Demivan:
Setting "type" field in package.json forces Webpack to treat all files in the package as either CommonJS or EcmaScript. 
Without it Webpack correctly uses "main" and "module" fields.
Fixes #517

### DIFF
--- a/fluent-bundle/package.json
+++ b/fluent-bundle/package.json
@@ -15,7 +15,6 @@
       "email": "stas@mozilla.com"
     }
   ],
-  "type": "commonjs",
   "main": "./index.js",
   "module": "./esm/index.js",
   "types": "./esm/index.d.ts",
--- a/fluent-dedent/package.json
+++ b/fluent-dedent/package.json
@@ -11,7 +11,6 @@
       "email": "stas@mozilla.com"
     }
   ],
-  "type": "commonjs",
   "main": "./index.js",
   "module": "./esm/index.js",
   "types": "./esm/index.d.ts",
--- a/fluent-dom/package.json
+++ b/fluent-dom/package.json
@@ -18,7 +18,6 @@
       "email": "stas@mozilla.com"
     }
   ],
-  "type": "commonjs",
   "main": "./index.js",
   "module": "./esm/index.js",
   "types": "./esm/index.d.ts",
--- a/fluent-langneg/package.json
+++ b/fluent-langneg/package.json
@@ -15,7 +15,6 @@
       "email": "stas@mozilla.com"
     }
   ],
-  "type": "commonjs",
   "main": "./index.js",
   "module": "./esm/index.js",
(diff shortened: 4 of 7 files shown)


Demivan:
Looks like this fixes #482 too. I was able to import @fluent/bundle both from .cjs and .mjs file.

Pike:
Brain dump for Greg:

I recall that the current setup was a compromise between supporting some set of nodejs versions with different capabilities.

We still have node 10 in our testing matrix, which we could drop now. We also don't seem to have any sorts of integration tests with common package managers or bundlers. That part is pretty manual testing, IIRC.

Demivan:
Just checked with bundlers: Webpack 5, parcel, rollup, vite - working fine. I'm using yarn as package manager - that is working too.
Node 10, 12, 14 all work.
Node 10 only works with old syntax like described in #482. But changing "type": "commonjs" does not affect this. So I'm not sure what "type": "commonjs" fixes.
Both 12 and 14 can use native ES imports without any issue, so looks like #482 is not relevant now.

ranisalt:
Node 10 has passed it's end-of-life, I don't think someone comfortable with using that old of a Node version would mind using an equally outdated version of a dependency

gregtatum:
image

Yeah, it seems like we should only support as far back as the LTS release.

gregtatum (commented):
Hah, forgot to hit approve on it before merging. Thanks for the fix!

Siilwyn:
:wave: would it be possible to publish this in a package update? This is blocking usage for me atm.

julienw:
Thanks so much for this!

PR #565, opened 2021-07-20
Refresh dependencies

eemeli:
This PR refreshes the lockfiles to use the latest versions supported by the ranges defined in `package.json` files, and updates internal dependencies to the latest, where those were using previous versions.

This should satisfy all of the dependabot bump PRs as well.

### DIFF
--- a/fluent-bundle/package-lock.json
+++ b/fluent-bundle/package-lock.json
@@ -1,8 +1,214 @@
 {
   "name": "@fluent/bundle",
   "version": "0.16.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@fluent/bundle",
+      "version": "0.16.1",
+      "license": "Apache-2.0",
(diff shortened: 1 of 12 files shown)


gregtatum (approved):
Since this my first review of your code, I like to point out that I use [conventional comments](https://conventionalcomments.org/) for my reviews. The Mozilla intl team so far has agreed that this leads to higher quality reviews that have less ambiguity, and we've adopted as a convention for our team.

Everything looks good on my end. I mostly looked at `package.json` updates, as the `package-lock.json` files are machine created. The CI is passing, so this looks good to me.

praise: Thanks for creating nice clean commits, it made the review very simple and non-controversial.

praise: Thanks for doing some of the chore work, the package-lock.json format version was annoying me, as I think I did it wrong when I was in here last.

Feel free to land code yourself when you have an r+ (approved PR) and a clean CI. It would be good to follow the same conventions of previous mergers on naming of the merge commit. I can't tell with a quick look if fluent.js is using just merges, or squash and merge.

PR #566, opened 2021-07-21
Refresh dependencies in fluent-react example

eemeli:
This does the same updates as in #565, but for `fluent-react/example/`, which was missed in the previous.

As this isn't covered by the test suite, I verified the changes manually by following the example's README instructions.


gregtatum (approved):
Thanks! Looks good to me.

PR #557, opened 2021-05-28
Drop Node 10.x and add 16.x to CI

ranisalt:
Follow up of https://github.com/projectfluent/fluent.js/pull/556#issuecomment-850264456

PR #567, opened 2021-07-21
Set type: "module" for esm/ and test/ in all packages & update to mocha@9

eemeli:
While the top-level `package.json` files include a `"type": "commonjs"` specification, the `esm/` directories can (and ought) to either have their contents use the `.mjs` extension, or include their own `package.json` file with the correct `"type"`.

This adds the `package.json` files while keeping the extensions as `.js`, as this is significantly easier for the TypeScript interop; proper `.mjs` support is currently an [unresolved issue](https://github.com/microsoft/TypeScript/issues/18442).

~For context, I noticed/encountered this issue while updating Mocha to its latest version; I'll file another PR on that separately if/once this is merged.~

Edit: Mocha update is now included in this PR, as the old version wasn't able to handle the `"module"` types.


eemeli:
Bother, it looks like the Mocha update needs to be done simultaneously. I'll include that here & rephrase the description.

gregtatum (changes requested):
Setting as "Request changes" until the mocha issue is resolved. Please re-request review from me when you are ready.

eemeli:
Tooling updates are fun. Also needed to patch the ESLint configs, which to be clear are explicitly using `babel-eslint` rather than `@babel/eslint-parser`, as the latter requires a newer ESLint version.

Each of the commits should explain themselves. I'll submit the actual ESLint update separately, as that looks like it'll require more changes to account for changed recommendations.

fluent-syntax/src/index.ts
```diff
@@ -2,11 +2,11 @@ import { Resource } from "./ast.js";
...
-export * from "./ast";
-export * from "./errors";
-export * from "./parser";
-export * from "./serializer";
-export * from "./visitor";
+export * from "./ast.js";
+export * from "./errors.js";
+export * from "./parser.js";
+export * from "./serializer.js";
+export * from "./visitor.js";
```
Demivan, line 9: Why does it import `.js` files from `.ts` file?

eemeli: Because this is the cleanest way of producing valid ES module code in the `esm/` directories. Node.js resolution requires the `.js` extension for `import` calls, and TypeScript automatically resolves e.g. `./ast.js` as `./ast.ts`.

Without these, we'd need a separate build step after the `tsc` call that would modify all of the relative paths to use fully qualified extensions. It'd be nice if TypeScript supported that natively, but, well, it doesn't. This is also why we need the `esm/package.json` files, as TypeScript _only_ handles the `.js` extension like this, and not e.g. `.mjs`.

fluent-bundle/test/arguments_test.js
```diff
@@ -3,9 +3,9 @@
 import assert from 'assert';
 import ftl from "@fluent/dedent";
 
-import {FluentBundle} from '../esm/bundle';
-import {FluentResource} from '../esm/resource';
-import {FluentType, FluentNumber, FluentDateTime} from '../esm/types';
+import {FluentBundle} from '../esm/bundle.js';
```
gregtatum, line 6: Question: If someone regresses this by changing it back to `'../esm/bundle'` will our CI catch it?

package-lock.json
```diff
@@ -18,15 +18,15 @@
...
+        "mocha": "^9.0.2",
         "nyc": "^15.1.0",
         "prettyjson": "^1.2.1",
         "rollup": "^1.9.3",
         "typedoc": "^0.20.34",
         "typescript": "^4.2.3"
       },
       "engines": {
-        "node": ">=10.0.0"
+        "node": ">=12.0.0"
```
gregtatum, line 30: Thought: I double checked that 12 is the current LTS, so this change looks good to me.

eslint_test.json
```diff
@@ -1,4 +1,5 @@
 {
+    "parser": "babel-eslint",
```
gregtatum, line 2: Suggestion (optional): It would be nice to document why this parser is needed in case it can be removed, however this is a JSON configuration file which doesn't support comments. I'm not sure if it's feasible to convert it to a `.js` configuration file, so I'm marking this suggestion as optional.

gregtatum (approved):
The changes all look reasonable to me, and CI is passing, so I'm happy to give my approval. Thanks for the maintenance work! I left one optional suggestion, and a few minor comments.

eslint_test.json
gregtatum: Ah, I believe this comment is outdated by your next PR which converts it to `.yml`.

fluent-bundle/test/arguments_test.js
eemeli: Yes, the Mocha tests will fail then. The exact error they fail with is a bit misleading though:

```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/eemeli/code/fluent-js/fluent-bundle/test/arguments_test.js
require() of ES modules is not supported.
require() of /Users/eemeli/code/fluent-js/fluent-bundle/test/arguments_test.js from /Users/eemeli/code/fluent-js/node_modules/mocha/lib/esm-utils.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename arguments_test.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/eemeli/code/fluent-js/fluent-bundle/test/package.json.
```

There's a `try { ... } catch` block around the `import()` [here](https://github.com/mochajs/mocha/blob/master/lib/esm-utils.js#L47) which falls back to `require()`, and logs _its_ error if both fail. I may end up filing an upstream bug about this.

package-lock.json
eemeli: Yeah, this is a result of the changes in #557 which weren't completely reflected in the lockfiles.

eslint_test.json
eemeli: This dependency gets updated to `@babel/eslint-parser` by #568. That PR also ports the ESLint configs to YAML in order to allow them to be later commented.

PR #568, opened 2021-07-22
Update ESLint and simplify its configs

eemeli:
This builds on top of #567, so only the last five commits are actually relevant here.

The starting point for this changeset is that ESLint was behind by a major version, and that needed to be updated. Some of its associated plugins & such were behind by a couple of major versions. Now they aren't anymore.

As I was working with the configs (due to updated rules, which needed to be addressed), I realised that there was scope to simplify/centralise the config files to follow ESLint and TypeScript defaults, which allows for e.g. IDEs to pick up on them much more easily. This way, e.g. VS Code's ESLint plugin can automatically highlight issues without any workspace-specific config.

While fixing the lint complaints, I think I got rid of almost all `any` types in the project. There's a detail or two on which I'll add line comments.


fluent-bundle/src/scope.ts
```diff
@@ -34,10 +34,27 @@ export class Scope {
...
+  ): Intl.PluralRules;
+  memoizeIntlObject(
+    ctor:
+      | typeof Intl.NumberFormat
+      | typeof Intl.DateTimeFormat
+      | typeof Intl.PluralRules,
+    opts: Intl.NumberFormatOptions &
+      Intl.DateTimeFormatOptions &
+      Intl.PluralRulesOptions
+  ): Intl.NumberFormat | Intl.DateTimeFormat | Intl.PluralRules {
```
eemeli, line 57: While this function used to be parametric, it should be easier to use with the overloads that allow the return type to be correctly typed.

fluent-syntax/src/ast.ts
```diff
@@ -6,6 +6,8 @@
  *
  */
 export abstract class BaseNode {
+  abstract type: string
```
eemeli, line 9: The immediate reason for adding this is to allow for `node.type` to be known to be a `string` in template expressions. While this typing is nominally new, it matches all existing practice at least within thie monorepo.

I'm not sure if this should techinically count as a breaking change, but as long as the next version of `@fluent/syntax` is published as 0.18.0, all should be fine even from a semver point of view.

eslint_js.yaml
```diff
@@ -0,0 +1,79 @@
+parserOptions:
+  ecmaVersion: 9
+  sourceType: module
+
+env:
+  es6: true
+
+rules:
+  valid-typeof: 2
```
gregtatum: Question: Are any of the rules changed here? Or are they all the same? I'm not manually reviewing that they are the same, but I just wanted to make sure I can assume they are.

eslint_js.yaml
```diff
@@ -0,0 +1,79 @@
+parserOptions:
```
gregtatum: praise: Ah, yaml is much more reasonable of a format to use here. I didn't know eslint could do this.

eslint_test.yaml
```diff
@@ -1,9 +1,6 @@
-parser: babel-eslint
+parser: "@babel/eslint-parser"
```
gregtatum: Suggestion: Please document the parser decision here.

fluent-bundle/tsconfig.json
```diff
@@ -1,16 +1,7 @@
...
-        "esModuleInterop": true,
-        "moduleResolution": "node",
-        "noEmitHelpers": true,
-        "declaration": true,
-        "outDir": "./esm",
-    },
-    "include": [
-        "./src/**/*.ts"
-    ]
+  "extends": "../tsconfig.json",
```
gregtatum, line 2: praise: Thanks for the consolidation!

fluent-bundle/src/resolver.ts
```diff
@@ -111,7 +111,7 @@ function getArguments(
   args: Array<Expression | NamedArgument>
 ): Arguments {
   const positional: Array<FluentValue> = [];
-  const named: Record<string, FluentValue> = Object.create(null);
```
gregtatum, line 114: Question: Were these new errors?

fluent-bundle/src/scope.ts
gregtatum: Question: How does typescript refine the usage here? Will the return type always be `Intl.NumberFormat | Intl.DateTimeFormat | Intl.PluralRules`, or can it refine the union to a single `Intl.NumberFormat`? I'm a bit surprised by this change, but perhaps this was a new warning.

fluent-syntax/src/ast.ts
gregtatum: Thought: I'd feel comfortable with a minor semver bump.

gregtatum (approved):
I'm comfortable marking approved here, although I was a bit surprised by some of the TypeScript changes. I think I asked in the questions if these were purely motivated by new errors surfaced. If they were drive-by fixes I would prefer them in a separate PR.

eslint_js.yaml
eemeli: None of these rules are changed, essentially this file is just renamed as `eslint-base.yaml`.

For TS I did need to disable a couple of rules that are switched on for JS, but which don't really work with TS or are better handled by `@typescript-eslint` rules.

fluent-bundle/src/resolver.ts
eemeli: Yes. `Object.create()` returns `any`, and the `@typescript-eslint/recommended` set includes [no-unsafe-assignment](https://github.com/typescript-eslint/typescript-eslint/blob/v4.28.4/packages/eslint-plugin/docs/rules/no-unsafe-assignment.md) which complains about assigning that to a more specific target.

So by moving the typing to an `as`, we do the same thing as before but ESLint stops complaining.

fluent-bundle/src/scope.ts
eemeli: The immediately preceding [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) allow for the return type to match the argument types.

And yes, there was a new linter error about the preceding `object` usage.

fluent-syntax/src/ast.ts
eemeli: Heh, we're still in 0.x land so a minor and a major bump are effectively the same thing. :)

eslint_test.yaml
eemeli: I sort of fixed this, and then almost immediately realised I could & should fix it better. The `@babel/eslint-parser` dependency is now only in `fluent-react/`, as it's really only required for the JSX content in its test files.

For the `import.meta` use in other test files, it's sufficient to use `ecmaVersion: 2020` as a parser option.

PR #569, opened 2021-07-28
Add npm workspaces & tool-specific config files, drop make

eemeli:
Apologies for the size of this PR, but the whole changeset seemed best kept together. This transitions the repo to use and depend on [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) for package management. This is what that means for each of the previous `make` commands, as run from the repo root:

### `make deps` and `make depsclean`

Completely dropped. Running `npm install` or `npm ci` at the root now installs all dependencies of all packages. Per-package lockfiles are also all dropped.

### `make lint` → `npm run lint`

Run now only from the repo root, linting all packages at once. Due to updated ESLint rules this will now complain if the packages have not been built first, as otherwise missing transitive dependencies resolve as `any`.

### `make test` → `npm test`

The common Mocha config allows testing 5/8 packages all at once (via `npm run test:common`). `@fluent/dom` gets a separate invocation due to its JSDOM setup, and `@fluent/react` on account of using Jest. `@fluent/gecko` remains without tests, but it's just repackaging other bits. Tests continue to use built rather than source assets. A `.mocharc.yaml` config file is added to the repo root.

Internally, the code coverage tool is switched from `nyc` to `c8`, as the latter works more dependably with ES modules as used by `mocha@9`. The output is the same for both, so this has no external visibility.

### `make build` → `npm run build -ws`

Note the `-ws` argument: that's short for `--workspaces`, and it runs the `build` command in all of the packages, in the order specified in the `"workspaces"` array of the root package.json file. The order matters, as npm doesn't automatically sort the packages topologically based on their internal dependencies. Rollup configs are now in `rollup.config.mjs` files in the repo root and in `fluent-gecko/`, rather than as CLI arguments inside makefiles.

### `make clean` → `npm run clean`

Run now only from the repo root, cleaning all packages at once. Uses `git clean`, as that's smart enough to only remove files that aren't stored in the repo.

### `make html` → `npm run docs -ws`

Config is moved from makefiles to `typedoc.config.cjs` in the repo root.

### `make dist` → `npm run dist`

Defined now by the `"predist"`, `"dist"` and `"postdist"` scripts in the root package.json file. Previously this looped packages/commands; now it's commands/packages.

## Other notable changes

- Each package provides its own `build`, `test` and `docs` scripts, so that they may be run individually.
[...]

PR #573, opened 2021-09-13
Improve the TypeScript lint/intellisense experience

eemeli:
This is a follow-up to #569, where some of the dev-env challenges were discussed with @gregtatum.

With these changes, the `lint` task is no longer dependent on `build` and intellisense works immediately for internal cross-package dependencies.

The dependency of `test` on `build` remains, and it's turning out to be more difficult without in-depth changes. Briefly, Mocha does not offer an out-of-the-box path aliasing as e.g. TypeScript's `paths` or Jest's `moduleNameMapper`. This means that something like a Babel plugin would need to be registered for tests, as well as `ts-node` for handling the TypeScript sources. Alternatively, all of the tests could be ported to Jest and/or TypeScript. That sounds like more work than it's worth at this time.


gregtatum (approved):
Thanks, everything seems to be working well on my end.

PR #575, opened 2021-09-14
Update dependencies

eemeli:
This updates almost all of the project's dependencies, including the following major dev dependency updates:
- `rollup` v1 → v2
- `typedoc` v0.20 → v0.22
- `typescript` v4.2 → v4.4
- `sinon` v4 → v11 (fluent-bundle)
- `jsdom` v15 → v17 (fluent-dom)
- `jest` v24 → v27 (fluent-react)
- `react` v16 → v17 (fluent-react)

Some of those require a few changes in code; these are also applied. The following externally visible changes are made:
- The updated typedoc's default theme uses dark mode if that's the user's preference; this can be toggled at the bottom of the page.
- **fluent-bundle:** the type of caught errors is validated at runtime
- **fluent-react:** `prop-types` is dropped as a dependency

The version of `commander` that's used in `tools/` is not updated here, as the distance in its updates (v2 → v8) is so great, and the tools aren't covered by any tests.

### DIFF
--- a/fluent-bundle/package.json
+++ b/fluent-bundle/package.json
@@ -50,6 +50,6 @@
   },
   "devDependencies": {
     "@fluent/dedent": "file:../fluent-dedent",
-    "sinon": "^4.2.2"
+    "sinon": "^11.1.2"
   }
 }
--- a/fluent-bundle/src/bundle.ts
+++ b/fluent-bundle/src/bundle.ts
@@ -196,7 +196,7 @@ export class FluentBundle {
       let value = resolveComplexPattern(scope, pattern);
       return value.toString(scope);
     } catch (err) {
-      if (scope.errors) {
+      if (scope.errors && err instanceof Error) {
         scope.errors.push(err);
         return new FluentNone().toString(scope);
       }
--- a/fluent-dom/test/index.js
+++ b/fluent-dom/test/index.js
@@ -13,3 +13,6 @@ Object.keys(document.defaultView).forEach(property => {
 global.navigator = {
   userAgent: 'node.js'
 };
+
+// silence console logging during tests
+console.warn = () => {}
--- a/fluent-react/package.json
+++ b/fluent-react/package.json
@@ -68,14 +68,13 @@
   "devDependencies": {
     "@babel/eslint-parser": "^7.14.7",
     "@babel/preset-env": "^7.5.5",
-    "@babel/preset-react": "7.0.0",
+    "@babel/preset-react": "^7.14.5",
     "@fluent/bundle": "file:../fluent-bundle",
-    "@types/react": "^16.9.22",
-    "@types/react-dom": "^16.9.5",
+    "@types/react": "^17.0.20",
     "babel-jest": "^27.2.0",
     "jest": "^27.2.0",
-    "react": "^16.8.0",
-    "react-dom": "^16.8.0",
-    "react-test-renderer": "^16.8.6"
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "react-test-renderer": "^17.0.2"
   }
 }
--- a/fluent-react/package.json
+++ b/fluent-react/package.json
@@ -42,7 +42,7 @@
   ],
   "scripts": {
     "build": "tsc -p ./tsconfig.build.json",
-    "postbuild": "rollup -c ../rollup.config.mjs --globals @fluent/sequence:FluentSequence,cached-iterable:CachedIterable,react:React,prop-types:PropTypes",
+    "postbuild": "rollup -c ../rollup.config.mjs --globals @fluent/sequence:FluentSequence,cached-iterable:CachedIterable,react:React",
     "docs": "typedoc --options ../typedoc.config.cjs --tsconfig ./tsconfig.build.json",
     "test": "jest --collect-coverage"
   },
@@ -59,8 +59,7 @@
   },
   "dependencies": {
     "@fluent/sequence": "^0.7.0",
-    "cached-iterable": "^0.3.0",
-    "prop-types": "^15.7.2"
+    "cached-iterable": "^0.3.0"
   },
   "peerDependencies": {
     "@fluent/bundle": ">=0.16.0 <0.18.0",
--- a/fluent-react/package.json
+++ b/fluent-react/package.json
@@ -54,7 +54,8 @@
     "setupFilesAfterEnv": [
       "./test/setup.js"
     ],
-    "restoreMocks": true
+    "restoreMocks": true,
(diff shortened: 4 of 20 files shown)


fluent-bundle/src/bundle.ts
```diff
@@ -196,7 +196,7 @@ export class FluentBundle {
       let value = resolveComplexPattern(scope, pattern);
       return value.toString(scope);
     } catch (err) {
-      if (scope.errors) {
+      if (scope.errors && err instanceof Error) {
```
gregtatum, line 199: Question: Does this risk being a behavior change? It could affect users. I don't know how errors move through the system and if we can assume that it will be properly an instance of Error. This applies to the other changes elsewhere. I'm assuming this change is addressing is a new type error.

fluent-react/package.json
```diff
@@ -72,9 +73,8 @@
     "@fluent/bundle": "file:../fluent-bundle",
     "@types/react": "^16.9.22",
     "@types/react-dom": "^16.9.5",
-    "babel-jest": "^24.8.0",
-    "jest": "^24.8.0",
-    "prettier": "^1.18.2",
```
gregtatum, line 77: Question: Why is this unneeded? Is prettier configured elsewhere?

fluent-react/src/provider.ts
```diff
@@ -29,14 +28,7 @@ export function LocalizationProvider(
...
     FluentContext.Provider,
-    {
-      value: props.l10n
-    },
+    { value: props.l10n },
     props.children
   );
 }
-
-LocalizationProvider.propTypes = {
```
gregtatum, line 39: Question: Is this a behavior change for our users? Not everyone uses TypeScript, and this could provide valuable runtime checks if users are relying on the added safety here.

fluent-react/test/provider_valid.test.js
```diff
@@ -25,18 +25,4 @@ describe("LocalizationProvider - validation", () => {
     );
     expect(renderer.toJSON()).toMatchInlineSnapshot(`<div />`);
   });
-
-  test("without a child", () => {
```
gregtatum, line 29: Suggestion (blocking): I'm nervous about this change, since _we_ can't do this with TypeScript type checks, but our users might. For instance in Firefox DevTools this is used without type checks.

fluent-dom/test/index.js
```diff
@@ -13,3 +13,6 @@ Object.keys(document.defaultView).forEach(property => {
 global.navigator = {
   userAgent: 'node.js'
 };
+
+// silence console logging during tests
+console.warn = () => {}
```
gregtatum, line 18: Suggestion: I don't think this is a good change. It would be better to add suppressions per-test. This is a global change for every test, and it could be very surprising as to why this function doesn't work. I generally treat console as a public API, and one that should be handled explicitly, and asserted in tests.

fluent-bundle/package.json
```diff
@@ -50,6 +50,6 @@
   },
   "devDependencies": {
     "@fluent/dedent": "file:../fluent-dedent",
-    "sinon": "^4.2.2"
+    "sinon": "^11.1.2"
```
gregtatum, line 53: Thought: Ah I guess this is using mocha instead of Jest. I was going to suggest that we don't need sinon, but perhaps it is needed here.

gregtatum (changes requested):
The upgrades all look reasonable to me, but I found an issue I would like to push back on before approving. I'm concerned about the removal of PropTypes from fluent-react, so I'm marking this as "Request changes".

fluent-bundle/src/bundle.ts
eemeli: I traced through all of the `throw` statements in the repo, and there's no reason to suspect that anything other than an `Error` or one of its subclasses might be thrown.

This change is due to a change of defaults in TypeScript 4.4, specifically this: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

fluent-react/package.json
eemeli: It's no longer required by Jest, which was its only user: facebook/jest#11400

It wasn't actually used to format code in `fluent-react/`.

fluent-react/src/provider.ts
eemeli: The `prop-types` methods are effectively no-ops in production, so I don't think that this should be considered a change in behaviour. It's removed here because issues like facebook/prop-types#354 make it difficult to support `react@17` while keeping it as a dependency.

fluent-bundle/package.json
eemeli: Yeah, Mocha doesn't include a spy/mock generator. Technically `sinon` isn't absolutely required, but it does make it easier to spy on `Intl.NumberFormat` calls in `fluent-bundle/test/constructor_test.js`.

fluent-react/test/provider_valid.test.js
eemeli: This PR drops two tests. They expected these to throw:

```jsx
<LocalizationProvider l10n={new ReactLocalization([])} />
```
```jsx
<LocalizationProvider />
```

I don't think there's any actual benefit from throwing for these, and doing so is in fact surprising. A `LocalizationProvider` is a [`Context.Provider`](https://reactjs.org/docs/context.html#contextprovider), which normally does not _require_ its children to be non-empty. It just does nothing in that case.

The second checks that the `l10n` prop is defined. This will continue to throw in actual use, as the users of the context get its value with the TS type `ReactLocalization | null`, and handle the latter by throwing. This is already effectively checked by tests in `fluent-react/test/localized_valid.test.js`.

@gregtatum Would you like me to add a more specific test to the `Localized` test suite covering the missing-prop case?

fluent-dom/test/index.js
eemeli: Ok, will change. I'll bump up `sinon` to the root package.json and use that in these tests as well.

fluent-react/src/provider.ts
gregtatum: I would somewhat disagree with the change in behavior, as it's still used in development mode. However, I won't block on this change if you feel the dependency management is too complex.

fluent-react/test/provider_valid.test.js
gregtatum: I think there is some value in exercising these code paths as users of this package could do surprising things with it. Without the throws, I'm less opinionated if the tests stay or if they go.

gregtatum (approved):
Marking as approved since we talked through my blocking comments. Feel free to re-request if you want me to look at the sinon changes, or feel free to merge without me looking at that.

PR #576, opened 2021-09-14
test(bundle): Use different Date constructor to get a local date

### LINKED ISSUE #572: Failing FluentDateTime test due to time zone differences of machine running test
I believe this is caused by my machine being in a different time zone.

```
  1 failing

  1) Variables
       and dates
         can be a FluentDateTime:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'Wednesday'
- 'Thursday'
      + expected - actual

      -Wednesday
      +Thursday

      at Context.<anonymous> (file:///Users/greg/dev/fluent.js/fluent-bundle/test/arguments_test.js:221:14)
      at processImmediate (node:internal/timers:464:21)
```

I would suggest rewriting the test to allow for either Thursday or Wednesday.

eemeli:
Fixes #572 by using a different form of `new Date()`. The constructor parses string input as a UTC date, while numeric input uses the local timezone.

### DIFF
--- a/fluent-bundle/test/arguments_test.js
+++ b/fluent-bundle/test/arguments_test.js
@@ -215,7 +215,8 @@ suite('Variables', function() {
     });
 
     test('can be a FluentDateTime', function(){
-      const arg = new FluentDateTime(new Date('2016-09-29'), {weekday: "long"});
+      const localDate = new Date(2016, 8, 29, 12);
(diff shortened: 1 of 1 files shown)


gregtatum (approved):
Thanks!

PR #577, opened 2021-09-21
Add ".js" extension to imports in fluent-react

indeyets:
Similar change was applied to different modules, but react was skipped somehow. Lack of this change breaks build on webpack5


eemeli (approved):
Yeah, this looks like a pretty clear case of oversight about applying that change to this file.

Issue #589, opened 2022-04-28
[fluent-syntax] Calling `clone()` on a node should return a value with node's type, not `BaseNode`


eemeli:
The correct TS type here would be `this`:

https://github.com/projectfluent/fluent.js/blob/a68c20eaea45ae8f948570e6b2ac13f2f0306b5b/fluent-syntax/src/ast.ts#L50

AleksandrSl:
The issue seems to be resolved, so it can be closed 😀

PR #590, opened 2022-04-28
Update dependencies & fix minor TS issue

eemeli:
These updates should for the most part have no external visibility, except for the #589 fix.

I was intending to also update `eslint-plugin-mocha`, but that's currently blocked by ramda/ramda#3278.

### DIFF
--- a/fluent-langneg/src/negotiate_languages.ts
+++ b/fluent-langneg/src/negotiate_languages.ts
@@ -1,4 +1,4 @@
-import {filterMatches} from "./matches.js";
+import { filterMatches } from "./matches.js";
 
 export interface NegotiateLanguagesOptions {
   strategy?: "filtering" | "matching" | "lookup";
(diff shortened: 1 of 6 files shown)


fluent-langneg/src/negotiate_languages.ts
```diff
@@ -51,22 +51,19 @@ export interface NegotiateLanguagesOptions {
...
-    defaultLocale,
-  }: NegotiateLanguagesOptions = {}
+  { strategy = "filtering", defaultLocale }: NegotiateLanguagesOptions = {}
 ): Array<string> {
-
   const supportedLocales = filterMatches(
-    Array.from(Object(requestedLocales)).map(String),
-    Array.from(Object(availableLocales)).map(String),
+    Array.from(requestedLocales ?? []).map(String),
+    Array.from(availableLocales ?? []).map(String),
```
eemeli, line 58: This has the same effect of avoiding throws for nullish inputs as before, but is clearer & type-friendlier about it.

gregtatum (approved):
Looks good to me!

PR #591, opened 2022-04-29
Add Prettier as formatter & apply its code style

eemeli:
This normalises the formatting across the repo, making it easier to work with. The Prettier config is minimally adjusted to select the style that's most commonly currently used.


gregtatum (approved):
Looks good to me! I popped off the last commit, and re-ran prettier locally. I got the same results, so the changes look good.

PR #593, opened 2022-05-03
Update fluent-react/example to use latest Parcel

eemeli:
Some minor example code updates to switch from Parcel v1 to v2.


gregtatum (approved):
The code changes look reasonable to me, I didn't test it locally. Let me know if you would like me to.

eemeli:
Thank you! No need, esp. as this is only touching example code.

PR #595, opened 2022-07-09
Add getFragment method to ReactLocalization

### LINKED ISSUE #544: `withLocalization`'s `getString` API is missing the `elems` parameter
We are usually using the `Localized` component, but recently I was trying the `withLocalization`'s `getString` API, and it looks like it is the missing the `elems` parameter, which is pretty commonly used in Firefox Profiler.

Also wiki seems to be missing more advanced usage information. It's not possible to see other parameters here: https://github.com/projectfluent/fluent.js/wiki/withLocalization

So, I tried to check the implementation of getString and realized that it doesn't have a `elems` parameter, like [Localized](https://github.com/projectfluent/fluent.js/blob/04733c099669f16245e4b6cc82ce45ad4cee8a00/fluent-react/src/localized.ts#L19-L25):
https://github.com/projectfluent/fluent.js/blob/a1ac29df2eba6b2e2652fddad96bc454c612acdf/fluent-react/src/localization.ts#L33-L52

Maybe there can be another way to pass these elements after getting the string, but I don't know how. I think it would be great to have an optional `elems` parameter inside the `getString` directly, so we can add elements directly with that API call. 

To be clear, we are not blocked on this since `Localized` is fixed now and we can continue using that. But it would be good to have this as well. What do you think?

Vinnl:
Fixes #544.

It's similar to `getString`, but returns a React Fragment instead
of a string. The main added benefit here is that it allows the
caller to pass in `elems` analogous to the similarly-named prop on
`<Localized>`.

We're mostly using `getString` in the Firefox Relay codebase, but have to switch to `<Localized>` every time we want to use an element in the translation string, which can be confusing to new contributors.

I do understand that generally it's best to be aligned on a new feature before submitting a PR, which I didn't do, so I want to emphasise that it's perfectly fine to let me know that this feature isn't desirable for one reason or another, or that I should redo everything if I want to get this in. It didn't seem like a feature that was too hard to implement, so I decided to just go ahead and give it a shot, with the worst case being a little wasted effort :)

### DIFF
--- a/fluent-react/src/localization.ts
+++ b/fluent-react/src/localization.ts
@@ -1,6 +1,6 @@
 import { FluentBundle, FluentVariable } from "@fluent/bundle";
 import { mapBundleSync } from "@fluent/sequence";
-import { Fragment, ReactElement, createElement, isValidElement, cloneElement, ReactNode } from "react";
+import { Fragment, ReactElement, createElement, isValidElement, cloneElement } from "react";
 import { CachedSyncIterable } from "cached-iterable";
 import { createParseMarkup, MarkupParser } from "./markup.js";
 import voidElementTags from "../vendor/voidElementTags.js";
@@ -80,7 +80,7 @@ export class ReactLocalization {
   }
 
   getElement(
-    component: ReactNode | Array<ReactNode>,
+    componentToRender: ReactElement,
     id: string,
     args?: {
       vars?: Record<string, FluentVariable>,
@@ -88,24 +88,6 @@ export class ReactLocalization {
       attrs?: Record<string, boolean>;
     },
   ): ReactElement {
-    let componentToRender: ReactNode | null;
-
-    // Validate that the child element isn't an array that contains multiple
-    // elements.
-    if (Array.isArray(component)) {
-      if (component.length > 1) {
-        throw new Error(
-          "Expected to receive a single React element to localize."
-        );
-      }
-
-      // If it's an array with zero or one element, we can directly get the first
-      // one.
-      componentToRender = component[0];
-    } else {
-      componentToRender = component ?? null;
-    }
-
     const bundle = this.getBundle(id);
     if (bundle === null) {
       if (!id) {
@@ -137,22 +119,6 @@ export class ReactLocalization {
     const msg = bundle.getMessage(id)!;
 
     let errors: Array<Error> = [];
-
-    // Check if the component to render is a valid element -- if not, then
-    // it's either null or a simple fallback string. No need to localize the
-    // attributes.
-    if (!isValidElement(componentToRender)) {
-      if (msg.value) {
-        // Replace the fallback string with the message value;
-        let value = bundle.formatPattern(msg.value, args?.vars, errors);
-        for (let error of errors) {
-          this.reportError(error);
-        }
-        return createElement(Fragment, null, value);
-      }
-
-      return createElement(Fragment, null, componentToRender);
-    }
 
     let localizedProps: Record<string, string> | undefined;
     // The default is to forbid all message attributes. If the attrs prop exists
--- a/fluent-react/src/localization.ts
+++ b/fluent-react/src/localization.ts
@@ -213,11 +213,11 @@ export class ReactLocalization {
       return cloneElement(componentToRender, localizedProps, messageValue);
     }
 
-    let elemsLower: Record<string, ReactElement>;
+    let elemsLower: Map<string, ReactElement>;
     if (args?.elems) {
-      elemsLower = {};
+      elemsLower = new Map();
       for (let [name, elem] of Object.entries(args?.elems)) {
-        elemsLower[name.toLowerCase()] = elem;
+        elemsLower.set(name.toLowerCase(), elem);
       }
     }
 
@@ -239,7 +239,7 @@ export class ReactLocalization {
         return childNode.textContent;
       }
 
-      const sourceChild = elemsLower[childName];
+      const sourceChild = elemsLower.get(childName);
 
       // Ignore elems which are not valid React elements.
       if (!isValidElement(sourceChild)) {
--- a/fluent-react/src/localization.ts
+++ b/fluent-react/src/localization.ts
@@ -44,7 +44,7 @@ export class ReactLocalization {
 
   getString(
     id: string,
-    args?: Record<string, FluentVariable> | null,
+    vars?: Record<string, FluentVariable> | null,
     fallback?: string
   ): string {
     const bundle = this.getBundle(id);
@@ -52,7 +52,7 @@ export class ReactLocalization {
       const msg = bundle.getMessage(id);
       if (msg && msg.value) {
         let errors: Array<Error> = [];
-        let value = bundle.formatPattern(msg.value, args, errors);
+        let value = bundle.formatPattern(msg.value, vars, errors);
         for (let error of errors) {
           this.reportError(error);
         }
--- a/fluent-react/src/localization.ts
+++ b/fluent-react/src/localization.ts
@@ -1,6 +1,12 @@
 import { FluentBundle, FluentVariable } from "@fluent/bundle";
 import { mapBundleSync } from "@fluent/sequence";
-import { Fragment, ReactElement, createElement, isValidElement, cloneElement } from "react";
+import {
+  Fragment,
+  ReactElement,
+  createElement,
+  isValidElement,
+  cloneElement,
+} from "react";
 import { CachedSyncIterable } from "cached-iterable";
 import { createParseMarkup, MarkupParser } from "./markup.js";
 import voidElementTags from "../vendor/voidElementTags.js";
@@ -83,10 +89,10 @@ export class ReactLocalization {
     sourceElement: ReactElement,
     id: string,
     args: {
-      vars?: Record<string, FluentVariable>,
-      elems?: Record<string, ReactElement>,
+      vars?: Record<string, FluentVariable>;
+      elems?: Record<string, ReactElement>;
       attrs?: Record<string, boolean>;
-    } = {},
+    } = {}
   ): ReactElement {
     const bundle = this.getBundle(id);
     if (bundle === null) {
@@ -145,7 +151,10 @@ export class ReactLocalization {
     // message value and do not pass it to cloneElement in order to avoid the
     // "void element tags must neither have `children` nor use
     // `dangerouslySetInnerHTML`" error.
-    if (typeof sourceElement.type === "string" && sourceElement.type in voidElementTags) {
+    if (
+      typeof sourceElement.type === "string" &&
+      sourceElement.type in voidElementTags
+    ) {
       return cloneElement(sourceElement, localizedProps);
     }
 
@@ -183,36 +192,38 @@ export class ReactLocalization {
     // If the message contains markup, parse it and try to match the children
     // found in the translation with the args passed to this function.
     const translationNodes = this.parseMarkup(messageValue);
-    const translatedChildren = translationNodes.map(({ nodeName, textContent }) => {
-      if (nodeName === "#text") {
-        return textContent;
+    const translatedChildren = translationNodes.map(
+      ({ nodeName, textContent }) => {
+        if (nodeName === "#text") {
+          return textContent;
+        }
+
+        const childName = nodeName.toLowerCase();
+        const sourceChild = elemsLower?.get(childName);
+
+        // If the child is not expected just take its textContent.
+        if (!sourceChild) {
+          return textContent;
+        }
+
+        // If the element passed in the elems prop is a known void element,
+        // explicitly dismiss any textContent which might have accidentally been
+        // defined in the translation to prevent the "void element tags must not
+        // have children" error.
+        if (
+          typeof sourceChild.type === "string" &&
+          sourceChild.type in voidElementTags
+        ) {
+          return sourceChild;
+        }
+
+        // TODO Protect contents of elements wrapped in <Localized>
+        // https://github.com/projectfluent/fluent.js/issues/184
+        // TODO  Control localizable attributes on elements passed as props
+        // https://github.com/projectfluent/fluent.js/issues/185
+        return cloneElement(sourceChild, undefined, textContent);
       }
-
-      const childName = nodeName.toLowerCase();
-      const sourceChild = elemsLower?.get(childName);
-
-      // If the child is not expected just take its textContent.
-      if (!sourceChild) {
-        return textContent;
-      }
-
-      // If the element passed in the elems prop is a known void element,
-      // explicitly dismiss any textContent which might have accidentally been
-      // defined in the translation to prevent the "void element tags must not
-      // have children" error.
-      if (
-        typeof sourceChild.type === "string" &&
-        sourceChild.type in voidElementTags
-      ) {
-        return sourceChild;
-      }
-
-      // TODO Protect contents of elements wrapped in <Localized>
-      // https://github.com/projectfluent/fluent.js/issues/184
-      // TODO  Control localizable attributes on elements passed as props
-      // https://github.com/projectfluent/fluent.js/issues/185
-      return cloneElement(sourceChild, undefined, textContent);
-    });
+    );
 
     return cloneElement(sourceElement, localizedProps, ...translatedChildren);
   }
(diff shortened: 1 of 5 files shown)


eemeli:
Apologies for the delay; I've been travelling and currently fluent.js reviews pretty much fall on me.

In general, I think I'd be fine with adding `getFragment()`, as it makes behaviour that's now only available via `Localized` usable from `Localization` as well.

I took a quick look at this PR; do I understand right that effectively you've combined code from `getString()` in `Localization`  with the `elems` and message attribute handling of `Localized` in order to create `getFragment()`? My main issue with this PR is that it ends up duplicating quite a bit of code, and I'd much rather have a single internal implementation that's called from two different places for that.

Vinnl:
No worries @eemeli, hope you enjoyed your travels :) 

Yes, you are correct about the approach I took. I understand the desire for a single internal implementation, but I'm hesitant about doing that myself - I find that introducing layers of abstraction ideally is informed by an idea of which lines of code are likely to change together in the future, which is hard to do for someone like me who's unfamiliar with the codebase and its history. The wrong abstractions are often worse than no abstraction, and I don't want to be responsible for that :) 

Possibly I could give it a shot if you'd have some pointers about the abstractions you'd like to see?

eemeli:
@Vinnl I'd be happy to have you poke at this. :)

My gut feeling would be to have it so that `Localized` calls `getFragment()` internally. If there's a reason why that wouldn't be a good idea, you'll probably find it on the way. 😇

Vinnl:
Thanks @eemeli, that made sense. There was a bunch of extra code in `<Localized>` dealing with its children and setting attributes. I've now added a `getElement` function that replicates that functionality, redefined `getFragment` to just call that, and redefined `<Localized>` to use it as well. That should cut down on a lot of similar code - let me know what you think.

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,195 @@ export class ReactLocalization {
...
+        !elemsLower ||
+        !Object.prototype.hasOwnProperty.call(elemsLower, childName)
+      ) {
+        return childNode.textContent;
+      }
+
+      const sourceChild = elemsLower[childName];
+
+      // Ignore elems which are not valid React elements.
+      if (!isValidElement(sourceChild)) {
```
eemeli, line 246: Nitpick: It's a bit odd to do this check only now rather than when building `elemsLower`.

eemeli (changes requested):
I like this! I do think we need to bikeshed the API a bit, though. See inline comments.

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,195 @@ export class ReactLocalization {
...
 
+  getFragment(
+    id: string,
+    args?: {
+      vars?: Record<string, FluentVariable>,
+      elems?: Record<string, ReactElement>,
+    },
+    fallback?: string
+  ): ReactFragment {
+    return this.getElement(createElement(Fragment, null, fallback || id), id, args);
```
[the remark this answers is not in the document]
Vinnl: Hmm, it's technically not doing much, but I think from the point of view of the mental model of the developer, having the this method still makes sense. The use case is essentially "I want `getString` but with tags in there", rather than "I want this element with the contents replaced".

As an example, I'd prefer to use the following in my code:

```js
  <SomeOtherComponent heading={l10n.getFragment("some-heading", { elems: { b: <b/> } })} />
```

over

```js
  // Not even sure if this is possible:
  <SomeOtherComponent heading={l10n.Element(</>, "some-heading", { elems: { b: <b/> } })} />
  // Otherwise maybe you'd need to do:
  <SomeOtherComponent heading={l10n.Element(<></>, "some-heading", { elems: { b: <b/> } })} />
  // Or even this?
  <SomeOtherComponent heading={l10n.Element(React.createElement(React.Fragment), "some-heading", { elems: { b: <b/> } })} />
```

But let me know, I can remove it if you prefer.

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,195 @@ export class ReactLocalization {
...
+  }
+
+  getElement(
+    component: ReactNode | Array<ReactNode>,
+    id: string,
+    args?: {
+      vars?: Record<string, FluentVariable>,
+      elems?: Record<string, ReactElement>,
+      attrs?: Record<string, boolean>;
+    },
```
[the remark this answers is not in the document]
Vinnl: Oh heh, good eye. I'd actually argue that renaming `getString`s second argument to `vars` would be more appropriate, since the Fluent docs also [call them variables](https://projectfluent.org/fluent/guide/variables.html), but I can understand that changing the name of an existing API might not be desirable.

I could name it something like `props`? Because they're basically analogous to `<Localized>`'s props, although it could also be confusing, given that they're not React props. Or how about `params`?

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,195 @@ export class ReactLocalization {
...
+  ): ReactElement {
+    let componentToRender: ReactNode | null;
+
+    // Validate that the child element isn't an array that contains multiple
+    // elements.
+    if (Array.isArray(component)) {
+      if (component.length > 1) {
+        throw new Error(
+          "<Localized/> expected to receive a single React node child"
```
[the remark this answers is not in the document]
Vinnl: Whoops: 435a08be

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,195 @@ export class ReactLocalization {
...
+      // If it's an array with zero or one element, we can directly get the first
+      // one.
+      componentToRender = component[0];
+    } else {
+      componentToRender = component ?? null;
+    }
+
+    const bundle = this.getBundle(id);
+    if (bundle === null) {
+      if (id === undefined) {
```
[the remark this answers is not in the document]
Vinnl: Updated in 29be3ddf, but also noting that this is a change in behaviour - [the current code](https://github.com/projectfluent/fluent.js/blob/8c2eae431ef89f22d6d07f860e5043fe4eac3fc3/fluent-react/src/localized.ts#L77) only checks for `undefined`. (But this change makes sense to me.)

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,195 @@ export class ReactLocalization {
...
+        this.reportError(
+          new Error("No string id was provided when localizing a component.")
+        );
+      } else {
+        if (this.areBundlesEmpty()) {
```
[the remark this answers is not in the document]
Vinnl: ab17fdb5

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,195 @@ export class ReactLocalization {
...
+    // bundle.getMessage returns an existing message.
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const msg = bundle.getMessage(id)!;
+
+    let errors: Array<Error> = [];
+
+    // Check if the component to render is a valid element -- if not, then
+    // it's either null or a simple fallback string. No need to localize the
+    // attributes.
+    if (!isValidElement(componentToRender)) {
```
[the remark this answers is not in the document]
Vinnl: Hmm, I'm not sure what the best way to do that is - there'd be quite a bit of duplication in initialising the bundle and handling its various error cases. Personally I wouldn't be too worried about doing _too much_ error handling. What I could do is update the type definition of `getElement` to only allow `ReactChild | ReactFragment`? Then we're technically requiring it to be a valid element in the public API.

Or, hmm, what I could also do is update `<Localized>` to check for non-element children and, if found, wrap them in a fragment and pass that to `getElement`? That would cause the latter to do some unnecessary work, but nothing too egregious I thinkg - what do you think?

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,195 @@ export class ReactLocalization {
...
+      elemsLower = {};
+      for (let [name, elem] of Object.entries(args?.elems)) {
+        elemsLower[name.toLowerCase()] = elem;
+      }
+    }
+
+    // If the message contains markup, parse it and try to match the children
+    // found in the translation with the args passed to this function.
+    const translationNodes = this.parseMarkup(messageValue);
+    const translatedChildren = translationNodes.map(childNode => {
```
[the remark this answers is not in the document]
Vinnl: 87c73d4d

fluent-react/src/localization.ts
Vinnl: Hmm, I'm not entirely sure how this code works (it also [already existed](https://github.com/projectfluent/fluent.js/blob/8c2eae431ef89f22d6d07f860e5043fe4eac3fc3/fluent-react/src/localized.ts#L200)), but I think "ignore" here means "preserve, but do not translate" rather than "filter out"?

fluent-react/src/localization.ts
```diff
@@ -106,7 +106,7 @@ export class ReactLocalization {
     if (Array.isArray(component)) {
       if (component.length > 1) {
         throw new Error(
-          "<Localized/> expected to receive a single React node child"
+          "Expected to receive a single React node element to inject a localized string into."
```
eemeli, line 109:
```suggestion
          "Expected to receive a single React element to localize"
```
Adding more words doesn't really help here.

fluent-react/src/localization.ts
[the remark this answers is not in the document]
eemeli: It should be pretty easy to define your own sugar to use that API, for example:

```jsx
const getL10nFragment = (id, args) => l10n.getElement(createElement(Fragment, null, id), id, args)

<SomeOtherComponent
  heading={getL10nFragment("some-heading", { elems: { b: <b/> } })}
/>
```

The reason why I'd prefer leaving this out of the API is because it's that easy, and because we're defining a public API that we need to continue to maintain. The smaller the API, the less there is to document and maintain.

fluent-react/src/localization.ts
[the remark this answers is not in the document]
eemeli: I'm ok with changing the internal name of the `getSrting()` argument to `vars`, because it only impacts the documentation.

`props` and `params` aren't really better, and both come with their own sets of baggage.

fluent-react/src/localization.ts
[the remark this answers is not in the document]
eemeli: The change in behaviour should only slightly improve how we're treating an error case, so it's fine.

fluent-react/src/localization.ts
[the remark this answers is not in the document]
eemeli: How about calling `getString()` rather than `getElement()` from a Localized with non-element children, and wrapping the result in a Fragment? As far as I can tell, that should replicate the current behaviour.

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,194 @@ export class ReactLocalization {
...
+        return textContent;
+      }
+
+      const childName = nodeName.toLowerCase();
+
+      // If the child is not expected just take its textContent.
+      if (
+        !elemsLower ||
+        !Object.prototype.hasOwnProperty.call(elemsLower, childName)
+      ) {
```
eemeli, line 238:
```suggestion
      if (!elemsLower?.has(childName)) {
```

fluent-react/src/localization.ts
eemeli: I meant that this check could/should be somewhere around line 220 instead, so that only valid elements are ever added into `elemsLower`.

eemeli (changes requested):
See replies above for requested changes, and a couple minor new inline comments.

fluent-react/src/localization.ts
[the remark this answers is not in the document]
Vinnl: Yeah that's a fine workaround, and indeed still adding this later would be easier than removing it: 6027f435

fluent-react/src/localization.ts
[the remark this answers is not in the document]
Vinnl: Ah yeah, that's clever: 59671ccd

I should note that that does change the behaviour a little bit: if you do not provide a child, it will now render the string ID, rather than `null`. Which makes sense to me and aligns it with the existing behaviour of `getString()`, but I can also imagine that that's not desirable.

One way to work around that might be to set the fallback string to some unique string, and return `null` instead of a fragment if `getString` returns the fallback string, but that feels pretty hacky :/ Alternatively, although still a change in behaviour but closer to the previous behaviour, would be to set the fallback string to the empty string, resulting in it returning a fragment containing an empty string instead of `null`. That's technically different from what it used to be, but should have the same result for the user.

fluent-react/src/localization.ts
Vinnl: Ah gotcha. b950f1c7

fluent-react/src/localization.ts
Vinnl: 998e7866

fluent-react/src/localization.ts
Vinnl: Oh heh, was just typing a reply above that I'd forgotten to run the tests after the change to `Map`. Fix in 515918e7c701859f98930b1df2de61fc211cb866.

Vinnl:
@eemeli Just wondering if you have time to look at my latest changes - I think I've addressed all your comments :)

fluent-react/src/localized.ts
```diff
@@ -1,4 +1,5 @@
-import {
+import React, {
+  isValidElement,
```
eemeli, line 2:
```suggestion
import React, {
  Children,
  isValidElement,
```

fluent-react/src/localized.ts
```diff
@@ -45,7 +46,40 @@ export function Localized(props: LocalizedProps): ReactElement {
...
+        "Expected to receive a single React element to localize."
+      );
+    }
+
+    // If it's an array with zero or one element, we can directly get the first
+    // one.
+    componentToRender = children[0];
+  } else {
+    componentToRender = children ?? null;
+  }
```
eemeli, line 65: I'm pretty sure that this does effectively the same thing:

```suggestion
  const componentToRender = Children.only(children);
```

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,144 @@ export class ReactLocalization {
...
   }
 
+  getElement(
+    componentToRender: ReactElement,
+    id: string,
+    args?: {
+      vars?: Record<string, FluentVariable>,
+      elems?: Record<string, ReactElement>,
+      attrs?: Record<string, boolean>;
+    },
```
eemeli, line 89:
```suggestion
    args: {
      vars?: Record<string, FluentVariable>,
      elems?: Record<string, ReactElement>,
      attrs?: Record<string, boolean>;
    } = {},
```

This removes the need to continuously check e.g. `args?.vars`.

fluent-react/src/localization.ts
```diff
@@ -73,6 +79,144 @@ export class ReactLocalization {
     return fallback || id;
   }
 
+  getElement(
+    componentToRender: ReactElement,
```
eemeli, line 83: As the type indicates, this argument is an "element". Calling it such would be clearer, maybe something like `sourceElement`?

eemeli (changes requested):
@Vinnl Sorry for the delay here, finally managed to take a look. A few small changes, and then I think we're good?

fluent-react/src/localized.ts
Vinnl: Almost, though it's definitely nicer to delegate the error handling to React. But `Children.only` also throws when given no children (which Fluent React accepts - it just renders the localised string), when given a text node, or when given an array with a single element. Thus, I did have to still add explicit checks for that: 71dfc622

fluent-react/src/localization.ts
Vinnl: Good one: dc130a591aaaa4023650b7be16808f38d70ed1be

fluent-react/src/localization.ts
Vinnl: Yeah that works. Can get lost in the weeds when moving code around from one place to another :sweat_smile: dbfbdb53

fluent-react/src/localized.ts
Vinnl: This was part of https://github.com/projectfluent/fluent.js/pull/595#discussion_r1006189103.

fluent-react/src/localized.ts
eemeli: Ah, I was wrong and `Children.only()` is less useful than I remembered. So let's not do that.

Vinnl:
Thank you so much @eemeli!

PR #602, opened 2022-12-30
use Intl.Locale for langneg

ImUrX:
Requires fixing some test errors


eemeli (changes requested):
The test errors seem rather significant here, as they're changing the behaviour of the locale matching. Given the different treatment of subtags like `macos` and `ca-japanese` in our `locale.ts` and the standard `Intl.Locale`, I'm not sure that these are easily reconcilable without some observable changes.

Effectively, this is pointing out that our current understanding of locale tags is nonstandard, and that's a bit of a problem.

I wonder if we could use `Intl.Locale` for matching on language, region & script, and do something else for the rest?

ImUrX:
Yes as you can see I handle the private-usage part by just reading the last subtag (because its always the last one), what should I do?

ImUrX:
I also personally ended up using a ponyfill of the [Intl.LocaleMatcher](https://github.com/tc39/proposal-intl-localematcher) standard in my code.

eemeli:
> Yes as you can see I handle the private-usage part by just reading the last subtag (because its always the last one), what should I do?

I see that. Hmmm. Could you make use of the normalization done in the `baseName`? As in, split it by `-` and compare the parts for each locale? That might allow for the implementation change to be considered a relaxation of the variant regexp.

PR #601, opened 2022-12-30
Keep Locale class up to standard

ImUrX:
- [ ] #600 
- [ ] Add support for extended language subtags (up to three)
- [ ] Add support for extension subtags
- [ ] Add support for private-use subtag

Issue #600, opened 2022-12-30
Locale regex doesnt take into account UN.M49


ImUrX:
This is a big mistake, it's usually used for things like ``es-419`` which isn't working in my case :sob:

ImUrX:
[seeing wikipedia](https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags), the regex itself is missing a lot of features

PR #605, opened 2023-02-07
Use Intl.Locale internally

### LINKED ISSUE #551: [fluent-langneg] Support BCP47 extended language subtags
Currently, https://github.com/projectfluent/fluent.js/blob/8991a1760a83f1f9d8900e325ecc83dcb636337c/fluent-langneg/src/locale.ts only allow the primary language subtag, and [extended language subtag](https://tools.ietf.org/html/bcp47#section-2.2.2) is not implemented.

It also does not take into account of macrolanguages. For example, requesting `zh` when there is `cmn-Hans` available will not match `cmn-Hans`.

zbraniecki:
Alternative to #602 and #601 approach to fix #551 and #600.

I'm using `Intl.Locale` internally, but wrapping it for the masks use. This passes all tests and hopefully will make our alignment with ECMA-402 better.

It also make us pass meta-zones and other BCP47 pieces using internal runtime parser.


fluent-langneg/src/locale.ts
```diff
@@ -1,30 +1,32 @@
...
+    result = locale;
+  };
+  return result.replace(/\-\*/g, "");
+}
 
-const languageCodeRe = "([a-z]{2,3}|\\*)";
-const scriptCodeRe = "(?:-([a-z]{4}|\\*))";
-const regionCodeRe = "(?:-([a-z]{2}|\\*))";
-const variantCodeRe = "(?:-(([0-9][a-z0-9]{3}|[a-z0-9]{5,8})|\\*))";
+function getVisibleLangTagLength(language: any, script: any, region: any) {
```
eemeli, line 11: These should really be typed as `string | undefined`; there should be no need for `any` here.

fluent-langneg/src/locale.ts
```diff
@@ -39,29 +41,55 @@ export class Locale {
...
+    try {
+      result = new Intl.Locale(normalized);
+    } catch (e) {
       this.isWellFormed = false;
+      this.language = "und";
       return;
     }
 
-    let [, language, script, region, variant] = result;
+    this.language = result.language || "und";
```
eemeli, line 54: How could `result.language` ever be empty?

fluent-langneg/src/locale.ts
```diff
@@ -39,29 +41,55 @@ export class Locale {
...
+
+    if (normalized.length > visiblelangTagLength) {
+      let extStart = getExtensionStart(locale);
+      this.variant = locale.substring(visiblelangTagLength + 1, extStart);
+    }
+
+    this.isWellFormed = true;
+  }
+
+  static fromComponents({language, script, region, variant}: {
```
eemeli, line 68: Could/should this also use `Intl.Locale`? The original included some normalizations that are included in `Intl.Locale` processing, but they're dropped in these changes. Code here could construct an options bag to pass as the second arg of the Intl constructor, and then read the language/script/region from the result.

fluent-langneg/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "target": "es2020",
```
eemeli, line 4: Is this enough? I was under the impression that TS only added the `Intl.Locale` types in `es2021`.

eemeli (changes requested):
Overall quite happy with this. I'm a little surprised that checking the string length works, but indeed it apparently does, as the canonicalizations never change the subtag length. Some nitpicky line comments, and it looks like we may need to drop Node.js 12 CI tests.

I'd be fine inlining the constructor code that's here moved to separate functions, but this isn't a strong stylistic opinion, as long as the linter is satisfied.

eemeli (approved):
Given how long this has been sitting around, I figured it'd be ok to take over the PR and update it somewhat. The new shape of this effectively uses `Intl.Locale`, with effectively a polyfill for `.variants`, which so far is only supported by Firefox.

@zbraniecki Could you take a look at this? I've kept all of the old functionality as far as possible (including the nonstandard handling of `*` as a subtag), but switching from a custom set of likely subtags to the one provided by `Intl.Locale` means that there's a small change at least for `sr-RU`, which now maximises to `sr-Curl-RU` -- which seems more correct than `sr-Latn-RU`.

fluent-langneg/test/langneg_test.js
```diff
@@ -42,11 +44,11 @@ const data = {
       ],
       [["fr"], ["fr-CA", "fr-FR"], ["fr-FR", "fr-CA"]],
       [["az-IR"], ["az-Latn", "az-Arab"], ["az-Arab"]],
-      [["sr-RU"], ["sr-Cyrl", "sr-Latn"], ["sr-Latn"]],
+      [["sr-RU"], ["sr-Cyrl", "sr-Latn"], ["sr-Cyrl"], undefined, 20],
```
zbraniecki: `question`: What's that?

eemeli: Ah, yeah, the last value is `minNodeVersion`, and this needs to be:
```suggestion
      [["sr-RU"], ["sr-Cyrl", "sr-Latn"], ["sr-Cyrl"], undefined, 22],
```

On Node.js 20 and earlier, `new Intl.Locale('sr-RU').maximize().toString()` is `sr-Latn-RU`, while on later versions it's `sr-Cyrl-RU`, due to an ICU/CLDR change.

PR #606, opened 2023-02-08
Treat tab as text, not whitespace

eemeli:
See projectfluent/python-fluent#173

Tabs were defined to be text, not whitespace, in projectfluent/fluent#167, but this change was not reflected in the trailing whitespace trimming that was added in #247.


gregtatum (approved):
LGTM

PR #607, opened 2023-02-15
Update dependencies & API docs

eemeli:
Update all the dependencies (except for `commander`, as in #575). This includes an update to `typedoc` that complains about more things in the code comments, so many of those are adjusted here as well; the changes should also help with including the same info in code completion popups.

Some of the updated dev dependencies require Node.js v14, so the minimum support level needs to be correspondingly raised. The Node.js v12 CI tests are also dropped.


gregtatum (approved):
Everything seems reasonable.

PR #611, opened 2023-03-22
Rename default branch from master to main

eemeli:
These changes should go together with the actual rename of the default branch, which should be pretty painless.


gregtatum (approved):
Looks good to me!

PR #612, opened 2023-03-22
Update dependencies & fluent-react/example

### LINKED ISSUE #610: @fluent/react marks @fluent/bundle as incompatible
`@fluent/react` says a version lower than 0.18 of `@fluent/bundle` is a peerDependency:

https://github.com/projectfluent/fluent.js/blob/67e038ec83f7136fea29af09a2b080df94e074d3/fluent-react/package.json#L65

But that version just got released last week:

https://github.com/projectfluent/fluent.js/blob/67e038ec83f7136fea29af09a2b080df94e074d3/fluent-bundle/package.json#L4

I assume the two should be compatible?

eemeli:
Fixes #610, as the fluent-react `peerDependency` condition was indeed too strict.

With modern `npm`, we don't need to use `file:` paths for internal dependencies and can also drop support of `npm<7` in the lockfile.

### DIFF
--- a/fluent-bundle/package.json
+++ b/fluent-bundle/package.json
@@ -49,6 +49,6 @@
     "npm": ">=7.0.0"
   },
   "devDependencies": {
-    "@fluent/dedent": "file:../fluent-dedent"
+    "@fluent/dedent": "^0.5.0"
   }
 }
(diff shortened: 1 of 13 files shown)


gregtatum (approved):
It all seems reasonable!

Vinnl:
Thanks for this! Could this also be released as a new version?

Issue #411, opened 2019-07-29
Control error reporting via a LocalizationProvider prop


stasm:
We should allow users to control exactly how formatting errors are handled in `Localized` and `getString`. We could do it via a `<LocalizationProvider reportError={...}>` prop, which is a function taking an error.

akdetrick:
I have a real world use case for this. It's common to configure production error reporting to ignore anything at warning level or below. This is an example of a function taking an error that would capture the key miss logging at error level:

```js
const reportKeyMiss = (fluentKeyMissWarning) => {
  Dashboard.withScope((scope) => {
    scope.setLevel("error");
    Dashboard.captureException(fluentKeyMissWarning);
  });
}
```

digeomel:
It is unfortunate that this was done only for React. I'm using `ngx-pdf-extended-viewer` with Angular, which in turn wraps pdf.js, which finally is using Fluent and is dumping a lot of console warnings about missing IDs which I don't want to see.
A global, non-framework-specific solution would be nice.

PR #619, opened 2023-07-31
feat(react): allow user provided reportError func

StaberindeZA:
Because:

* We should allow users the ability to provide a reportError function.

This commit:

* Adds a class property, reportError, to ReactLocalization that allows
  consumers to provide a function on initialization to report errors.
  By default the errors will console.warn.

Closes: #411

### DIFF
--- a/fluent-react/src/localization.ts
+++ b/fluent-react/src/localization.ts
@@ -15,6 +15,12 @@ import voidElementTags from "../vendor/voidElementTags.js";
 // &amp;, &#0038;, &#x0026;.
 const reMarkup = /<|&#?\w+;/;
 
+const defaultReportError = (error: Error): void => {
+  /* global console */
+  // eslint-disable-next-line no-console
+  console.warn(`[@fluent/react] ${error.name}: ${error.message}`);
+};
+
 /**
  * `ReactLocalization` handles translation formatting and fallback.
  *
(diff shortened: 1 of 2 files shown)


eemeli (changes requested):
When @stasm originally suggested using a provider prop for this, the code was structured differently, and that may have made more sense. Now, all of the error reporting calls are made from within ReactLocalization, so really the control should be via an argument of its constructor, rather than as a side effect of including it in a provider. Also, why make it a boolean rather than a function accepting an error argument?

StaberindeZA:
> When @stasm originally suggested using a provider prop for this, the code was structured differently, and that may have made more sense. Now, all of the error reporting calls are made from within ReactLocalization, so really the control should be via an argument of its constructor, rather than as a side effect of including it in a provider. Also, why make it a boolean rather than a function accepting an error argument?

Thank you for the quick reply and comments. Hmm, I did not notice the age of the issue. 😅 Your suggestions make sense. 

No specific reason for the boolean. For some reason when I initially read the issue, I missed some comments, and had in my mind that the ask was for an on/off toggle. Agreed though the function is better, and I've made the changes.

fluent-react/src/localization.ts
```diff
@@ -29,13 +35,16 @@ const reMarkup = /<|&#?\w+;/;
 export class ReactLocalization {
   public bundles: Iterable<FluentBundle>;
   public parseMarkup: MarkupParser | null;
+  private reportError: (error: Error) => void;
```
eemeli, line 38: Making this private would be a breaking change, given that it's currently public.

```suggestion
  public reportError: (error: Error) => void;
```

eemeli (approved):
Much better! One small change, and this should be good to merge.

PR #623, opened 2023-10-20
Make pauseObserving and resumeObserving public

calixteman:
This is consistent with mozilla-central.


eemeli (approved):
Given the minimum supported version updates in #627, the breaking changes here won't add a new semver-major update requirement. (well, semver-minor given that we're still 0.x)

Issue #634, opened 2025-02-12
Additional variable types / Temproal support


rkh:
Currently, fluent-bundle cannot handle [Temporal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal).

# What would need to change

It is hardcoded only to support FluentVariable entries (FluentType, string, number, or Date).

Here's how they are resolved:

https://github.com/projectfluent/fluent.js/blob/48e2a62ed45ff2a62a231b2e83cfd8b332d27acb/fluent-bundle/src/resolver.ts#L178-L200

It is, therefore, not possible to pass in other values.

## Possible fixes

* Built-in support for Temporal objects
* A way to pass in a function to convert custom objects into FluentType instances
* All of the above

The first one would be very convenient but would also require checking whether Temporal exists.

# Project status

As a side note, I'm seeing a lot of old issues with no progress. I'm happy to work on a PR for this, but I'm wondering what the maintenance status of the project is.

rkh:
Example of what the API could look like:

```ts
new FluentBundle(locales, {
  wrapVariable(value: Temporal.ZonedDateTime | FluentVariable): FluentVariable {
    if(value instanceof Temporal.ZonedDateTime) {
      return new FluentDateTime(value.epochMilliseconds, { timeZone: value.timeZoneId })
    } else {
      return value
    }
  }
})
```

eemeli:
The library's still maintained, but new feature development is indeed not very active. I'd be very interested in reviewing new PRs, including ones bringing in Temporal support.

rkh:
Pull Request: #635

rkh:
Alternative implementation (more complete Temporal support, but no API for custom values): #636 – happy to integrate this into #635 if that's the preferred approach.

rkh:
@eemeli I assume this can be closed now?

eemeli:
Completed in #636. Release due shortly, once I get some of the dev dependencies refreshed.

PR #635, opened 2025-02-13
Variable casting and Temporal support

rkh:
Implements #634

## Variable casting

Went for a slightly different API than in the issue.

```typescript
bundle = new FluentBundle("en-US")

// these aren't needed, as they come built-in, but to illustrate the new API
bundle.addCast("number", FluentNumber)
bundle.addCast(Date, (value: Date) => new FluentDateTime(value.getTime()));
```

## Temporal support

I've put this in a separate package for now, as Temporal is still considered experimental.

It also currently only supports `Instant` and `ZonedDateTime`.

Best illustration is probably [bundle_test.js](https://github.com/rkh/fluent.js/blob/801a3646af093e3ab37e876a19fd81bfe25ca781/fluent-temporal/test/bundle_test.js#L26-L46).

### DIFF
--- a/fluent-temporal/package.json
+++ b/fluent-temporal/package.json
@@ -1,0 +1,56 @@
+{
+  "name": "@fluent/temporal",
+  "description": "Experimental Temporal support for Project Fluent",
+  "version": "0.1.0",
+  "homepage": "https://projectfluent.org",
+  "author": "Mozilla <l10n-drivers@mozilla.org>",
+  "license": "Apache-2.0",
+  "contributors": [
+    {
+      "name": "Konstantin Haase",
+      "email": "github@rkh.im"
+    }
+  ],
+  "main": "./index.js",
+  "module": "./esm/index.js",
+  "types": "./esm/index.d.ts",
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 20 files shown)


fluent-temporal/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@fluent/temporal",
+  "description": "Experimental Temporal support for Project Fluent",
+  "version": "0.1.0",
+  "homepage": "https://projectfluent.org",
+  "author": "Mozilla <l10n-drivers@mozilla.org>",
+  "license": "Apache-2.0",
+  "contributors": [
+    {
+      "name": "Konstantin Haase",
```
rkh, line 10: Put myself in here. You can take me out, though. I'm happy to "gift" this little library to Mozilla.

rkh:
I am happy to work on the remaining Temporal classes, especially Duration.

We are working on a production app (pre-alpha atm) where we use a Temporal polyfill and Fluent, and we're expecting to use almost all Temporal classes eventually, so I'm also happy to keep an eye on any bugs etc that pop up in the future.

PR #636, opened 2025-02-15
Add Temporal support

rkh:
This is an alternative to #635, adding only support for Temporal time objects without implementing a public variable caster API.

It still has some special handling to work fine in environments without Temporal. It also avoids running any such checks at load time, to allow loading a Temporal polyfill after loading fluent-bundle.

## Duration

It does not implement `Temporal.Duration` support yet, which would require:
* Augmenting built-in TypeScript `Intl` namespace to support `Intl.DurationFormat` and `Intl.DurationFormatOptions`
* Add signature for `Intl.DurationFormat` to `scope.memoizeIntlObject`
* Add `FluentDuration` wrapper for `Temporal.Duration`
* Optional: Add `DURATION` builtin

Happy to work on that as well, but can also create a separate PR for that.

## FluentVariable

I've taken the liberty to move `FluentVariable` to `types.js` – that made more sense to me, especially with the expanded list.

### DIFF
--- a/fluent-bundle/src/types.ts
+++ b/fluent-bundle/src/types.ts
@@ -11,7 +11,6 @@ export type FluentVariable =
   | Temporal.PlainTime
   | Temporal.PlainYearMonth
   | Temporal.PlainMonthDay
-  | Temporal.ZonedDateTime
   | string
   | number
   | Date;
@@ -132,7 +131,6 @@ export class FluentDateTime extends FluentType<
   | Temporal.PlainMonthDay
   | Temporal.PlainTime
   | Temporal.PlainYearMonth
-  | Temporal.ZonedDateTime
 > {
   /** Options passed to `Intl.DateTimeFormat`. */
   public opts: Intl.DateTimeFormatOptions;
@@ -150,8 +148,7 @@ export class FluentDateTime extends FluentType<
         value instanceof Temporal.PlainDate      || // @ts-ignore
         value instanceof Temporal.PlainMonthDay  || // @ts-ignore
         value instanceof Temporal.PlainTime      || // @ts-ignore
-        value instanceof Temporal.PlainYearMonth || // @ts-ignore
-        value instanceof Temporal.ZonedDateTime
+        value instanceof Temporal.PlainYearMonth
       ) {
         return true;
       }
@@ -176,7 +173,6 @@ export class FluentDateTime extends FluentType<
       | Temporal.PlainMonthDay
       | Temporal.PlainTime
       | Temporal.PlainYearMonth
-      | Temporal.ZonedDateTime
       | FluentDateTime
       | FluentType<number>,
     opts: Intl.DateTimeFormatOptions = {}
@@ -189,29 +185,9 @@ export class FluentDateTime extends FluentType<
       value = value.valueOf();
     }
 
-    if (typeof value === "object") {
-      // Intl.DateTimeFormat defaults to gregorian calendar, but Temporal defaults to iso8601
-      if ('calendarId' in value) {
-        if (opts.calendar === undefined) {
-          opts = { ...opts, calendar: value.calendarId };
-        } else if (opts.calendar !== value.calendarId && 'withCalendar' in value) {
-          value = value.withCalendar(opts.calendar);
-        }
-      }
-
-      // Temporal.ZonedDateTime is timezone aware
-      if ('timeZoneId' in value) {
-        if (opts.timeZone === undefined) {
-          opts = { ...opts, timeZone: value.timeZoneId };
-        } else if (opts.timeZone !== value.timeZoneId && 'withTimeZone' in value) {
-          value = value.withTimeZone(opts.timeZone);
-        }
-      }
-
-      // Temporal.ZonedDateTime cannot be formatted directly
-      if ('toInstant' in value) {
-        value = value.toInstant();
-      }
+    // Intl.DateTimeFormat defaults to gregorian calendar, but Temporal defaults to iso8601
+    if (typeof value === "object" && 'calendarId' in value && opts.calendar === undefined) {
+      opts = { ...opts, calendar: value.calendarId };
     }
 
     super(value);
--- a/fluent-bundle/src/types.ts
+++ b/fluent-bundle/src/types.ts
@@ -1,6 +1,19 @@
 import { Scope } from "./scope.js";
+import type { Temporal } from "temporal-polyfill";
 
 export type FluentValue = FluentType<unknown> | string;
+
+export type FluentVariable =
+  | FluentValue
+  | Temporal.Instant
+  | Temporal.PlainDateTime
+  | Temporal.PlainDate
+  | Temporal.PlainTime
+  | Temporal.PlainYearMonth
+  | Temporal.PlainMonthDay
+  | Temporal.ZonedDateTime
+  | string
+  | number;
 
 export type FluentFunction = (
   positional: Array<FluentValue>,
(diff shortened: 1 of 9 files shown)


eemeli (changes requested):
Looks very promising!

fluent-bundle/src/temporal.d.ts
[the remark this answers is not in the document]
rkh: I do think `import type { Temporal } from "temporal-polyfill"` would be the most ideal, assuming their type definitions mirror the correct ones, and I think the built result would only impact type definitions.

For our app, I'm just straight up declaring the Temporal namespace, which makes this very easy, but that only works in build environments that don't support Temporal (compiled result works fine in those that do).

I was trying that initially, but ran into issues with using these as `instanceof` values, and `(globalThis as any).Temporal as Temporal` didn't work either. My first attempt ended up with quite a bit of `@ts-ignore` in there.

Can look into this further.

fluent-bundle/test/temporal_test.js
```diff
@@ -0,0 +1,186 @@
...
+      bundle = new FluentBundle("en-US", { useIsolating: false });
+      bundle.addResource(
+        new FluentResource(ftl`
+        direct = { $arg }
+        dt = { DATETIME($arg) }
+        month = { DATETIME($arg, month: "long", year: "numeric") }
+        timezone = { DATETIME($arg, timeZoneName: "shortGeneric") }
+        `)
+      );
+    }
```
Demivan, line 34: Setting up the bundle should not be done conditionally, just the pollyfilling.

fluent-bundle/src/types.ts
```diff
@@ -1,12 +1,34 @@
...
+function isTemporal(value: any, type: "PlainYearMonth"): value is FluentTemporal.PlainYearMonth;
+function isTemporal(value: any, type: "ZonedDateTime"): value is FluentTemporal.ZonedDateTime;
+function isTemporal(value: any, type: keyof FluentTemporal.TemporalContainer): boolean {
+  const temporal = (globalThis as any).Temporal as FluentTemporal.TemporalContainer | undefined;
+  const prototype = temporal && temporal[type];
+  if (!prototype) {
+    return false;
+  }
+  return value instanceof prototype;
+}
```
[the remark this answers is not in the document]
rkh: Ah yes, a Proxy could also solve deferring Temporal checks.

fluent-bundle/src/types.ts
```diff
@@ -109,32 +131,128 @@ export class FluentNumber extends FluentType<number> {
...
+      isTemporal(info, "PlainYearMonth") || isTemporal(info, "ZonedDateTime");
+  }
+
+  /**
+   * Converts various types representing dates and times to a `FluentDateTime`.
+   */
+  static from(
+    info: FluentDateInfo | FluentType<FluentDateInfo>,
+    opts: Intl.DateTimeFormatOptions = {}
+  ): FluentDateTime {
```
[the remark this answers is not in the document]
rkh: Good point. I was mirroring what Temporal itself does.

Demivan (changes requested):

fluent-bundle/src/temporal.d.ts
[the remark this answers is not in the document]
rkh: I have done this for now. However, this will make `temporal-polyfill` a dev dependency for anyone using `fluent-bundle` with TypeScript (see types.d.ts).

Could improve the situation by adding `temporal-spec` as an actual dependency or stating so in the readme.

fluent-bundle/src/types.ts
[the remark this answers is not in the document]
rkh: Solved it without the Proxy (the tests already demonstrate that it is better to resolve this at run time not at load time). Significantly reduced type checking overall with the latest adjustments.

fluent-bundle/src/types.ts
```diff
@@ -1,12 +1,34 @@
 import { Scope } from "./scope.js";
 
 export type FluentValue = FluentType<unknown> | string;
+export type FluentStrictDateValue = FluentTemporal.Plain | Date;
+export type FluentDateValue = FluentStrictDateValue | number;
+export type FluentDateInfo = FluentDateValue | Date | FluentTemporal.Instant | FluentTemporal.ZonedDateTime;
+export type FluentVariable = FluentValue | FluentDateInfo | string | number;
```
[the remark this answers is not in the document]
rkh: Removed and inlined.

rkh:
Incorporated the feedback and added more tests (different calendars, number conversion).

The main thing I see right now is the type import, which creates a TypeScript-only dependency on temporal-polyfill.

fluent-bundle/src/types.ts
```diff
@@ -1,7 +1,20 @@
...
+  | FluentValue
+  | Temporal.Instant
+  | Temporal.PlainDateTime
+  | Temporal.PlainDate
+  | Temporal.PlainTime
+  | Temporal.PlainYearMonth
+  | Temporal.PlainMonthDay
+  | Temporal.ZonedDateTime
+  | string
+  | number;
```
eemeli, line 16:
```suggestion
  | number
  | Date;
```

fluent-bundle/src/types.ts
```diff
@@ -104,37 +117,134 @@ export class FluentNumber extends FluentType<number> {
...
+    if ('Temporal' in globalThis) {
+      if (
+        // @ts-ignore
+        value instanceof Temporal.Instant        || // @ts-ignore
+        value instanceof Temporal.PlainDateTime  || // @ts-ignore
+        value instanceof Temporal.PlainDate      || // @ts-ignore
+        value instanceof Temporal.PlainMonthDay  || // @ts-ignore
+        value instanceof Temporal.PlainTime      || // @ts-ignore
+        value instanceof Temporal.PlainYearMonth || // @ts-ignore
+        value instanceof Temporal.ZonedDateTime
```
eemeli, line 153: Why are all these ts-ignore directives required? If they're needed, could ts-expect-error be used instead?

fluent-bundle/src/types.ts
```diff
@@ -104,37 +117,134 @@ export class FluentNumber extends FluentType<number> {
...
+  | Temporal.PlainDate
+  | Temporal.PlainMonthDay
+  | Temporal.PlainTime
+  | Temporal.PlainYearMonth
+  | Temporal.ZonedDateTime
+> {
   /** Options passed to `Intl.DateTimeFormat`. */
   public opts: Intl.DateTimeFormatOptions;
 
+  static supportsValue(value: any): value is ConstructorParameters<typeof Temporal.Instant>[0] {
```
eemeli, line 139: Isn't `ConstructorParameters<typeof Temporal.Instant>[0]` the `epochNanoseconds` bigint? That seems wrong here.

fluent-bundle/src/types.ts
```diff
@@ -104,37 +117,134 @@ export class FluentNumber extends FluentType<number> {
...
+
+    if (typeof value === "object") {
+      // Intl.DateTimeFormat defaults to gregorian calendar, but Temporal defaults to iso8601
+      if ('calendarId' in value) {
+        if (opts.calendar === undefined) {
+          opts = { ...opts, calendar: value.calendarId };
+        } else if (opts.calendar !== value.calendarId && 'withCalendar' in value) {
+          value = value.withCalendar(opts.calendar);
+        }
+      }
```
eemeli, line 199: This automatic conversion is probably a bad idea, esp. as the formatter option `calendar` is already excluded from `DATETIME_ALLOWED`.

fluent-bundle/src/types.ts
```diff
@@ -104,37 +117,134 @@ export class FluentNumber extends FluentType<number> {
...
+          opts = { ...opts, timeZone: value.timeZoneId };
+        } else if (opts.timeZone !== value.timeZoneId && 'withTimeZone' in value) {
+          value = value.withTimeZone(opts.timeZone);
+        }
+      }
+
+      // Temporal.ZonedDateTime cannot be formatted directly
+      if ('toInstant' in value) {
+        value = value.toInstant();
+      }
```
eemeli, line 213: This automatic conversion is also a bad idea. As your comment notes, zoned datetimes are not formattable at least for now, so let's stick to that in Fluent as well.

eemeli (changes requested):
I'm a bit concerned about the temporal types getting included in the FluentVariable and FluentDateTime external types, as that's effectively making the `temporal-spec` package a runtime dependency. Could that be avoided somehow, even at the cost of making the types less exact?

fluent-bundle/src/types.ts
rkh: Yes, that is wrong, not sure how that happened. I meant it to be `typeof FluentDateTime`.

fluent-bundle/src/types.ts
rkh: They are needed because `Temporal` is only imported as a type, but not declared. Will look into `ts-expect-error`.

fluent-bundle/src/types.ts
rkh: Okay, makes sense. But setting the option is acceptable? I wanted to avoid users having to create their own `FluentDateTime`, and as I said in the comment, `Temporal` defaults to a different calendar than `Intl`.

rkh:
> I'm a bit concerned about the temporal types getting included in the FluentVariable and FluentDateTime external types, as that's effectively making the `temporal-spec` package a runtime dependency. Could that be avoided somehow, even at the cost of making the types less exact?

Yes, that's what I meant with my comment as well. Could type them as Temporal-like objects (which is a bit what I was trying to do in the first version). It would make more sense with what the code is like now, as type checks are only done in a single place. I'll see what I can do without copying over the full type spec.

rkh:
FYI, I plan to work on this again later in the week.

fluent-bundle/src/types.ts
eemeli: Yeah, if no `calendar` value is set, then it's reasonable to default to the one in the temporal instance.

rkh: That sounds good (and reasonable). I do wish that this was the `Intl.DateTimeFormat` behavior to begin with.

rkh:
Defined a minimal interface for the few fields the code uses from Temporal. The main issue I can see with that one is that TypeScript at least won't catch if an object is passed in that matches the interface but isn't a Temporal object.

But that's probably preferable over a dependency or shipping types for all the various Temporal classes.

This can be dropped as soon as TypeScript comes with Temporal support.

eemeli (approved):
This looks good! Sorry for taking so long to reply; I've been on vacation. Once the tests pass, this is good to merge.

rkh:
All good. Hope you had a nice vacation.
We've solved this differently for our app weeks ago, so this isn't a blocker for us.
Tests should be passing now.

rkh:
Can reproduce the CI failure locally (seems to be Node 18 specific, I wonder if the polyfill is broken on that version).

rkh:
I did a one line check whether iso8601 is supported or not, and skip the tests if it isn't. Probably out of scope for fluent.js to have a workaround.

PR #641, opened 2025-04-02
Fix FluentDateTime and FluentNumber primitive conversions

julienw:
This fixes a problem when using a custom function taking a FluentDateTime as a parameter, like we're doing in the profiler code.

We're relying on being able to convert a FluentDateTime value to a number:
https://github.com/firefox-devtools/profiler/blob/f699b44fcf763bdfb6ebf36c32d668e4d296f5d6/src/utils/l10n-ftl-functions.js#L87-L91
```js
export const SHORTDATE: FluentFunction = (args, _named) => {
  const date = args[0];
  const nowTimestamp = Date.now();


  const timeDifference = nowTimestamp - +date;
```

used in:
https://github.com/firefox-devtools/profiler/blob/9c8fb5561537894aee2fe8a265d2c127e83ed0e4/locales/en-US/app.ftl#L679-L683
```
# The function SHORTDATE is specific to the profiler. It changes the rendering
# depending on the proximity of the date from the current date.
# Variables:
#   $date (Date) - The date to display in a shorter way
NumberFormat--short-date = { SHORTDATE($date) }
```

The good thing is that the code was already present as a `toNumber` function. It's not clear to me why the original code didn't use `valueOf`, but doing it in this patch fixes the problem for me.

I actuallty tested locally with our profiler code, that it fixes the issue.

Hope this solution works for you :-)

### DIFF
--- a/fluent-bundle/src/builtins.ts
+++ b/fluent-bundle/src/builtins.ts
@@ -88,7 +88,7 @@ export function NUMBER(
   }
 
   if (arg instanceof FluentDateTime) {
-    return new FluentNumber(arg.valueOf(), {
+    return new FluentNumber(arg.toNumber(), {
       ...values(opts, NUMBER_ALLOWED),
     });
   }
--- a/fluent-bundle/src/builtins.ts
+++ b/fluent-bundle/src/builtins.ts
@@ -88,7 +88,7 @@ export function NUMBER(
   }
 
   if (arg instanceof FluentDateTime) {
-    return new FluentNumber(arg.toNumber(), {
+    return new FluentNumber(arg.valueOf(), {
       ...values(opts, NUMBER_ALLOWED),
     });
   }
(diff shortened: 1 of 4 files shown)


eemeli:
As discussed on Matrix, I took over this PR from @julienw as the proper fix here required a bit more finessing. Effectively, as #636 changed the FluentDateTime `value` and `valueOf()` result to not always be a number, casting a FluentDateTime to a number now also tries to call `toString()` (due to [primitive conversion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Data_structures#primitive_coercion) rules), which fails because it's expecting to get a non-empty `scope` argument.

So to make all of this work and not create further breaking changes, the right thing to do is to:
1. use [Symbol.toPrimitive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive) to more explicitly control the conversion, and
2. allow `toString()` to be called without an argument.

flodolo (approved):

PR #643, opened 2025-07-18
Fix typescript types in DOMLocalization

julienw:
I noticed some types were not good enough when I started checking the types in firefox' lit-utils file, for example in these lines: https://searchfox.org/mozilla-central/rev/a9e77c69ed8a2b408bbe64553316c7849ec3d267/toolkit/content/widgets/lit-utils.mjs#160,179

Indeed in Lit we're dealing with DocumentFragments more. This works fine in Fluent, only the types aren't correct.


julienw:
ping @eemeli

PR #668, opened 2026-06-08
Add @fluent/prettier-plugin package to autoformat Fluent resources

### LINKED ISSUE #667: autoformatting using a prettier plugin
hi 👋🏼,

it would be great if there was an officially supported way to do autoformatting, preferably as a prettier plugin (see below for rationale).

while this is technically possible as a 3rd party package, that's not ideal (see below). if the current maintainers (hi @eemeli 👋🏼 i guess?) of this project are receptive to the idea (hence this issue), i may work on a pull request to this repo to add a `@fluent/prettier-plugin` package.

(fwiw, i locally have some hacky proof of concept code to convince myself that the things described below are actually feasible without too much maintenance effort.)

what i would love to see:

- part of the [projectfluent/fluent.js](https://github.com/projectfluent/fluent.js/) set of JS libraries
- based on `@fluent/syntax`, which can parse + serialize, which solves all `.ftl` syntax technicalities
- sensible defaults, no configuration
- based on prettier
  - widely used
  - straight-forward implementation b/c of the [plugin system](https://prettier.io/docs/plugins#developing-plugins) and healthy ecosystem around it
  - many tools integrate with it
  - automatic support for [fluent-vue](https://github.com/fluent-vue/fluent-vue) b/c the prettier support for `.vue` files, which can contain js, html, and css… and `<fluent>` ftl blocks 😎, can also format embedded languages using any installed prettier plugin
- strict parsing; invalid syntax causes errors. (prettier has similar behaviour for invalid javascript syntax, for example)
- entries of type `Message` and `Term` sort alphabetically, with terms simply sorting between messages (e.g. `example` → `-some-term` → `zzz`)
- [fluent comments](https://projectfluent.org/fluent/guide/comments.html) bound to messages stay bound to messages, and stand-alone comments stay in place, effectively becoming group delimiters for sortable blocks in the `.ftl` file.

prior art:

- https://github.com/luca-iachini/prettier-plugin-fluent (hi @luca-iachini 👋🏼)
  - implements its own output formatting while `@fluent/syntax` also does that
  - no support for sorting
  - not very actively maintained

- https://github.com/desmosinc/prettier-plugin-fluent by (hi @timstallmann @slbls 👋🏼)
  - rather minimal and not very strict. keeps junk, minimal sorting, no special comments handling.
  - not very actively maintained

thoughts welcome!

wbolster:
This adds a @fluent/prettier-plugin subpackage containing a Prettier plugin to autoformat FTL resource files. See added README for details.

The packaging, metadata, license, documentation, etc. follows the conventions used by the the other packages in this monorepo. The package is not included in the typedoc docs, because it is not intended to be used as a library, but indirectly via Prettier as a plugin.

The vitest configuration now also runs tests written in TypeScript, and @types/node is added as dev dependency, because it gets referenced from tsconfig.

Closes #667.

### DIFF
--- a/fluent-prettier-plugin/src/index.ts
+++ b/fluent-prettier-plugin/src/index.ts
@@ -1,0 +1,100 @@
+import * as fluentSyntax from "@fluent/syntax";
+import type { Plugin } from "prettier";
+
+const plugin: Plugin<fluentSyntax.Resource> = {
+  languages: [
+    {
+      name: "Fluent",
+      parsers: ["fluent"],
+      extensions: [".ftl"],
+      linguistLanguageId: 206353404, // see https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml
+    },
+  ],
+  parsers: {
+    fluent: {
+      parse(text) {
+        const resource = fluentSyntax.parse(text, { withSpans: true });
+        const firstJunkEntry = resource.body.find(
+          entry => entry instanceof fluentSyntax.Junk
+        );
+        if (firstJunkEntry) {
+          throw createParseError(text, firstJunkEntry);
+        }
+        return resource;
+      },
+      astFormat: "fluent-ast",
+      hasIgnorePragma(text) {
+        return Boolean(
+          text.trimStart().match(/^#{1,3}\s*(?:@noformat|@noprettier)\b/)
+        );
+      },
+      locStart(node) {
+        return node.span?.start ?? 0;
+      },
+      locEnd(node) {
+        return node.span?.end ?? 0;
+      },
+    },
+  },
+  printers: {
+    "fluent-ast": {
+      print(path) {
+        return fluentSyntax.serialize(sortResource(path.node), {});
+      },
+    },
+  },
+};
+export default plugin;
+
+function sortResource(resource: fluentSyntax.Resource): fluentSyntax.Resource {
+  type SortableEntry = fluentSyntax.Message | fluentSyntax.Term;
+  function compare(a: SortableEntry, b: SortableEntry): number {
+    // Compare entry identifiers, ignoring letter case. For
+    // deterministic results, this deliberately does not use
+    // locale-aware comparisons (e.g. localeCompare()) and case
+    // mappings (e.g. toLocaleLowercase()).
+    const nameA = a.id.name.toLowerCase();
+    const nameB = b.id.name.toLowerCase();
+    if (nameA === nameB) return 0;
+    return nameA > nameB ? 1 : -1;
+  }
+  const entries: fluentSyntax.Entry[] = [];
+  const pending: SortableEntry[] = [];
+  for (const entry of resource.body) {
+    if (
+      entry instanceof fluentSyntax.Message ||
+      entry instanceof fluentSyntax.Term
+    ) {
+      pending.push(entry);
+      continue;
+    }
+    if (pending.length) {
+      entries.push(...pending.sort(compare));
+      pending.length = 0;
+    }
+    entries.push(entry);
+  }
+  entries.push(...pending.sort(compare));
+  return new fluentSyntax.Resource(entries);
+}
+
+type FluentParseError = Error & {
+  // Optional, but Prettier gives nicer error messages when set.
+  loc?: { start: { line: number; column: number } };
+};
+
+function createParseError(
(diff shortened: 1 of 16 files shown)


wbolster:
hi @eemeli, please have a look!

i tried to follow project conventions as much as possible. all tests and packaging pass locally, and also in github actions in my own fork.

note that i chose 0.1.0 as the version number, and the changelog (which i presume is maintained manually?) reflects that.

wbolster:
fwiw, here's a direct link to the readme, which makes it easier to read: https://github.com/wbolster/fluent.js/blob/prettier-plugin/fluent-prettier-plugin/README.md

wbolster:
paging @Demivan (oh hi 👋🏼) b/c this prettier plugin enhances [fluent-vue](https://github.com/fluent-vue/fluent-vue) with formatting inside `<fluent>` SFC blocks

wbolster:
the [prettier docs about plugin development](https://prettier.io/docs/plugins#developing-plugins) may be useful for reviewers. `@fluent/prettier-plugin` strictly follows the official guidelines, without hacks or anything, so this should just keep working as the prettier API has proven to be pretty (🙃) stable over time.

eemeli (changes requested):
Let's hold off on jumping to a solution right away -- it's only been a few days since the discussion on this even started. As I mention in https://github.com/projectfluent/fluent.js/issues/667#issuecomment-4642388966, we should work towards having just one Prettier plugin for Fluent.

One additional reason I'd prefer to have @luca-iachini on board with the work here is to ensure that any new code we might add to this repo will also be maintained, and that's always easier with multiple maintainers.

wbolster:
sure, to keep the discussion centralized i've answered here: https://github.com/projectfluent/fluent.js/issues/667#issuecomment-4658660968

fluent-prettier-plugin/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      },
+    },
+  },
+};
+export default plugin;
+
+function sortResource(resource: fluentSyntax.Resource): fluentSyntax.Resource {
+  type SortableEntry = fluentSyntax.Message | fluentSyntax.Term;
+  function compare(a: SortableEntry, b: SortableEntry): number {
+    return a.id.name.localeCompare(b.id.name);
```
NoahTheDuke: Should this call [`toLocaleLowerCase`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase) on the `id` name? I would prefer it.

wbolster: sure, i'm open to changing that. but now i wonder in general whether this would need a _fixed_ locale, b/c as i understand it `localeCompare()` will fallback to the ‘current’ (whatever that may be) locale, which is nondeterministic.

perhaps this can use the locale `und` (undefined language, e.g. `new Intl.Locale('und')`) instead of hardcoding (some variation of) english.

wdyt?

NoahTheDuke: oh yeah, good call. i don't remember all of the subtleties with choosing locales. we could also "just" choose `"en"`, as it's (sadly) the default language in most cases, but maybe that will have other negatives i've not thought of yet.

wbolster: it seems locale-agnostic case-insensitive deterministic sort can be achieved using `toLowerCase()` (_not_ `toLocaleLowercase()`) and plain `>` for string comparisons. see e.g. https://stackoverflow.com/questions/68785888/deterministic-string-compare-in-js-ts

i've amended this pull request to use this sort function instead (and added test coverage for it):

```typescript
  function compare(a: SortableEntry, b: SortableEntry): number {
    // Compare entry identifiers, ignoring letter case. For
    // deterministic results, this deliberately does not use
    // locale-aware comparisons (e.g. localeCompare()) and case
    // mappings (e.g. toLocaleLowercase()).
    const nameA = a.id.name.toLowerCase();
    const nameB = b.id.name.toLowerCase();
    if (nameA === nameB) return 0;
    return nameA > nameB ? 1 : -1;
  }
```

lgty @NoahTheDuke?

NoahTheDuke: that's great, thank you

PR #671, opened 2026-09-11
feat(fluent-dom): Add an option to connectRoot to opt out of the mutation observer

julienw:
This is a workaround for issue #648.

In apps that are primarily implemented with WebComponents, the number of roots can grow a lot. This proved to be a performance issue. So instead of using a MutationObserver to run translations, the WebComponent framework can call the translate functions directly as part of its lifecycle.


julienw:
What do you think @eemeli ?

fluent-dom/src/dom_localization.js
```diff
@@ -121,8 +123,12 @@ export default class DOMLocalization extends Localization {
...
    *
    * @param {Element | DocumentFragment}      newRoot - Root to observe.
+   * @param {{useObserver: boolean}} use { useObserver: false} to opt out the
+   *                                 Mutation Observer for this root. True by default.
    */
-  connectRoot(newRoot) {
+  connectRoot(newRoot, opts = {}) {
+    const useObserver = "useObserver" in opts ? opts.useObserver : true; // Defaults to true
+
     for (const root of this.roots) {
```
julienw, line 132: I notice that this loop is also quadratic when a lot of roots are added, and it's very unlikely for Lit-based apps, plus the drawbacks of not doing this check is very low if I understand this correctly (again, in a lit-based app where each root has a small DOM).

Maybe a better solution would be a separate localization class... I'll look into that.

julienw: https://github.com/projectfluent/fluent.js/pull/673

